Issue #467, opened 2017-04-17
Ensure correct annotation properties are always used


cmungall:
It's easy in Protege to slip and use e.g. a namespace AP for a definition and vice versa.

We also need to enforce people using hasDbXref in axiom annotations for syns and defs.

I have fixed a bunch of these manually, but we need to implement checks

gouttegd:
WARNING: This issue has been automatically closed because it has not been updated in more than 3 years. Please re-open it if you still need this to be addressed – we are now getting some resources to deal with such issues.

PR #480, opened 2017-07-21
namespace and def APs were inverted. Fixed.

cmungall:
See #467


cmungall:
I'm going to merge even though it's a fail; references to obsolete classes. Needs a new ticket

PR #508, opened 2018-06-05
Stale branch recovery rbc

dosumis:


dosumis:
@cmungall  - Sorry. This appears to be pretty much un-reviewable based on rdfxml diff.  Need to switch to ofn for master first.  Will give that a go tomorrow.

cmungall:
The source is currently ofn, looks like you're saving as rdf/xml

we can still compare using @balhoff's diff tool

dosumis:
Scratched this.  Have moved to fixing in original retinal_bipolar_cell branch.

PR #510, opened 2018-07-10
Pre release

cmungall:


cmungall:
false positive fail, looks like need to add --no-check-certificate to wget. Shouldn't affect things so will release anyway

PR #550, opened 2019-02-20
Retinal bipolar cells

dosumis:


addiehl:
But platelets do secrete serotonin...

dosumis:
> But platelets do secrete serotonin...

This is inferred and will be asserted in release:

image

dosumis:
BTW - thanks for checking.  Diff is a bit bigger than I would like due to a few annotation axioms having type 'string'  in master but not in my branch.   This explains most apparent immune system / blood cell edits. I went through all these quickly to check for any other changes and all looks OK.

dosumis:
Looks like some header axioms dropped too.  I'll add back.

addiehl:
Thanks for clearing that up

PR #551, opened 2019-02-20
Cleaned up cl-edit.owl a bit

matentzn:
Removed a few axioms as well and moved them to a new ontology called cl-removedaxioms.owl.

Please review these axioms carefully and move them to the appropriate place, which, IMHO, is not CL.


dosumis:
Looks good to me.  Was hoping to find more Uberon links that might explain inconsistency I'm getting with pCL.

dosumis:
Merging despite build fail - as that is happening everywhere right now due to 

The command "wget http://build.berkeleybop.org/job/robot/lastSuccessfulBuild/artifact/bin/robot -O bin/robot" failed and exited with 8 during .

PR #552, opened 2019-02-21
ODK 1.2.2 upgrade

matentzn:
- CLO had to be removed from CL imports, because of incompatibilities with UBERON
- PR is currently not mirrored anymore, because it is too big. Same ncbitaxon.


matentzn:
I think this is mergeable - however, I am not 100% certain that this PR preserves all the build functionality that was previously there. @cmungall you may have to make sure that cl.Makefile still does what it should. Not sure what other scripts you have building stuff.

cmungall:
Also file a ticket with CLO

matentzn:
> If it correctly builds the required products, let's merge, we can make changes later.
Ok, cant be 100% sure, but looks as! Eyeballing revealms some cosmetic issues, like the presence of the owl:Nothing class in the signature (coming from RO).

> Will you also check in the project.yml?
Yes, on todo. Currently checked in latest ODK as example file.

> Ideally there would be no hand-edits to the makefile, can some of this be controlled in the yml?
You mean excluding PR and NCBI taxon? I mean, I want them in, and just dont want them in always. So they are commented out for now. But I can add a note to cl.Makefile

> You say there are unsatisfiable classes in pr? They should be notified
They are! I think they have fixed it already.

> We need a better solution for big upstream ontologies. pr should be releasing modules
More next week. 

> Looks like we need to add a basic product to the ODK Makefile as this is used across multiple ontologies
ok

dosumis:
I'm a bit worried about changes to release files affecting downstream users.  See https://github.com/INCATools/ontology-development-kit/issues/175

It seems me that there is no reason to make new release files as part of this pull request - as the rest seems fine.  Can we revert release and then merge?

I guess the downside is we'll need to wait for a base file.  But perhaps OK to push this before other releases, given that (AFAIK) we are currently the only users ?

matentzn:
Ok david, I removed cl.obo, cl.owl and the release imports again. Please merge when content.

PR #556, opened 2019-04-10
Add new Cell Ontology terms and modify several existing terms

1136514475:

PR #569, opened 2019-06-16
1136514475 master

dosumis:
Edits to https://github.com/obophenotype/cell-ontology/pull/556 to comply with patterns (& fix up some related terms)

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -9,13 +9,13 @@ Prefix(oboInOwl:=<http://www.geneontology.org/formats/oboInOwl#>)
 
 
 Ontology(<http://purl.obolibrary.org/obo/cl.owl>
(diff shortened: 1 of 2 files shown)


1136514475:
I have no idea why there is a failing check, what should I do then?

dosumis:
@1136514475  - Apologies this has been hanging so long. Merging now.  Please review and let me know if you need further changes or make another pull request, otherwise I'll assume good to release as-is.  I plan to run a release today to make these changes public. Perhaps we could meet to discuss setting up an efficient pipeline for such requests in future?

1136514475:
Thanks for your work. No changes needed. This pull request is a little difficult for me because I am not familiar with these owl works. Now I think your pipeline is efficient enough. If I need to make more pull requests, I will continue using this pipeline.

PR #582, opened 2019-08-12
Recovered class lost in a previous edit

matentzn:
Recovered a class that was somehow lost in a previous edit. Note that the new release I have just merged in **has this class**; I only made this pull request so when you are back from leave, you can review the existence of this term more carefully.


matentzn:
@dosumis dont forget :)

matentzn:
@dosumis ping :P

Issue #633, opened 2020-02-29
NTR - decidual pericyte


dosumis:
REf: PMID:30429548

lubianat:
Hello, was this closed by #634, then?

dosumis:
> Hello, was this closed by #634, then?

Should have been.  Stayed open due to type in commit message.  'decidual pericyte' should now be available.  Please let me know if any problems.

Issue #626, opened 2020-02-26
New Term Request: decidual stromal cell


dosumis:
BTO has: 

decidual stromal cell http://purl.obolibrary.org/obo/BTO_0005756

Decidual stromal cells are the main cellular component of the human decidua. [ http://www.ncbi.nlm.nih.gov/pubmed/11719592 ]

Add & keep Xref?

dosumis:
Added as synonym on descidual cell = see #625

dosumis:
Fixed (see linked pull requests)

Issue #625, opened 2020-02-26
New Term Request: decidual cell 


dosumis:
In NCIT and BTO but not CL

Decidual Cell  NCIT:C32429
http://purl.obolibrary.org/obo/NCIT_C32429
An endometrial fibroblast that differentiates during pregnancy in response to the implanting embryo by accumulating lipid and glycogen. It is polygonal, possesses a large, vesicular nucleus and is believed to secrete placental prolactin. The decidual cells form a tightly adherent, massive cellular matrix that first surrounds the implanting embryo and later occupies most of the endometrium.…
Ontology: NCI Thesaurus OBO Edition NCIT

decidual cell  BTO:0002770
http://purl.obolibrary.org/obo/BTO_0002770
An enlarged, ovoid, connective tissue cell in the uterine mucous membrane that enlarges and specializes during pregnancy.
Ontology: BRENDA tissue / enzyme source BTO

part_of some decidua http://purl.obolibrary.org/obo/UBERON_0002450

dosumis:
Zoe pointed out that we already have 

'decidua cell' EquivalentTo: 'native cell' and ('part of' some decidua)

This was added in TermGenie and has no reference 👎 .  It only works for this  request if the decidua has only one cell type and it corresponds to the cell type described above.  Worth looking at motivating paper: https://www.nature.com/articles/s41586-018-0698-6

dosumis:
The requested term needed by HCA is actually 'decidual stromal cell'.   The use of stromal here seems to come from the fact that they develop from 'endometrial stromal cells' (https://en.wikipedia.org/wiki/Decidualization#Endometrial_stromal_cells_(ESCs)).  Many pubs just call them decidual cells, as does wikipedia here: https://en.wikipedia.org/wiki/Decidual_cells  + synonyms.

I strongly suspect that the current CL term has only been used for this, not for other cells of the decidua (e.g. decidual natural killer cells) so I think OK to redefine - removing the crappy logical def and  adding a nice text  def.
Good ref: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6046526/


decidual cell
def: "A specialized, enlarged, connective tissue cell of the decidua with enlarged nucleus,  dense membrane‐bound secretory granules and cytoplasmic accumulation of glycogen and lipid droplets.  These cells develop from endometrial stromal cells during decidualization." [xrefs - nature paper; PMC review, wikipedia, BTO, NCIT]
exact_synonym: decidua cell
[...]

PR #634, opened 2020-02-29
Decidual cells

dosumis:
Edits for review: 
http://purl.obolibrary.org/obo/CL_2000002
image
http://purl.obolibrary.org/obo/CL_0008033
image


dosumis:
Wrong ticket numbers on commit messages.  Should be #625  #626  & #633

dosumis:
CC @zperova

PR #689, opened 2020-08-01
Adding immune cell terms; also adding a new template based pipeline to CL

matentzn:
This PR has for components:

1. A general pipeline for syncing templates with googlesheets. This is configured through `src/templates/config.txt`.
2. A general pipeline for compiling all templates in the templates directory into a single template component. This works with all templates configured through `src/templates/config.txt` as well as custom additions. Note that the template `dependencies` is treated as a special case - this will be compiled before the other templates are generated.
3. An instanciation of that generic pipeline using the [XCL ontology templates](https://github.com/jamesaoverton/experimental-cell-ontology).
4. A few new convenience methods in the makefile to generate robot diffs in markdown to make it easier to review pull requests like this.


cmungall:
I have many concerns

jamesaoverton:
@cmungall Our preliminary work was discussed on the CL calls March 11, 25, and 31, as recorded in the call agenda doc:

https://docs.google.com/document/d/1bSpiZc5p9VNCZ83I4VUEwOVGoFTZ3cpXZKOSRAaGBZc/edit

Based on the buy-in we got on those calls, we went forward with the work. This PR makes that work concrete. We have been trying to schedule time on another CL call to go over the details and implications. Because of summer vacations, we have not been able to schedule that time yet. We plan to run it past people on the call, update based on feedback, and then the larger mailing list. If CL developers aren't interested in templates, we can easily merge into `cl-edit.owl`.

I'm quite surprised by your comments. I don't see how templates are a step backwards, and there was strong support for logically-defined species-specific nodes on the CL calls.

cmungall:
How hard is it to separate out this PR? It combines a major change for the workflow, and introduces problematic species-specific classes. These should be distinct PRs.

Apologies, the I am never able to make the Wednesday call at the current time, I thought I'd flagged this several times. I joined the Monday call today where I heard about this work. I can however make a one-off Wednesday call to discuss these major issues.

I'll post more later but in the interim would it be possible to include more documentation in this PR? It looks like README-editors.md is unaffected yet this brings in a major change to the workflow. The editors file should have a link to the spreadsheets. I assume there is an SOP for bringing in chunks of the edit file into the spreadsheets (looks like just certain portion of the immune hierarchy?). Is there any system to assist editors in knowing where the source of truth is for any class/axiom? It looks like at the moment there is an assumption that someone editing in Protege will do a mouse-over to see axiom source, if it's not cl-edit then they go to the google sheet. Is there a system that prevents zombie axioms? E.g. if I accidentally delete an axiom in Protege rather than in the sheet how do I know I made a mistake?

matentzn:
These concerns are easy to separate. The template thing is really just a side loader that does very few things, and it does not advocate a new style of curation per se. This PR is purely about the cell terms - you can forget about the templates. Once this gets merged, we can replace the patterns or move back to a purely Protege based approach, or improve QC - whatever the cell editors vote for.

matentzn:
Reviewing the hierarchy (https://ols.monarchinitiative.org/ontologies/cl) and the [robot diff](https://github.com/obophenotype/cell-ontology/blob/3405e205074dd4b3576fcbbd93278b771b8caddc/src/ontology/reports/diff_edit_imports.md) are the only things that matter here I think!

cmungall:
@matentzn I'm not following. This PR removes a lot of classes from cl-edit.owl. If it's intended as a new term request this should be separate.

matentzn:
Just to clarify: no class is removed - some axioms that are, in this PR, moved to `components/all_templates.owl`!

bpeters42:
I do think we need another call. As pointed out, there was an initial call that included everyone here (apart from Nico) and based on that James and I thought that we were all on the same page regarding the need to fix the current complete mess of inconsistencies regarding taxon specificity in CL, the proposed solution to use template-driven design patterns to make it happen, the desire to create taxon-specific views of CL, and the approach to create taxon-specific grouping classes to make the merged CL to not look super crappy.  
Clearly we are not all on the same page, but I am hopeful that the differences can be identified and overcome.

cmungall:
> Just to clarify: no class is removed - some axioms that are, in this PR, moved to components/all_templates.owl!

I'm aware of this. It is still removed from the edit file necessitating a different workflow for editing these.

all: here is what I suggest. We can easily make incremental progress here. The current PR is too big and combines too much in one, species-specific grouping classes, changes in edit workflow. But as a compromise I suggest:

 1. take out broad species-specific groupings like {mouse,human} {cell,lymphocyte}. We can discuss the pros and cons of these separately but there is no need for them to go in in the initial PR.
 2. go ahead and merge and we can iterate on the workflow. Nico assures me it will be easy to merge selected axioms back into the edit file. My position is that it is a mistake to force a template workflow on non-catalog/compositional classes, but this should be a group decision made when the pros and cons are clearly articulated to all stakeholders. However we can iterate on this, I don't want to block work.

Moving forward I would like the following:

 - proposals should be made as tickets so that those not able to attend calls can comment asynchronously, and that things on the call can be made more concrete and articulated with minimal ambiguity
 - PRs are generally kept small, and separate concerns are divided into separate PRs
 - make more use of the cl_edit mailing list for proposed changes to edit workflow

dosumis:
I think this is important and necessary work, but think we need some discussion before actioning.  

### Why I think this work is important and justified:

It's difficult to see how we can have marker-based definitions without recording species.  We can’t avoid marker-based definitions for immune cells once we get past the most general terms.  The original marker-based definitions were a mess because they tried to be species independent:  mixing mouse and human markers in a single EquivalentClass definition makes no sense. This needs to be fixed somehow. Adding species specific terms is the simplest solution. Another possible solution is to use GCIs referencing species neutral terms, but I don't think this can work for the most specific terms.

### Specific issues to address (partly overlapping with Chris' concerns):

1. Re - the template issue: I think we need some clarity on plans for long term usage of templates for these terms. Is this just one-off term addition, or will the be a standard route for immune cells.  I think it makes sense to have the infrastructure for template-based term addition built into the repo, even if it is being used for one-off term addition.  We need a better convention to indicate where this is the case though.

2. Do we really need very general species specific terms (e.g. cell, mouse; lymphocyte, human)?  Many users are likely to be confused to see such limited numbers of cells under cell, mouse and then worry about CL being radically incomplete. As Chris suggests, I think we should split these out from the current pull request. 

3. Inferred species specific classes with non-species specific names:
image
Are these intended?  Should they be renamed?  Would renaming change the meaning too much - assuming terms have already been used?

4. Logical definitions with a mix of species specific and species neutral marker terms seem to be common.  I'd like to see some justification of this based around specific examples.

cmungall:
Agree 100% with David. Note my objections to species-specificity is about the groupings. Species specificity at the leaves for marker-defined cells makes complete sense. But see #695 we need design patterns and rules for how this will be done.

I mention the species grouping classes problem in this post. See the section of ragged lattices. We actually have this problem already in CL, e.g. 'kidney cell'
https://douroucouli.wordpress.com/2020/09/04/the-open-world-assumption-considered-harmful/

addiehl:
Several comments:
1.  The choice of using a template approach for this larger revision was somewhat experimental.  As the scientific curator for this content, I was game to try something different, but for smaller changes to CL, I would probably still prefer just using Protege.

2.  The actual changes to the innate lymphoid cell hierarchy and the term additions are requested by HIPC and long overdue.  There are a few additional fixes that I want to do after this has been merged as well.

3.  I'm not a huge fan of the 'cell, human', 'lymphocyte, human' classes, personally, for all the reasons Chris has pointed out.  However a compromise position could be to create a separate CL_taxon.owl file that only contains these sorts of defined class grouping classes, and for certain use cases, this could be loaded with the CL.owl file and reasoned over to create a hierarchy with cells grouped by taxon.

4. The reasoning of the Ly-76 erythroblasts, etc. under 'mouse, cell' was unexpected, since these were not part of this revision (not present in the Google spreadsheet template or xcl.owl) and do not have explicit taxon restrictions currently, though they have a mouse specific PRO marker (nice to see the reasoning works indirectly).  Thus, appending the labels with ", mouse" was not performed, and could be done as part of a later revision.

5.  I agree that species-specific PRO markers should be used for species-specific cell type classes.

cmungall:
On Tue, Sep 15, 2020 at 1:10 PM Alexander Diehl <notifications@github.com>
wrote:

> Several comments:
>
>    1.
>
>    The choice of using a template approach for this larger revision was
>    somewhat experimental. As the scientific curator for this content, I was
>    game to try something different, but for smaller changes to CL, I would
>    probably still prefer just using Protege.
>
>
I'm all for  experiments, but maybe separate the experimental part into a
separate PR?

>
>    1.
>
>    The actual changes to the innate lymphoid cell hierarchy and the term
>    additions are requested by HIPC and long overdue. There are a few
>    additional fixes that I want to do after this has been merged as well.
>
> I see, so this PR combines both a change in source (ofn->tsv) as well as
structural changes to the hierarchy.. all the more reason for separate PRs

But to reemphasize, go ahead and merge, we can sort workflow issues out
later

>
>    1. I'm not a huge fan of the 'cell, human', 'lymphocyte, human'
>    classes, personally, for all the reasons Chris has pointed out. However a
>    compromise position could be to create a separate CL_taxon.owl file that
>    only contains these sorts of defined class grouping classes, and for
>    certain use cases, this could be loaded with the CL.owl file and reasoned
>    over to create a hierarchy with cells grouped by taxon.
>
> I don't think this works. They have to be included in the main CL.owl
otherwise the PURLs don't resolve.

Why not instead just make mouse views human views etc as we do for uberon?

Of course this will require more comprehensive TCs, see my recent issues/PRs


>    1.
>
>    The reasoning of the Ly-76 erythroblasts, etc. under 'mouse, cell' was
>    unexpected, since these were not part of this revision (not present in the
>    Google spreadsheet template or xcl.owl) and do not have explicit taxon
>    restrictions currently, though they have a mouse specific PRO marker (nice
>    to see the reasoning works indirectly). Thus, appending the labels with ",
>    mouse" was not performed, and could be done as part of a later revision.
>
> Another reason to edit these in an IDE where reasoning is integrated and
transparent!

What is the SOP for noticing these kinds of things with the proposed
spreadsheet workflow?

>
>    1. I agree that species-specific PRO markers should be used for
>    species-specific cell type classes.
>
>
good

bpeters42:
As a reminder: this is where the work got initially scoped out, and we had gotten buy in: 
https://docs.google.com/document/d/1bSpiZc5p9VNCZ83I4VUEwOVGoFTZ3cpXZKOSRAaGBZc/edit#heading=h.f3vxn8dhngq

There are implicit suggestions in the document how to create taxonomy-specific views of ontologies (for CL and beyond).

dosumis:
Hi @bpeters42,

I don't think anyone's objecting to using in_taxon for taxonomically restricted classes.  This can work with taxon restrictions that are recorded much more extensively elsewhere (GO & Uberon).  The only major objection to the proposed changes in CL content is to adding high level grouping classes such as `cell, mouse` and `lymphocyte, human`.  Other, more minor issues mentioned by Chris and myself could be dealt with incrementally after addition a merge.

cmungall (approved):

PR #690, opened 2020-08-01
Updated CL to new ODK 1.22

matentzn:
I made a release without refreshing imports,  and all files look correct in the diff.


matentzn:
As no one can easily judge the correctness of this, I will merge it after travis passes.

PR #692, opened 2020-08-02
fixing errors in the cl-edit file

matentzn:
- changing CL license URL
- Removing some forgein class declarations
- Removing any mentions of GO:cell, as it is obsolete now (in particular: equivalence class axiom, comment, and one logical definition, CL_0001061).
- Changed two definitions with @addiehl as per #691
- Removed a bunch of foreign axioms from GO and PATO (see diff)


matentzn:
@dosumis I am pretty confident this is all in order so I will merge it, but please take a look when you see it next week ;)

PR #745, opened 2020-09-25
Delete README.txt

cmungall:
I think this is outdated and redundant.

As per ODK, we should have

 - high level info in root README.md
 - editors info in src/ontology in README-editors.md


dosumis:
+1

PR #754, opened 2020-09-28
First draft of annotation filtering

matentzn:
I tried this on all ontologies; seems to work. I will leave the rest to you @dosumis


matentzn:
(i added Pato as an example)

dosumis:
Still seeing a lot of cruft in Protege.  e.g. PATO import has:

image

Do I need to run update imports locally to test?

matentzn:
I  have now updated pato.. No it was a major headache with make.. GR. Check out pato import now!

src/ontology/cl-edit.owl
```diff
@@ -2403,7 +2403,6 @@ Declaration(Class(obo:CP_0000039))
 Declaration(Class(obo:CP_0000040))
 Declaration(Class(obo:CP_0000043))
 Declaration(Class(obo:D96882F1-8709-49AB-BCA9-772A67EA6C33))
```
cmungall, line 2405: this is pretty wacky (not in your PR, I know)

cmungall (approved):
The allowed list looks good. I don't see the logic for getting subAPs of synonym_type and subset_type

matentzn:
not there yet, will do it tomorrow.

matentzn:
I am doing this now:

```
remove -T allowed_annotations.txt --select "self descendants" --select complement --select annotation-properties 
```

which is a bit of a trick, assuming that we generally want to include child APs. If we didnt, we would have to create special handling for subsets - like this it is generic (in the case of PATO, does exactly what what we want). The downside is that this could pull in a few APs we didnt want - but these would not show up in the Protege list, since they are children.

dosumis (approved):

dosumis:
Updating other imports before I merge.

dosumis:
(I guess this might cause problems with import updates in branches + merging ...)

matentzn:
No worries at all. Its just one, and that is mine (I can fix the conflicts quickly)

PR #756, opened 2020-09-29
WIP: Issue 709 uberon base

cmungall:
Don't merge yet. The problem is (as is typical with base) we bring in too many danglers


matentzn:
This is now obsolete.

PR #766, opened 2020-10-01
ntr: Add Axl+ DC (human)

### LINKED ISSUE #764: [NTR] Axl+ dendritic cell, human

**Preferred term label:**

Axl+ dendritic cell, human

**Synonyms**
Axl+ DC, AS DC 

A dendritic cell characterized by the expression of _AXL_, and _SIGLEC6_ genes.
PMID:28428369
PMID:29221729
PMID:31156637

http://purl.obolibrary.org/obo/CL_0001056 

Immune System -> UBERON:0002405

0000-0003-2473-2313

It is not clear on the literature if the parent class would be a plasmacytoid dendritic cell, a myeloid dendritic cells, or neither.

Obs: is taxon constraint inherited from parent class?

lubianat:
Insert term for  Axl+ DC (human) with 3 refs.

I wasn`t sure how to add a taxon constraint, so that was not added yet. 

Closes #764 .

I understand that there are other immune system projects going on, so if it does not fit the big picture it is totally ok.


dosumis (changes requested):
Edit is to wrong file.  Should be to src/ontology/cl-edit.owl

dosumis:
@lubianat will recreate on correct file and make another pull request

PR #767, opened 2020-10-01
Upper motor neuron

dosumis:
image
image


paolaroncaglia:
@dosumis 
If I'm reading [this diff](https://github.com/obophenotype/cell-ontology/pull/767/commits/fb0ab35195a2999bedd27b70f72e0da4906e3cf0) right, you made 'Betz cell' a part of 'primary motor cortex'. Based on [this ref](https://www.ncbi.nlm.nih.gov/books/NBK10962/) ("The pyramidal cells of cortical layer V (also called Betz cells) are the upper motor neurons of the primary motor cortex"), we could even be more specific and point to UBERON:0022315 'primary motor cortex layer 5'? Also confirmed by [this ref](https://www.ncbi.nlm.nih.gov/books/NBK554616/) ("In the primary motor cortex are the cell bodies of the upper motor neurons, referred to as Betz cells. Specifically, these cells are located in layer 5 of the motor cortex and have long apical dendrites that branch up into layer 1.") and by [this Wikipedia entry](https://en.wikipedia.org/w/index.php?title=Betz_cell&oldid=977472330) ("Betz cells (also known as pyramidal cells of Betz) are giant pyramidal cells (neurons) located within the fifth layer of the grey matter in the primary motor cortex.").
That is, if my understanding is correct that for neurons with potentially long processes, the part_of axiom indicates where the body of the cell is found (not where it synapses). Please correct me if that's not the case. Cheers!

dosumis:
Hi Paola, I agree that we _could_ add an axiom linking to layer 5, but I'm a bit worried that this might not generalise across species.  Maybe best to add a new ticket for this.  I can check with collaborators at Allan.  They are working on cell types in motor cortex for a range of mammalian species.

src/ontology/cl-edit.owl
```diff
@@ -27855,6 +27867,29 @@ SubClassOf(obo:CL_4023050 obo:CL_4023040)
...
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref "DOI:10.1101/2020.03.31.016972") obo:IAO_0000115 obo:CL_4023052 "a Betz cell that synapses with lower motor neurons directly.")
+AnnotationAssertion(obo:RO_0002175 obo:CL_4023052 obo:NCBITaxon_9606)
+AnnotationAssertion(<http://purl.org/dc/elements/1.1/creator> obo:CL_4023052 <http://orcid.org/0000-0001-7258-9596>)
+AnnotationAssertion(rdfs:label obo:CL_4023052 "Betz upper motor neuron")
+SubClassOf(obo:CL_4023052 obo:CL_0008049)
+SubClassOf(obo:CL_4023052 ObjectSomeValuesFrom(obo:RO_0002120 obo:CL_0008039))
+
+# Class: obo:CL_4023053 (spinal interneuron syanpsing Betz cell)
+
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref "DOI:10.1101/2020.03.31.016972") obo:IAO_0000115 obo:CL_4023053 "a Betz cell that syanpses with spinal interneurons")
```
dosumis, line 27883: Worth chasing up some of the refs in DOI:10.1101/2020.03.31.016972 that directly investigate synapsing to see if you could use one of these as an a ref (DOI:10.1101/2020.03.31.016972 doesn't have any connectivity data itself).

dosumis: These should be caught by QC checks + please let me know if not: First letter sentence not capitalised; No full stop at end.

dosumis (commented):
Some relatively minor tweaks.  See comments

shawntanzk (approved):

PR #771, opened 2020-10-03
add new class CL_00000000000000000000000000000000000000004012501 (mus…

meena74:
…cle migratory precursor cell)

added new term
just for example


dosumis:
I guess this was just added as an exercise during training.  Most obvious issue:  ID length should be set to 7, but appears to be set to some large (default?) number.

Issue #435, opened 2016-05-23
CL:0000598 pyramidal cell, part_of relationship


jahilton:
CL:0000598
pyramidal cell
Pyramidal neurons have a pyramid-shaped soma with the apex and an apical dendrite pointed toward the pial surface and other dendrites and an axon emerging from the base. The axons may have local collaterals but also project outside their region. Pyramidal neurons are found in the cerebral cortex, the hippocampus, and the amygdala. [database_cross_reference: GOC:tfm][database_cross_reference: MESH:A08.186.211.577.405.700]

This term is missing part_of relationships
-part_of some cerebral cortex (UBERON:0000956)
-part_of some Ammon's horn (UBERON:0001954)
-part_of some amygdala (UBERON:0001876)

ybradford:
The part of relationships have to be true for all instances of that cell type. To deal with the various locations, a new term 'cerebral cortex pyramidal neuron' can be created with child sister classes: 'hippocampal pyramidal neuron', 'Meynert cell' (with an updated term def.), 'primary motor cortex pyramidal cell', and new term ' amygdala pyramidal cell'.

jahilton:
That sounds like a great plan. Thank you

bonitalam:
Hello, would it be possible to add a part_of relationship to brain (UBERON:0000955)?

shawntanzk:
made a mistake and accidentally commit to master instead of a branch >.< reopening

PR #774, opened 2020-10-08
Add part_of brain to CL_0000598 (pyramidal neuron)

bonitalam:
This addresses issue  #435 

Currently the definition states: "Pyramidal neurons are found in the cerebral cortex, the hippocampus, and the amygdala." We think it's valid to add a part_of relationship to brain.


matentzn:
Better ask @dosumis for a review

dosumis:
Here's a pyramidal cell type that projects down the spine: I think the best we can do is part_of the CNS.

bonitalam:
> Here's a pyramidal cell type that projects down the spine: I think the best we can do is part_of the CNS.

Ahh I see, thanks for the insight @dosumis. Closing this PR then.

PR #783, opened 2020-10-20
Fixes #782

### LINKED ISSUE #782: [Typo/Bug] Insane GCI
This is bad:

image

It needs to go ASAP.

dosumis:
I'm inclined to force merge this as soon as tests pass.  @addiehl - wondering if you recognise this as a stray axiom that should be defining some term, rather than hanging around as a GCI assertion about all things that are in_taxon some human.

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -3682,7 +3682,7 @@ SubClassOf(Annotation(oboInOwl:is_inferred "true"^^xsd:string) obo:CL_0000096 ob
 
 # Class: obo:CL_0000097 (mast cell)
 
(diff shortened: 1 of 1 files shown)


dosumis:
@matentzn this kind of thing could potentially do some serious damage.  I'm wondering what tests we might think up to try to avoid this happening again.  It might be sufficient to add some disjointness declarations - or perhaps also to have a warning for GCIs lacking a genus on the left.

dosumis:
@matentzn @nicolevasilevsky - I think worth running a second release today to ensure this is not floating around causing weird inferences in the OBO world.  Do you have time to run one?

Issue #824, opened 2020-12-04
[NTR] nephrogenic zone cell


paolaroncaglia:
cc @pnejad 
Stemming from https://github.com/HumanCellAtlas/ontology/issues/46

**Preferred term label:**
nephrogenic zone cell

**Synonyms**
cortical nephrogenic zone cell
cortical nephrogenic niche cell

**Definition (free text, please give PubMed ID in format PMID:XXXXXX)**
A kidney cortical cell that is part of the nephrogenic zone.
PMID:29449449

**Parent cell type term (check the hierarchy here https://www.ebi.ac.uk/ols/ontologies/cl):**
equivalent to CL:0002681 ’kidney cortical cell’ part_of some UBERON:0009871 'nephrogenic zone'

**What is the anatomical structure that the cell is a part of? Please check Uberon: https://www.ebi.ac.uk/ols/ontologies/uberon**
See above

**Your nano-attribution (ORCID)**
https://orcid.org/0000-0002-3163-0115 [this is Parisa Nejad's]

paolaroncaglia:
- [ ] Add subset tag "added_for_HCA" to CL_0009019 (nephrogenic zone cell)

PR #837, opened 2020-12-18
Add new class CL_0009019 (nephrogenic zone cell)

paolaroncaglia:
Addresses #824

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -1867,6 +1867,7 @@ Declaration(Class(obo:CL_0009003))
 Declaration(Class(obo:CL_0009004))
 Declaration(Class(obo:CL_0009005))
 Declaration(Class(obo:CL_0009015))
+Declaration(Class(obo:CL_0009019))
 Declaration(Class(obo:CL_0010000))
 Declaration(Class(obo:CL_0010001))
 Declaration(Class(obo:CL_0010002))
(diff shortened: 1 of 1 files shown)


src/ontology/cl-edit.owl
```diff
@@ -21954,6 +21955,19 @@ AnnotationAssertion(rdfs:label obo:CL_0009015 "Peyer's patch follicular dendriti
...
+AnnotationAssertion(obo:CL_0009008 obo:CL_0009019 "")
+AnnotationAssertion(obo:IAO_0000115 obo:CL_0009019 "A kidney cortical cell that is part of the nephrogenic zone.")
+AnnotationAssertion(oboInOwl:created_by obo:CL_0009019 <http://orcid.org/0000-0002-2825-0621>)
+AnnotationAssertion(oboInOwl:hasDbXref obo:CL_0009019 "PMID:29449449")
+AnnotationAssertion(oboInOwl:hasDbXref obo:CL_0009019 "https://orcid.org/0000-0002-3163-0115")
+AnnotationAssertion(oboInOwl:hasExactSynonym obo:CL_0009019 "cortical nephrogenic niche cell")
+AnnotationAssertion(oboInOwl:hasExactSynonym obo:CL_0009019 "cortical nephrogenic zone cell")
+AnnotationAssertion(rdfs:label obo:CL_0009019 "nephrogenic zone cell")
+EquivalentClasses(obo:CL_0009019 ObjectIntersectionOf(obo:CL_0002681 ObjectSomeValuesFrom(obo:BFO_0000050 obo:UBERON_0009871)))
+SubClassOf(obo:CL_0009019 obo:CL_0002681)
```
dosumis, line 21969: Delete subClassOf assertion - don't need to duplicate genus also in EquivalentClass assertion.  Not a big deal, but I prefer to avoid.  More problematic would be if asserted something that is non-trivially inferred by reasoner.

dosumis (changes requested):
Minor quibble but would prefer to avoid redundancy here - see comment.

src/ontology/cl-edit.owl
paolaroncaglia: @dosumis 

> Delete subClassOf assertion - don't need to duplicate genus also in EquivalentClass assertion. Not a big deal, but I prefer to avoid. More problematic would be if asserted something that is non-trivially inferred by reasoner.

Done, thanks.

dosumis (approved):

PR #855, opened 2021-01-19
Recreation of cl-edit.owl from #775 PR

dosumis:
Serialization was so badly screwed up in the original PR (#775),
I had to recreate by mining old commits + final file.
Think I caught everything, but will have missed any changes
on final commit outside of 3 added terms.

@mkeays - please review ASAP in case I missed anything important.  Will merge first thing tomorrow.


mkeays:
Sorry @dosumis , only just seen this (too many notifications!) -- looks good to me, thanks for sorting it out.

Issue #860, opened 2021-01-21
[NTR] Lamp5 GABAergic Interneuron


shawntanzk:
Note - please check that the term does not already exist (check OLS: https://www.ebi.ac.uk/ols/ontologies/cl)

**Preferred term label:**
Lamp5 GABAergic Interneuron

**Synonyms**

**Definition (free text, please give PubMed ID in format PMID:XXXXXX)**
A GABAergic neuron that expresses Lamp5

**Parent cell type term (check the hierarchy here https://www.ebi.ac.uk/ols/ontologies/cl):**
GABAergic interneuron

**What is the anatomical structure that the cell is a part of? Please check Uberon: https://www.ebi.ac.uk/ols/ontologies/uberon**

**Your nano-attribution (ORCID)**

**Any additional notes or concerns**

Related to #827
xref (ILX:0770149)
Ticket not complete!

Issue #859, opened 2021-01-21
[NTR] Alpha 7 Cell (mus MOp)


shawntanzk:
Note - please check that the term does not already exist (check OLS: https://www.ebi.ac.uk/ols/ontologies/cl)

**Preferred term label:**

Alpha 7 Lamp5 GABAergic interneurons of the primary motor cortex (Mus musculus)

**Synonyms**
ɑ7 Lamp5 MOp (mus)

**Definition (free text, please give PubMed ID in format PMID:XXXXXX)**

A GABAergic interneuron of Lamp5 subclass that is strongly labelled for α7 nAChRs. that are located in L1 of the primary motor cortex and have multipolar dendrites with vertically descending axonal collaterals that go deep into the column, usually branching in L5A, but often terminating there. Unlike typical Lamp5 electrophysiology,  alpha7 cells have Vip- and Sst-like firing, with active depolarizing hump at near-threshold membrane potentials, and show larger hyperpolarization sag, stronger burst, and rebound firing

*may have too much details in description 

**Parent cell type term (check the hierarchy here https://www.ebi.ac.uk/ols/ontologies/cl):**
GABAergic interneuron 

**What is the anatomical structure that the cell is a part of? Please check Uberon: https://www.ebi.ac.uk/ols/ontologies/uberon**

http://purl.obolibrary.org/obo/UBERON_0001384 - Primary Motor Cortex
*has some soma location, can specify layer

**Your nano-attribution (ORCID)**

**Any additional notes or concerns**
Add Lamp5 GABAergic Interneuron as parent

patrick-lloyd-ray:
> that are located in L1 of the primary motor cortex and have multipolar dendrites with vertically descending axonal collaterals that go deep into the column, usually branching in L5A, but often terminating there. 

Suggest: "These cells are located in L1..." rather than "that are located in L1."

PR #861, opened 2021-01-21
Added Lamp5 and alpha7 cell

shawntanzk:
Draft for #860 and #859

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -2439,6 +2439,8 @@ Declaration(Class(obo:CL_4023003))
 Declaration(Class(obo:CL_4023004))
 Declaration(Class(obo:CL_4023005))
 Declaration(Class(obo:CL_4023006))
+Declaration(Class(obo:CL_4023010))
+Declaration(Class(obo:CL_4023011))
 Declaration(Class(obo:CP_0000000))
 Declaration(Class(obo:CP_0000025))
 Declaration(Class(obo:CP_0000027))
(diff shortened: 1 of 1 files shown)


patrick-lloyd-ray (changes requested):
The definition you have is: 

> "A GABAergic neuron of the Lamp5 subclass that is strongly labelled for α7 nAChRs. These cells are located in L1 of the primary motor cortex and have multipolar dendrites with vertically descending axonal collaterals that project deep into the column, usually branching and terminating in L5A. Unlike typical Lamp5 electrophysiology, they have Vip- and Sst-like firing properties, with active depolarizing hump at near-threshold membrane potentials, and  larger hyperpolarization sag, stronger burst, and rebound firing")"

I would change to the more standard: "A <genus> that is <differentia" such as "A Lamp 5 GABAergic cortical interneuron that is strongly labelled for α7 nAChRs." (assuming I have the genus label correct). 

I think the rest of the text might technically be a comment annotation, rather than part of the definition. Although, it could probably be moved later if it's providing necessary clarity. There is also a period missing at the end of the last sentence.

shawntanzk:
Thanks @patrick-lloyd-ray - will make the changes. 
@dosumis thoughts on putting the rest of text into a comment annotation now?

I think what we can do in the end (and I'll edit the rest of the definitions likewise) is have that first line like definition + forest generated definition (with minimum identifiable (sorry if im using wrong terms) markers) as definition, with the rest of the information going down as a comment annotation. Cell cards can then combine these as a description (with any other added info). Thoughts? (please do help me cc anyone who should be involved in this discussion too, thanks) 

PS will hold of editing the pull requests until we make a decision on this.

dosumis (approved):

PR #901, opened 2021-01-29
Add ensembl import

matentzn:
I am _not_ taking the heart for this. :D


dosumis (approved):
I take full responsibility for this hack.

bvarner-ebi:
Checking in on this PR. It has been open since Jan 2021 with no action since July 2021.
Should it be merged or cancelled?

matentzn:
I would close this but send an email to David that you did.

bvarner-ebi:
@dosumis, I plan to close this PR by end of week unless there is a plan on deck for next steps.

bvarner-ebi:
Added to https://github.com/orgs/obophenotype/projects/13.

dosumis:
@shawn - Please remind me. Did we drop this plan and use PRO after all?

shawntanzk:
historical:

- most ontologies used HGNC instead of ensembl -> not wrong just have to be aware
- HGNC representation is closer to what people look at (eg genes responsible for something) 
- currently using conflation -> referring to a gene and reference protein = the same thing (merge ref protein node with gene node) @matentzn to dig up article about this

Review necessity of referring to genes in logical axioms: 
if it doesn't matter then we can just conflate -> probably wont affect 
! need use case for logical modeling !

shawntanzk:
@dosumis - have thought about this, and I think we should keep this away from CL until we figure out how PCL comes into CL. atm the only use case I can think of is scRNAseq types, and even that we currently don't use markers to build the hierarchy. As @matentzn pointed out we can then just conflate PRO and genes as it has no impact on the logical side of things (might not be accurate but it does keep things neater and avoid a lot of difficult questions). Lmk if you agree :) thanks

bvarner-ebi:
@dosumis, this PR is from 2021. Can we close this?

bvarner-ebi:
Closing this PR: It originated from 2021 and has not had any noted significant progression.
fyi @dosumis

Issue #811, opened 2020-11-19
NTR: gastrointestinal tract (lamina propria) macrophage of small intestine


ngeorgeebi:
Note - please check that the term does not already exist (check OLS: https://www.ebi.ac.uk/ols/ontologies/cl)

**Preferred term label:**
small intestine lamina propria resident macrophage

**Synonyms**
macrophage of small intestine

**Definition (free text, please give PubMed ID in format PMID:XXXXXX)**
A macrophage which is resident in the lamina propria of the small intestine

**Parent cell type term (check the hierarchy here https://www.ebi.ac.uk/ols/ontologies/cl):**
gastrointestinal tract (lamina propria) macrophage:
http://purl.obolibrary.org/obo/CL_0000865

**What is the anatomical structure that the cell is a part of? Please check Uberon: https://www.ebi.ac.uk/ols/ontologies/uberon**

lamina propria of small intestine: http://purl.obolibrary.org/obo/UBERON_0001238

**Your nano-attribution (ORCID)**
https://orcid.org/0000-0003-4183-8865

**Any additional notes or concerns**

paolaroncaglia:
@ngeorgeebi 
You requested 'gastrointestinal tract (lamina propria) macrophage of small intestine'

This already exists in fact:

CL:0000865 'gastrointestinal tract (lamina propria) macrophage'
equivalent to
'gut-associated lymphoid tissue macrophage'
 and ('part of' some 'lamina propria of small intestine')

@dosumis @addiehl 
CL:0000865 has no PMID dbxref or additional info, but the equivalence axiom clearly states small intestine. Should we add an exact synonym "gastrointestinal tract (lamina propria) macrophage of small intestine" and a gloss to the definition "A gut-associated lymphoid tissue macrophage found in the lamina propria of _small intestine_."?

Thanks,
Paola

dosumis:
@addiehl - any preference?
Might be safer to make new term and move restriction to small intestine to logical def of that?

paolaroncaglia:
Resolution from today's HCA/SCA curators call: Paola to convert existing term into a broader parent, and add specific children

addiehl:
1) So the broader parent will be 'gut-associated lymphoid tissue macrophage' without the 'part of' some 'lamina propria of small intestine' axiom?  Will you propose a more general axiom for anatomical localization?

2) What child classes do you foresee:
gastrointestinal tract (lamina propria) macrophage of small intestine?
[...]

PR #947, opened 2021-02-16
Revise CL_0000865 (gastrointestinal tract (lamina propria) macrophage…

paolaroncaglia:
…) and add 2 children

Addresses #811

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -1868,6 +1868,8 @@ Declaration(Class(obo:CL_0009002))
 Declaration(Class(obo:CL_0009003))
 Declaration(Class(obo:CL_0009004))
 Declaration(Class(obo:CL_0009005))
+Declaration(Class(obo:CL_0009007))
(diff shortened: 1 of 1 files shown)


paolaroncaglia:
@dosumis 
Could you review please. In particular:
These are my first CL edits done on a branch rather than a fork;
Please double-check the new terms' IDs. I had already created and committed terms with IDs from CL:0009001 to 9005, and from CL:0009009 to CL:0009019; the 2 new terms I created now were given IDs 9007 and 9008, is that all right?
Thanks,
Paola

dosumis:
> The 2 new terms I created now were given IDs 9007 and 9008, is that all right?

 Assume these are just what Protege generates from your ID range, so would expect sequential with order of generation.  They appear to be free though.  Maybe check any other branches you have open for ID clashes.

dosumis:
Other than that, looks good, so will set to approve.

dosumis (approved):

PR #958, opened 2021-02-22
Added (Mus musculus) to interneurons

shawntanzk:


dosumis:
Hi Shawn,  Having looked into these some more, I think we need to keep them species neutral. We can support this via the shift to using species neutral PRO terms for these, and the mouse (uniprot derived) terms for the mouse specific subclasses.  Sorry for the confusion.

shawntanzk:
Got it - was just about to talk to you about keeping it species neutral after doing some reading today, seems more sensible to do so as human and marmoset also falls into the same groupings etc. Anyway, will close this without merging the pull request then

PR #962, opened 2021-02-25
Changed Ensembl to Pro through proper import

shawntanzk:


dosumis (approved):
Looks good.

dosumis (commented):
Looks like we're relying on the species specific markers to infer species restriction on cell?

Given that, can you use a DL query to check it's working?  If not, we may need to add a property chain to RO.

expresses o in_taxon -> in_taxon (Written as a rule in English: If X expresses Y and Y is in taxon Z then X is in taxon Z.

Issue #840, opened 2021-01-07
[NTR] fetal hepatobiliary progenitor cell


paolaroncaglia:
Cc @pnejad 
Stemming from a request from a Human Cell Atlas curator (https://github.com/HumanCellAtlas/ontology/issues/72):

**Preferred term label:**
fetal hepatobiliary progenitor cell

**Synonyms**
exact synonym: fetal hepatobiliary hybrid progenitor cell
related or narrow synonym (as it’s an abbreviation): HHyP cell

**Definition (free text, please give PubMed ID in format PMID:XXXXXX)**
A progenitor cell with hepatic and biliary lineage potential, identified in mouse and human, and anatomically restricted to the ductal plate of fetal liver.
PMID:31350390

**Parent cell type term (check the hierarchy here https://www.ebi.ac.uk/ols/ontologies/cl):**
subclass of progenitor cell

**What is the anatomical structure that the cell is a part of? Please check Uberon: https://www.ebi.ac.uk/ols/ontologies/uberon**
part of UBERON:0004820 bile duct epithelium
part of UBERON:0000323 late embryo

**Your nano-attribution (ORCID)**
https://orcid.org/0000-0002-3163-0115

**Any additional notes or concerns**
subset tag: added_for_HCA

I’m unsure about axiomatising any of the markers listed in the [paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6659636/). If we did add any, it should be as (‘has plasma membrane part' some [appropriate PRO ID]), but probably avoiding equivalences? @dosumis @addiehl please.

paolaroncaglia:
Note for self and FYI @pnejad: 
I'll add 'fetal hepatobiliary progenitor cell' as soon as the CL editors resolve an issue with merging CL files.
I added the question about markers for discussion at the next CL meeting. Depending on resolution, I might add axioms to the new term, but I'll aim to create the new term first and add details later if I can't do both at once.

dosumis:
Are markers required to distinguish from other, similar cell types?

We probably don't have sufficient logic available for an equivalentClass definition if we need to define wrt developmental potential.

paolaroncaglia:
@dosumis 

> Are markers required to distinguish from other, similar cell types?

Not in this case, no. I noticed some mentions of (potentially not even unique) markers in the paper, hence my question. I'll add the term without markers - better anyway as they could be of limited use in this case. Thanks!

> 
[...]

PR #964, opened 2021-02-25
Add CL_0009020 (fetal hepatobiliary progenitor cell)

paolaroncaglia:
Addresses #840


paolaroncaglia:
@dosumis 
@cmungall 
Asking for your opinion please on whether "part of some liver bud" is appropriate (it was the best I could find).
Also, is there any way to indicate that the synonym "HHyP cell" is an abbreviation? That way I'd feel more comfortable in having it as an exact synonym.
Thanks,
Paola

dosumis:
May be borderline whether 'liver bud' is suitable.  Wikipedia has this has liver bud as 'reabsorbed in tubular structures of the septum transversum by the eighth week. [2]'.  Found this for ductal plate:

image - in https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5385778/ 

From these two sources, it sounds like ductal plate may be a later structure than liver bud.  Either way, it should be added to Uberon.

If adding term is urgent then I think OK for now as long as we have an uberon ticket for 'ductal plate' and another CL ticket to update this term when the more accurate Uberon term is available.

Agree would be good to have standard way to record that a synonyms in an abbreviation using synonym type.  Is there any precedent in Uberon for this?

paolaroncaglia:
Note for self, I'll use UBERON:8410003 ductal plate (created today but not released yet).

matentzn:
- [ ] I resolved conflicts here, @paolaroncaglia please review again (I used an id from my own range because your ID was already used for something else).
- [ ] Paola assign someone as `Assignee`

paolaroncaglia:
Hi @matentzn ,

>     * [ ]  I resolved conflicts here, @paolaroncaglia please review again (I used an id from my own range because your ID was already used for something else).
> 
>     * [ ]  Paola assign someone as `Assignee`

I'm really sorry, looks like my old PR was a duplicate and I had already committed the relevant edits since. I forgot to close it. I usually am quite careful but this one slipped. Apologies Nico, I'll double-check now and close as necessary.

Issue #977, opened 2021-03-09
[NTR] esophagus nonkeratinized squamous epithelial cell


mkeays:
Note - please check that the term does not already exist (check OLS: https://www.ebi.ac.uk/ols/ontologies/cl)

**Preferred term label:**
esophagus nonkeratinized squamous epithelial cell

**Synonyms**
esophageal nonkeratinized squamous epithelial cell
nonkeratinized squamous epithelial cell of esophagus
oesophageal nonkeratinized squamous epithelial cell
nonkeratinized squamous epithelial cell of oesophagus

**Definition (free text, please give PubMed ID in format PMID:XXXXXX)**
Any cell that is part of the esophageal nonkeratinizing stratified squamous epithelium. The esophagus, which requires flexibility to accommodate swallowing of a bolus, is covered with a nonkeratinizing epithelium. PMID:11694559

**Parent cell type term (check the hierarchy here https://www.ebi.ac.uk/ols/ontologies/cl):**
[Nonkeratinizing barrier epithelial cell](https://www.ebi.ac.uk/ols/ontologies/cl/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FCL_0000238&viewMode=PreferredRoots&siblings=false)

**What is the anatomical structure that the cell is a part of? Please check Uberon: https://www.ebi.ac.uk/ols/ontologies/uberon**

**Your nano-attribution (ORCID)**

**Any additional notes or concerns**
CL already has [keratinized squamous cell of esophagus](https://www.ebi.ac.uk/ols/ontologies/cl/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FCL_0002221&viewMode=PreferredRoots&siblings=false) but I think this may be incorrect -- the reference that is listed in the definition only seems to say that esophagus squamous epithelium is nonkeratinizing so possibly this is a typo.

bvarner-ebi:
Hi @mkeays, it appears that the linked PR has been closed. Can we close this ticket?

PR #978, opened 2021-03-09
adding esophagus nonkeratinized squamous epithelial cell (#977)

mkeays:
Added a new cell type based on pathologist's information, as well as info in https://academic.oup.com/jncimono/article/2001/29/7/911372 . 

Also, this paper is cited for [keratinized squamous cell of esophagus](https://www.ebi.ac.uk/ols/ontologies/cl/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FCL_0002221&viewMode=PreferredRoots&siblings=false) but from what I understand this paper says that esophagus squamous epithelium is nonkeratinized so maybe this term is incorrect?


paolaroncaglia:
Hi @mkeays ,
Tricky one. My understanding from a quick literature scan is that esophageal epithelium is keratinized in e.g. herbivores and mouse, but not in human ("Although mouse esophageal epithelium is keratinized and uniformly 4–5 cell layers thick, human esophageal epithelium is nonkeratinized", in [this paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5453881/)). So it would make sense for CL to have both cell types. Unfortunately I don't have access to the full text of the publication referenced by that CL term, and I see that it refers to a "keratinized" FMA term (I thought the FMA subset was human-centric?). I found a few papers describing expression of keratins in human esophageal epithelium, but also evidences that the keratinization is a marker of neoplastic transformation. I'd suggest to accept your new term, and review/improve annotations for the existing keratinized one. Let's see if @dosumis or others have any further feedback.

paolaroncaglia (commented):
Hi @mkeays ,
Tricky one. My understanding from a quick literature scan is that esophageal epithelium is keratinized in e.g. herbivores and mouse, but not in human ("Although mouse esophageal epithelium is keratinized and uniformly 4–5 cell layers thick, human esophageal epithelium is nonkeratinized", in this paper). So it would make sense for CL to have both cell types. Unfortunately I don't have access to the full text of the publication referenced by that CL term, and I see that it refers to a "keratinized" FMA term (I thought the FMA subset was human-centric?). I found a few papers describing expression of keratins in human esophageal epithelium, but also evidences that the keratinization is a marker of neoplastic transformation. I'd suggest to accept your new term, and review/improve annotations for the existing keratinized one. Let's see if @dosumis or others have any further feedback.
(Otherwise your PR looks good to me.)

mkeays:
Thanks @paolaroncaglia , of course I forgot about non-humans again!

mkeays:
I'm updating the definition to make it clear that in humans the esophagus is lined with non keratinizing squamous epithelium; I thought to also add `present in taxon` `Homo sapiens` but I can't see this `present in taxon` property when I look at Cell Ontology in Protege. Is it enough to just have it in the definition?

paolaroncaglia:
@mkeays 
@dosumis 
cc @cmungall for taxon-related issues

> I'm updating the definition to make it clear that in humans the esophagus is lined with non keratinizing squamous epithelium; I thought to also add `present in taxon` `Homo sapiens` but I can't see this `present in taxon` property when I look at Cell Ontology in Protege. Is it enough to just have it in the definition? I see `only in taxon` in the `SubClass Of` section but I don't want to use that in case there are other taxa with it that I don't know about; there is also `never in taxon` in the list of Annotation Properties.

Maria is right, there's no annotation property 'present in taxon' in CL. I suppose that the definition gloss may suffice, but in general, is there a plan to make taxon-related properties more consistent between Uberon and CL please?

Maria, I noticed you also changed the label, but I'd suggest to keep the wordings "nonkeratinized" and "non-keratinized"; you could make them exact synonyms if you prefer to avoid them in the primary label. They are used in the literature, plus some text mining tools may mistake e.g. "non keratinized" for "keratinized" if they've been trained to avoid common text strings such as "non".

mkeays:
Thanks @paolaroncaglia , I'll change it back to "non-keratinized" and add "non keratinizing" as a synonym. I think I did it to follow other classes in CL already, but it does seem in this case "non-keratinized" is used more often. Good to know about the text mining issue, I hadn't thought of that!

mkeays:
Have made this change, ended up adding quite a lot of synonyms due to different spellings of esophagus and presence/absence of hyphenation! Maybe this is too many/not necessary, can remove some if it seems excessive!

paolaroncaglia:
Note for self/David/Maria: the annotation property 'present in taxon' is now in CL and can be used as appropriate.

paolaroncaglia:
I'll close and re-do (see #1463 ).

PR #1036, opened 2021-04-19
Added morphological subclasses

shawntanzk:
*** still at drafting stage, do not merge yet ***

added:
martinotti L2/3 PV-like SST (Mus)
basket L2/3 PV-like SST (Mus)
stellate L6 IT MOp (Mus)
inverted L6 IT MOp (Mus)


shawntanzk:
Done in separate PRs

PR #1063, opened 2021-05-04
Arachnoid barrier cell

shawntanzk:


src/ontology/cl-edit.owl
```diff
@@ -28105,6 +28106,16 @@ SubClassOf(obo:CL_4023065 ObjectSomeValuesFrom(obo:RO_0002100 obo:UBERON_0000956
 SubClassOf(obo:CL_4023065 ObjectSomeValuesFrom(obo:RO_0002162 obo:NCBITaxon_10090))
 SubClassOf(obo:CL_4023065 ObjectSomeValuesFrom(obo:RO_0002292 obo:PR_P97367))
 
+# Class: obo:CL_4023066 (arachnoid barrier cell)
+
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref "PMID:23298861") obo:IAO_0000115 obo:CL_4023066 "A mesothelial fibroblast of the leptomeninx that makes up the tight-junctioned layer that functions as the physiologic barrier between the cerebrospinal fluid in the subarachnoid space and the fenestrated capillaries in the dura.")
```
patrick-lloyd-ray, line 28111: Looks fine to me, but two small things: 

1. I prefer 'constitutes' rather than 'makes up' as the former is more precise. What do you think? 
2. Consider whether using 'functions as' aims at the proper meaning. I think functions are pretty tricky, so maybe 'forms' (or something like that) is closer to the intended meaning. I understand that it is the layer (arachnoid mater) that has the function (not the cell), but maybe that attribution of function should be in 'arachnoid mater', not 'arachnoid barrier cell'. 

Neither of these are necessary changes, but just considerations.

patrick-lloyd-ray (commented):
in-line

src/ontology/cl-edit.owl
shawntanzk: 1. Agree, constitutes is defs more precise and does sound better too
2. I get what you mean, when i revisited this, it took me a second read to remember what I meant with that too (its the layer that is the barrier not the cell). I considered removing it, but I think its useful to have for end user. Definition for arachnoid mater does not really cover this point which I think is an important one given the "barrier" in the cell name. I think changing it to forms would help.

patrick-lloyd-ray: Fantastic! Thanks, Shawn!

lubianat: My 2 cents: what about "mesothelial fibroblast of the arachnoid barrier layer" with a part of arachnoid barrier layer axiom (http://purl.obolibrary.org/obo/UBERON_0000437)?  

Maybe extra information could go  by   ‘editor  note’  (IAO_0000116), as per http://revista.ibict.br/ciinf/article/view/4015

shawntanzk: Thanks for the suggestions :) I think youre right that it should be part of arachnoid barrier layer, however, I hesitate to move the other information into the editor note or comments as without it, the definition would, in my opinion, be less useful to the end user. Furthermore, I think the current information in the def helps explain the part of relationship - in that it forms the the tight-junction layer.

lubianat: @shawntanzk good point. Another option (if desired) would be  to actually keep the info on the definition, but write details after a first full stop, in the concise definition + gloss pattern (as per https://douroucouli.wordpress.com/2019/07/08/ontotip-write-simple-concise-clear-operational-textual-definitions/).

shawntanzk: I think a def + gloss for this is a good idea, maybe something like: 

"A mesothelial fibroblast of the arachnoid barrier layer. Arachnoid barrier cells make up the tight-junctioned layer in the leptomeninx that functions as the physiologic barrier between the cerebrospinal fluid in the subarachnoid space and the fenestrated capillaries in the dura."

dosumis (approved):

PR #1078, opened 2021-05-10
Improved CL:0000118 'basket cell'

shawntanzk:
***DRAFT - DO NOT MERGE***

Fixes #1077

will make equivalent class with 

```
Interneuron
and ('capable of' some 'gamma-aminobutyric acid secretion, neurotransmission')
and ('bearer of' some 'basket morphology') 
```
when Basket morphology gets pulled in from PATO

PR #1079, opened 2021-05-10
added some PATO morphology annotations

shawntanzk:
***DRAFT - DO NOT MERGE***

To add:
Bipolar Morphology for Bipolar Neuron (make equivalent to)

Notes: 
Basket cell will be done on PR #1078

When making this PR, I noticed that cerebral cortex GABAergic interneuron (CL:0010011) does not fall under CNS interneuron (CL:0000402) with reasoner

I noticed that Pyramidal neuron def might be wrong and have added an editors note (instead of changing it yet): "Not all pyramidal neurons have apical dendrite pointed toward the pial surface. Horizontal pyramidal cells have an apical tree which is oriented parallel to the pia."


patrick-lloyd-ray (approved):

shawntanzk:
Continued in PR #1108

PR #1095, opened 2021-05-17
Updated pyramidal neuron

shawntanzk:
See PR #1079

Also added horizontal pyramidal neuron, happy to add others (those under standard pyramidal morphology) if need be. @dosumis - let me know if I should


shawntanzk:
I've also pushed Betz cell as a CNS neuron instead (reasoner places it correctly with new edits)

patrick-lloyd-ray (approved):

PR #1098, opened 2021-05-17
Fixed L2/3/5 Fan martionotti SST

shawntanzk:


dosumis (changes requested):
Looks good, except that Martionotti should be capitalised (as it is someone's name), but 'fan' should not.  Please apply this consistently to other terms too.

shawntanzk:
> Looks good, except that Martionotti should be capitalised (as it is someone's name), but 'fan' should not. Please apply this consistently to other terms too.

whoops, sorry about that - noticed it when I was fixing but didn't catch it all - fixed now. Thanks

dosumis (approved):

PR #1129, opened 2021-06-21
Release 2021-06-21

shawntanzk:
New Terms
•	stromal cell of lamina propria of vermiform appendix
•	appendix macrophage
•	horizontal pyramidal neuron
•	Martinotti morphology L2/3 Pvalb-like Sst GABAergic cortical interneuron (Mus musculus)
•	medial ganglionic eminence derived GABAergic cortical interneuron
•	caudal ganglionic eminence derived GABAergic cortical interneuron
•	L5/6 CCK expressing cortical GABAergic interneuron (Mus musculus)
•	Martinotti neuron
•	bitufted neuron
•	basket morphology L2/3 Pvalb-like Sst GABAergic cortical interneuron (Mus musculus)
•	stellate L6 intratelencephalic projecting glutamatergic neuron of the primary motor cortex (Mus musculus)
•	inverted L6 intratelencephalic projecting glutamatergic neuron of the primary motor cortex (Mus musculus)
•	chandelier cell
•	T Martinotti neuron
•	fan Martinotti neuron
•	large basket cell
•	nest basket cell
•	small basket cell
•	inverted pyramidal neuron
•	stellate pyramidal neuron
•	tufted pyramidal neuron
•	untufted pyramidal neuron

Edits
•	Improved basket cell entry (adding def and axioms)
•	Improved stellate neuron entry (adding def and axioms)
•	Edited pyramidal neuron definition 
•	Edited stellate interneuron axioms
•	Improved neurogliaform cell (edited def and axioms)
•	Added synonyms for brush cell of large intestine 
•	Edited lymphoid tissue-inducer cell definition 
•	Added PATO axioms for various neural cells
•	Improved A7 cell (edited def and axioms)


cl.obo
```diff
@@ -1484,7 +1485,7 @@ relationship: capable_of GO:0002495 ! antigen processing and presentation of pep
 relationship: capable_of GO:0002540 ! leukotriene production involved in inflammatory response
 relationship: capable_of GO:0045730 ! respiratory burst
 relationship: develops_from CL:0000774 ! band form eosinophil
-relationship: has_part GO:0098537 ! lobed nucleus
+relationship: has_part GO:0098537 ! obsolete lobed nucleus
```
matentzn, line 1488: Using an obsolete class, should make a ticket to fix

matentzn (approved):

cl.obo
dosumis: Quick fix: The 3 places where this is used, you can substitute a compound entity (anonymous class expression) to say the same thing:
'has part some 'lobed nucleus'
->
'has part' some (nucleus and 'bearer or' some lobed) 
# lobed = PATO:0001979 (edited) 

@matentzn - do we need to fix before release?

matentzn: I dont really mind. It probbaly happens often, but we really should generally not use obsolete classes!

dosumis:
I'm still uncomfortable with the Pvalb-like classes.  

rdfs:label "L2/3 Pvalb-like Sst GABAergic cortical interneuron (Mus musculus)"
definition "A Sst GABAergic cortical interneuron transcriptomically between the Sst and the Pvalb families with a soma found in L2/3. Some L2/3 PV-like SST cells have Martinotti morphology while some have basket cell morphology. L2/3 PV-like SST cells have lower AP width and higher firing rate than typical Sst subclass cells. L2/3 PV-like SST cells have high maximum firing rate variability, with many classified as belonging to the Pvalb subclass on the basis of electrophysiology."

1. There's not really any way for anyone outside of the BDS project to use these terms for annotation.  How will they know when they've found a sufficiently Pvalb-like SST GABAergic to qualify?  There's not even an data linked to the term that could be used to project annotations. At least the region specific ones (in BDSO) come with both data and distinguishing markers.  
2. The definitions read very much like a case of 'we have a heterogenous cluster in an scRNAseq analysis - here's a summary of various properties of types in it.'  

In the interests of getting a release out, I'm happy to let them through.  But we should make a ticket proposing obsoletion and remove them ASAP.  

For CL, I think it will be sufficient to have 

L2/3 SST basket neuron
L2/3 SST Martinotti neuron

Looking at how easy it was to simplify the names of these (assume all basket and Martonotti are GABAergic), I think we should open a ticket to review long names of recently added neuron types & shorten/simplify where possible.

dosumis (approved):
Approved - but see conversation for fixes that should be prioritised after release.

shawntanzk:
@dosumis will make a ticket on BDS ontology - think it fits there better for now. Will create tickets based on that on CL when appropriate

cl.obo
dosumis: Do we have a warning system in place for obsoletion imported terms?

matentzn: Unfortunately this is one of these few cases where we cant easily have a check, because there are too many errors in cl.owl, which means we restrict error checking to cl-edit.owl, but if we look at cl-edit.owl in isolation, we don't know which external classes are obsolete. I have made the robot report checking highly configurable now, but what I cant yet do is assign different error levels to different ontology flavours (cl.owl should have different things being flagged as an ERROR as cl-edit.owl etc because it includes external axioms).In any case, not easy. The most straight way is to do something like this: https://github.com/obophenotype/cell-ontology/pull/1132/files

Its hacky, and the problem deserves a deeper luck, but perhaps that sufficies for now.

PR #1132, opened 2021-06-22
Add mega-hacky check for obsoletes

matentzn:
see #1129


matentzn:
- [ ] Make sure test passes
- [ ] Merge

Issue #1124, opened 2021-06-16
[NTR] anorectum macrophage


sfexova:
Note - please check that the term does not already exist (check OLS: https://www.ebi.ac.uk/ols/ontologies/cl)

**Preferred term label:**
anorectum macrophage

**Synonyms**
macrophage of anorectum

**Definition (free text, please give PubMed ID in format PMID:XXXXXX)**
Macrophage that is located in the anorectum.

**Parent cell type term (check the hierarchy here https://www.ebi.ac.uk/ols/ontologies/cl):**
macrophage http://purl.obolibrary.org/obo/CL_0000235

**What is the anatomical structure that the cell is a part of? Please check Uberon: https://www.ebi.ac.uk/ols/ontologies/uberon**
anorectum UBERON:8410050

**Your nano-attribution (ORCID)**
0000-0003-3440-1876

**Any additional notes or concerns**

paolaroncaglia:
Note to self: can't create this term as desired until Uberon releases 'anorectum'. Should be any day now (David said possibly today).

paolaroncaglia:
@sfexova 
CL_0009048 (anorectum macrophage)

PR #1169, opened 2021-07-07
Add CL_0009048 (anorectum macrophage)

paolaroncaglia:
Addresses #1124


paolaroncaglia:
Hi @shawntanzk and/or @dosumis ,
Could you please double-check the diff to make sure that Shawn's term 'arachnoid barrier cell' is not affected. To me it looks like a non-diff (change of place in file?), but just in case. Thanks!

shawntanzk:
@paolaroncaglia looks alright to me, just a change in location, probably just cause I had to change the IRI cause of a clash and didnt reorder it to the correct location - will do that in the future :) thanks for checking!

shawntanzk (approved):

Issue #1116, opened 2021-06-16
[NTR] B cell of anorectum


sfexova:
Note - please check that the term does not already exist (check OLS: https://www.ebi.ac.uk/ols/ontologies/cl)

**Preferred term label:**
B cell of anorectum

**Synonyms**

**Definition (free text, please give PubMed ID in format PMID:XXXXXX)**
B cell that is located in the anorectum. B cells are at the centre of the adaptive humoral immune system and are responsible for mediating the production of antigen-specific immunoglobulin (Ig). PMID: 31002794

**Parent cell type term (check the hierarchy here https://www.ebi.ac.uk/ols/ontologies/cl):**
B cell http://purl.obolibrary.org/obo/CL_0000236

**What is the anatomical structure that the cell is a part of? Please check Uberon: https://www.ebi.ac.uk/ols/ontologies/uberon**
UBERON:8410050 anorectum

**Your nano-attribution (ORCID)**
0000-0003-3440-1876

**Any additional notes or concerns**

paolaroncaglia:
Can't create as desired until Uberon releases 'anorectum'.

paolaroncaglia:
@sfexova 
CL_0009050 (B cell of anorectum)

PR #1174, opened 2021-07-16
DO NOT MERGE - Add CL_0009049 'B cell of anorectum'

paolaroncaglia:
Addresses #1116


paolaroncaglia:
CL editors: 
Similar to what @lubianat observed recently, there's a huge number of "^^xsd:string" added that make reviewing a pull request a daunting prospect. What happened that made the number of those additions explode, and how can we fix it please? (Mock-editing cl-base works, but only temporarily.) 
Thanks,
Paola

dosumis:
It seems to be an OWL-API bug and it happens randomly, usually hitting just a few terms, but occasionally blowing up to the point where there are hundreds and the PR becomes very hard to review.

There's a ticket on this, but I'm not sure there's any progress.  @matentzn - maybe we should push on this again?

dosumis:
Found the Protege ticket I submitted on this: https://github.com/protegeproject/protege/issues/993.

All discussion seems to be on Slack:
https://obo-communitygroup.slack.com/archives/C01BFAZPQ86/p1603183433006700 (fortunately not lost down the memory hole yet) - the idea that this might be an OWL-API bug is discussed in this thread.  @matentzn - any idea how best to proceed?

matentzn:
I have reached out to Ignazio and Matthew to see whether they have an idea. It's so annoying we can't replicate it this reliably.. in any case, solving this will take ages, because even if solved in OWLAPI, it needs new robot and protege release to permeate through. One idea: can you check wether an import adds a range constraints on the ap that is blowing up? I am on lunch  break now but will investigate more later. 

If nothing works, we will start having ontobot post diffs on new pull request; or revert to obo as the base serialisation - ahhhhhhh!

paolaroncaglia:
Thanks @dosumis . I'm concerned that the "occasional" explosive occurrence seems to have become the norm in the last week or so. 
If the bug is with Protege, it may not be fixed and released quickly enough. Should we think of a workaround based on Tiago's diagnosis that 
"Protege seems to have added automatically ^^xsd:string to all existing oboInOwl:hasDbXref "GOC:tfm""
maybe replace GOC:tfm with Terry's ORCID?
As you say the bug seems random and as far as I know we've never seen it co-occur with anything other than GOC:tfm...

paolaroncaglia:
If it's any help: searching for "GOC:" in CL in Protege, 
a) GOC:tfm is not the only GOC dbxref
b) the xsd:string thing is appended to many, but not all, terms that have a GOC: dbxref, and this is irregardless of ontology space (GO, CL, ...). 
Can't find a pattern as to why only some of Terry's terms are affected.

dosumis:
* This happens sometimes in other ontologies too (e.g. in EFO).  
* It happens at some level in almost every PR - so shouldn't be too hard to replicate.
* I've only ever seen it happen to database_cross_reference axioms when used to annotate an annotation axioms (e.g. a definition), but I suspect it is a general bug hitting annotations on annotation axioms (or maybe annotations on any axiom) and we're just seeing it hit these because they are by far the most common.  There seem to be some serious issues with the spec for this (see Ignazio's rant here: https://github.com/owlcs/owlapi/issues/874#issuecomment-525905187).
* There is no sign of range constraints on database_cross_reference in CL, so I think that's a red herring.

dosumis:
> If nothing works, we will start having ontobot post diffs on new pull request

I like this idea

matentzn:
Ready for review. There were indeed spurious diffs on master, chasing down the culprit now.

matentzn:
- [ ] @dosumis to review

paolaroncaglia:
@matentzn I'll close and re-do, this was mostly just a test for the xsd:string issue, hence the "do not merge". Thanks again.

PR #1182, opened 2021-07-20
Fibroblasts are not epithelial cells!

kiri93:
Two changes:
- Fibroblasts are not epithelial cells! As such, I removed the "is_a" relation to CL:0000066 ! epithelial cell
- For consistency with other terms, e.g. "fibroblast of papillary layer of dermis" I changed "reticular layer of dermis fibroblast" to "fibroblast of reticular layer of dermis"

### DIFF
--- a/cl-basic.obo
+++ b/cl-basic.obo
@@ -23676,10 +23676,9 @@ creation_date: 2015-05-26T18:58:34Z
 
 [Term]
 id: CL:2000096
-name: reticular layer of dermis fibroblast
+name: fibroblast of reticular layer of dermis
 namespace: cell
 def: "Any fibroblast that is part of a reticular layer of dermis." [GOC:TermGenie]
-is_a: CL:0000066 ! epithelial cell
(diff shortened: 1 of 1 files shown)


dosumis (approved):
Yes- that's definitely an error.  Many thanks for fixing.
- David

dosumis:
Hi @kiri93 - I've just realised that this edit was to the wrong file. cl-basic.obo is a release file and so your change will be over-written and the next release.  The editor's file is: src/ontology/cl-edit.owl.   Would you be able to make the changes there?

Many thanks again, 
David

kiri93:
Ah yeah sorry, I'm still quite new to all of this. I'll change the src/ontology/cl-edit.owl file!

kiri93:
Apparently, the relation `reticular layer of dermis fibroblast "is_a" epithelial cells` is not hard-coded in `src/ontology/cl-edit.owl`. I believe it could be related to the annotation `UBERON_0001993 (reticular layer of dermis)` which is `"part_of" epidermal-dermal junction`.

PR #1188, opened 2021-07-21
Adding normalisation documentation

shawntanzk:


matentzn:
Do you want to try and edit the navigation as well?

https://github.com/obophenotype/cell-ontology/blob/master/mkdocs.yaml

shawntanzk:
> Do you want to try and edit the navigation as well?

Done. Thanks! https://github.com/obophenotype/cell-ontology/blob/Adding-normalisation-documentation/mkdocs.yaml

matentzn (approved):
I hope the ^^ in the navigation is not a problem but shouldn't.

shawntanzk:
> I hope the ^^ in the navigation is not a problem but shouldn't.

was thinking that too, but github didn't flag it so i assumed .yaml would be ok with it. Will see, if theres errors, I'll change the title to something else

PR #1193, opened 2021-07-22
Make xsd:string normalisation work with MacOS's sed.

gouttegd:
A difference between GNU sed and MacOS (or BSD) sed is that the latter
expects a *mandatory* argument after the -i option (that argument is the
suffix to use for creating a backup file), whereas that argument may be
omitted with GNU Sed.

A consequence of that difference is that invoking sed as follows:

```
$ sed -i -E ...
```

will work as expected with GNU sed, but with MacOS sed, the -E option
will be interpreted as the argument expected by the -i option. As a
result, the regular expression that follows is interpreted as a "basic"
regular expression instead of an "extended" one.

This makes the xsd:string normalisation rule fail on MacOS with the
following error:

```
sed: 1: "s/Annotation[(](oboInOw ...": \1 not defined in the RE
make: *** [normalise_xsd_string] Error 1
```

because '(' characters are interpreted differently between basic and
extended regular expression.

This patch makes sure an argument is provided to the -i option, so that
the normalisation step can be used portably and without requiring the
use of Docker for a simple sed call.

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -10190,7 +10190,7 @@ SubClassOf(Annotation(oboInOwl:is_inferred "true"^^xsd:string) obo:CL_0000787 ob
 
 # Class: obo:CL_0000788 (naive B cell)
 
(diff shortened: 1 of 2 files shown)


matentzn:
Awesome tested it, works!

PR #1219, opened 2021-08-09
Imported Callithrix jacchus

shawntanzk:
@matentzn not sure if this is safe to commit - I imported term to ncbitaxon without updating ncbitaxon for now, if it's ok, could you help me give this a pass and ill merge it down and start working on terms that need this.
not a huge rush, so if you'd prefer if I waited for a fix before doing this, happy to delete this PR too and do it when time its ready.
Thanks! :)


matentzn (approved):

shawntanzk:
massive clashes when updating from master due to new release - will close this and redo this instead

PR #1220, opened 2021-08-10
2021-08-10 release

matentzn:
@shawntanzk 

your comments you made about broken cl-full.obo - I think this was always true. I did fix some duplicate definitions, but many not. Did you want the terms in this PR in as well?


shawntanzk:
@matentzn this branch looks ok :) checked a few entries that were duplicated before and it seems fixed here. Thanks!

shawntanzk:
Clashes -> will make a new release PR

PR #1227, opened 2021-08-10
Uberon import UBERON:8410037

shawntanzk:
testing to see if this import passes check as I can't seem to find the error causing items in uberon import


shawntanzk:
just using this as a check

PR #1237, opened 2021-08-25
Added subtypes for spiral ganglion neuron

### LINKED ISSUE #163: Add subtypes for spiral ganglion neuron
```
https://en.wikipedia.org/wiki/Spiral_ganglion#Anatomy

Type I:  innervate the inner hair cells. myelinated, bipolar.
Type II: innervate the outer hair cells. unipolar and unmyelinated in most 
mammals. 
```

Original issue reported on code.google.com by `cmung...@gmail.com` on 21 Dec 2014 at 12:35

shawntanzk:
Fixes #163
@cmungall added you as reviewer as issue has your email on it


shawntanzk:
Note: did some massive removal of obsolete term GO:0044464 including in cl.obo (which somehow doesn't show here) - not sure if this is ok

matentzn:
When removing axioms with obsolete classes, you should check the metadata on these classes for potential replacements, like here: http://www.ontobee.org/ontology/GO?iri=http://purl.obolibrary.org/obo/GO_0044464

If I understand this correctly (ignore syntax just guessing now),

`isa: obsolete cell part`

should be replaced by

`relationship: part of cell`

shawntanzk:
uberon import fixed - ready for review

cmungall (approved):
Looks good. Judging by the text definition I would be tempted to make a logical def, but first a DP for cell types differentiated by what they innervate. But this could always be done after - maybe make an issue? @dosumis is often more careful than I am making logical defs

PR #1239, opened 2021-08-25
Import fix

shawntanzk:


matentzn (approved):
Great thanks! Please remember to also fix this upstream at PATO!

matentzn:
QC is failing for some reason!

shawntanzk:
> QC is failing for some reason!

ironically oort error, looks like the same issue with the RO term?

matentzn:
i wonder whether the uberon import was actually necessary for the fix.. Maybe uberon.owl has the duplicate definitions?

shawntanzk:
just looked at my history, it seems uberon_import.owl change was all just blank lines, but it might be a previous error that was not caught till this check was added?

Issue #38, opened 2015-06-04
Add synonym to CL:0000020 spermatogonium


GoogleCodeExporter:
```
Hi,

Could CL:0000020 spermatogonium
please have a synonym 'spermatogonial cell'
to parallel GO terms such as 'spermatogonial cell division ; GO:0007284'.

Thanks,
Paola
```

Original issue reported on code.google.com by `pa...@ebi.ac.uk` on 8 May 2013 at 2:20

paolaroncaglia:
I'll take this (as I requested it myself in 2013)

PR #1246, opened 2021-09-02
Add synonym to CL:0000020 spermatogonium

paolaroncaglia:
Addresses #38


paolaroncaglia:
@matentzn 
I successfully ran the [revised command](https://github.com/obophenotype/cell-ontology/blob/master/docs/Fixing_xsdstring_diffs.md) to get rid of xsd:strings, but one persists. May I merge anyway please?
Thank you.
Paola

matentzn (approved):

matentzn:
The change is a correct one! Adding xsd:string is always acceptable!

PR #1261, opened 2021-09-17
Trying to switch GO to GO Plus

matentzn:


balhoff:
@matentzn you could also use `http://purl.obolibrary.org/obo/go/go-base.owl`.

matentzn:
Does http://purl.obolibrary.org/obo/go/go-base.owl contain all the go-plus axioms?

dosumis:
> Does http://purl.obolibrary.org/obo/go/go-base.owl contain all the go-plus axioms?

Jim confirmed here that it does https://github.com/geneontology/go-ontology/issues/22170

shawntanzk:
just ran a check, insane amounts of duplicate defs from uberon apparently (3831 errors - duplicate def)
Screenshot 2021-10-01 at 16 00 16
seems like the GO import isnt importing database_cross_reference, hence the clashing defs, not sure how to fix this myself unfortunately @matentzn

shawntanzk:
If I run make "imports/go_import.owl MIR=false" it cuts it down to a manageable amount where I can go in and figure stuff out, but it will be relatively manual - seems to have a whole bunch of odd duplications still. But I don't know if the importing defeats the purpose of this to switch to GO Plus?

matentzn:
- [ ] switch go import to use go-base @shawntanzk

matentzn:
hahhahah our FAVOURITE ERROR @shawntanzk Amazing. How is this fair? This is so annoying I wish we could obliterate the OORT pipeline right now..

shawntanzk:
@matentzn kindda expected the errors while fixing - am currently re-importing GO (I assume that's needed), will see after :)

shawntanzk:
@matentzn checks have passed :) hope its all done correctly - would like a second eye on it

matentzn:
You did everything correctly, but using go-plus (or base) blows up the module size by 300K lines. This is to be expected, due to the way that object properties work in the module extraction process. I think the PR is done, but to be very honest, I am not 100% we can merge this without solving the "external module" problem yet (SLME for qc, existential graph for release).

Such a pity! You did everything perfect though, thank you @shawntanzk.

matentzn:
This is the issue I think we need to fix first: https://github.com/INCATools/ontology-development-kit/issues/458

anitacaron:
This ticket will be fixed as part of base migration by @ubyndr  and @matentzn

shawntanzk:
not relevant after migrating to base file

PR #1270, opened 2021-10-04
Added layer axioms

shawntanzk:
*** DRAFT DO NOT MERGE ***
Terms still to add to UBERON
L2/3
L4/5
L5/6
L6b


shawntanzk:
massive duplicated IDs again with uberon import - have to check what is up with this and figure out where the issue is, might try to redo this with MIR=false to see if it's just my mirror issue cause of previous work

shawntanzk:
errors still there - not sure what is causing this as in protege, offending terms duplicated def look exactly the same which should not trigger the break. Need to figure out exactly why there is this problem - will tackle in next Monday technical sprint meeting.

matentzn:
@shawntanzk I now added a way to simply delete offending annotations when `oort` fails. 

Just add the class that causes the problem to src/ontology/config/remove_annotations.txt when oort fails again - I think I removed all failures from the current build, hopefully.

matentzn:
You do however need to reduce the size of the diff of cl-edit if any review needs to be done.. probably the xsd:string code?

shawntanzk:
> just add the class that causes the problem to src/ontology/config/remove_annotations.txt when oort fails again

does this delete all the annotations? lol. 

> I think I removed all failures from the current build, hopefully.

will take a look later today, but thanks so much for that - this PR needs a bit more work that has uberon new terns dependencies anyway, so it won't be merged that soon. 

> You do however need to reduce the size of the diff of cl-edit if any review needs to be done.. probably the xsd:string code?

will look into this too - from quick glance of your commits, can't tell if it is just that or something else, but will normalise and also compare it to master when i get to this.

Thanks so much for helping me with this, really appreciate it!

shawntanzk:
> probably the xsd:string code?

yeah it is this - fixed :)

src/ontology/cl-edit.owl
```diff
@@ -24301,7 +24301,7 @@ SubClassOf(obo:CL_1000345 ObjectSomeValuesFrom(obo:BFO_0000050 obo:UBERON_000198
 
 # Class: obo:CL_1000346 (enterocyte of epithelium proper of large intestine)
 
-AnnotationAssertion(Annotation(oboInOwl:hasDbXref "GOC:tfm"^^xsd:string) obo:IAO_0000115 obo:CL_1000346 "An enterocyte that is part of the epithelium proper of large intestine."^^xsd:string)
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref "GOC:tfm"^^xsd:string) obo:IAO_0000115 obo:CL_1000346 "MERGED DEFINITION: TARGET DEFINITION: A columnar cell of the colon that is part of the epithelium proper of large intestine. SOURCE DEFINITION: An enterocyte that is part of the epithelium proper of large intestine."^^xsd:string)
```
shawntanzk, line 24304: @matentzn not sure what this is, I assume this is part of the fix?

paolaroncaglia: @shawntanzk I did that edit yesterday (cleaned up def of CL_1000346). Was the edit reverted? Thanks.

shawntanzk: the def now should contain the whole string "MERGED DEFINITION: TARGET DEFINITION: A columnar cell of the colon that is part of the epithelium proper of large intestine. SOURCE DEFINITION: An enterocyte that is part of the epithelium proper of large intestine." - If possible could you fix this in a separate PR to whichever the correct def is please? Thanks

shawntanzk: just to be 100% sure, the correct def is "An enterocyte that is part of the epithelium proper of large intestine." right?

src/ontology/cl-edit.owl
```diff
@@ -3168,7 +3168,7 @@ SubClassOf(obo:CL_0000028 obo:CL_0000540)
 SubClassOf(obo:CL_0000028 ObjectSomeValuesFrom(obo:BFO_0000050 obo:UBERON_0001017))
 SubClassOf(obo:CL_0000028 ObjectSomeValuesFrom(obo:RO_0002202 obo:CL_0000338))
 
-# Class: obo:CL_0000029 (neural crest derived neuron)
+# Class: obo:CL_0000029 (neuron neural crest derived)
```
shawntanzk, line 3171: this is super weird, I'll fix this, but no idea why this is popping up

shawntanzk:
@dosumis with layer information - should I make equiv classes (think it should be quite safe)

shawntanzk:
Closing and redoing - easier than figuring out conflicts

PR #1284, opened 2021-10-11
Add a workflow to automatically close stale issues.

gouttegd:
As discussed in the technical board meeting of October 11th 2021, this PR does for CL the same thing as obophenotype/Uberon#2030 does for Uberon: it proposes a mechanism to prevent the accumulation of old issues by regularly and automatically closing them.

Parameters used here are the same as in the Uberon repo, that is:
* an issue is to be considered "stale" after 6 months of inactivity;
* a stale issue is to be closed after 1 year of inactivity.

### DIFF
--- a/.github/workflows/stale-issues.yml
+++ b/.github/workflows/stale-issues.yml
@@ -1,0 +1,18 @@
+name: 'Mark and close stale issues'
+
+on:
+  schedule:
+    - cron: '0 1 * * *'
+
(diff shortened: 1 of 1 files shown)


matentzn:
@shawntanzk please raise at next editors call, I very much agree with this.

shawntanzk:
got thumbs up - will merge

PR #1289, opened 2021-10-20
Update ODK

shawntanzk:
@matentzn CL seems to already be on v1.2.30, but there are still a few changes - either up updated the ODK here


matentzn:
in src/ontology/uberon-odk.yaml, you need to set a new parameter:

```
git_main_branch: master
```

to make sure the docs are correct,


then regenerate.

shawntanzk:
refresh imports works well - will merge once CI checks pass

Issue #1279, opened 2021-10-07
[NTR] Add CL terms for thymus ASCT+B table


paolaroncaglia:
cc @dosumis and @emquardokus 
As discussed at yesterday’s ASCT+B harmonisation meeting, I’ll be adding CL terms needed for the thymus ASCT+B table.
Ellen has collected information and suggestions from experts in a spreadsheet.
- [x] I’ll review to address any potential issue (e.g. should any term go in Uberon instead; remove redundant superclasses). 
- [x] I’ll add the annotation property ‘present in taxon’ ‘Homo sapiens’. (Done for all terms except the 3 DN thymocytes)
- [x] As these terms are immune-related, I might ask Alex to review my PRs (pointing him to the spreadsheet, mentioning that the gene + protein biomarkers will be in HubMap, not CL, and asking if he'd like to suggest synonyms to aid in text searches). 
- [x] I'll add the new CL IDs in the spreadsheet when committed (already done for the first 12 terms from [this PR](https://github.com/obophenotype/cell-ontology/pull/1297/commits/1f5757ca1cb2845a7ad816ef5b743dfb650bbd51)).
- [x] Push for a CL release?

For reference, the spreadsheet is [here](https://docs.google.com/spreadsheets/d/1BihrOU9vWrsA_wYjcbdPUr42_Sb22CO3gkBDieiYQPs/edit#gid=725346151), and the broader ticket about ASCT+B thymus anatomy work is [here](https://github.com/obophenotype/CCF_tools/issues/29).

paolaroncaglia:
Note for self: following up on double negative thymocyte terms, here are the relevant minutes from the May 19th 2021 CL eds meeting: 
"Taxon constraints (again)
We now have better doc, but some guidance is needed on when to add taxon-specific classes:
General agreement that taxon specific general classes with no taxon specific differentia are out, e.g. we don't add "lymphocyte, human"
We do add taxon specific classes where definition require taxon specific marker 
Are there cases in between these two?
Examples from immune system and brain.
In this case we could reasonably have a more species general term (+ a present in taxon link). 
Note - we have some inconsistency in naming here - "(Mus musculus" vs ", mouse" used elsewhere - but common names are harder to support with automated naming from patterns using NCBI taxonomy.)"
[...]

PR #1297, opened 2021-10-28
 [NTR] Add CL terms for thymus ASCT+B table #1279

paolaroncaglia:


paolaroncaglia:
Hi @addiehl (cc @emquardokus and @dosumis ),
I'll commit these thymus cell types now. If there's any major objection, we can address later. Thanks.

PR #1299, opened 2021-10-28
Added displaced amacrine cell

shawntanzk:
Fixes #380
Currently blocked as PR import is stopped due to insufficient memory
Still need to add expresses some Sox2 (PR:000015426)

PR #1307, opened 2021-11-11
DO NOT MERGE Added 3 human DN terms for #1279

paolaroncaglia:


paolaroncaglia:
@dosumis and/or any other kind CL editor soul
Could you please look into this failed check? I've spent the last 2 hours trying to make some seemingly innocent edits, but I've run into issues - first an equivalent to owl:nothing that I couldn't compute explanations for (Protege got hung multiple times) and apparently due to addition of a part_of subclass; then I didn't include the part_of, reasoner was happy, but checks fail and I can't see why from the (huge) log. May be a trivial error from my side, but I thought I was following the same ontology practice as always... thanks

paolaroncaglia:
P.S. This PR is to complete addition of thymus cell type terms for ASCT+B work

dosumis:
Looks like this is causing obo check fail:

         --> in frame CL:0000808
             duplicate "comment" clauses
         --> in frame CL:0002423
             duplicate "comment" clauses
         --> in frame CL:0002424
             duplicate "comment" clauses
         --> in frame CL:0000806
             duplicate "comment" clauses

dosumis:
Re - huge log, I suspect someone has set Robot to verbose.  @shawntanzk - have you been working on this?

paolaroncaglia:
@dosumis thanks, I hadn't realized that there can't be more than one rdfs:comment on a term, makes sense. Will fix now

shawntanzk:
> Re - huge log, I suspect someone has set Robot to verbose. @shawntanzk - have you been working on this?

Only change I've made here is to PR import, see https://github.com/obophenotype/cell-ontology/pull/1299/files 

Might be the update in ODK here https://github.com/obophenotype/cell-ontology/pull/1289

Will try to take a look at the fail log too

shawntanzk:
made a string fix to make it easier to see the diff


> I hadn't realized that there can't be more than one rdfs:comment on a term,

I too thought multiple comments was fine - I think I had a few terms with multiple comments before, is it a new check implemented and is that correct @dosumis?

paolaroncaglia:
@shawntanzk 
I keep having trouble with my edits and I'd rather re-do them when I'm fresher, so I'd like to close this PR. However I see that the checks have passed now, do they refer to your string fix PR above?

dosumis:
> I too thought multiple comments was fine - I think I had a few terms with multiple comments before, is it a new check implemented and is that correct @dosumis?

Not legal OBO

shawntanzk:
> so I'd like to close this PR.

Don't think you need to, the only issue was multiple comments which you have fixed. 
the string diffs are just something that happens when you save using protege. 
You can follow this guide to remove them: https://obophenotype.github.io/cell-ontology/Fixing_xsdstring_diffs.html
makes for a cleaner diff that is easier to review :)

PR #1318, opened 2021-11-23
add new terms for alveolar capillary cells and pulmonary capillary endothelial cell

JoshuaFortriede:


paolaroncaglia (changes requested):
@JoshuaFortriede congratulations on creating your first CL terms!
Here are my comments:

The longer you keep a branch to work on it, the more likely you are to run into conflicts. Not a blocker, just mildly annoying and more time-consuming.

You added this assertion:
SubClassOf(obo:CL_2000016 ['lung microvascular endothelial cell’] obo:CL_2000008 ['microvascular endothelial cell’])
which isn’t wrong, but actually I think what’s not fully correct in the existing ontology is that 'lung microvascular endothelial cell’ has the equivalence axiom
'lung endothelial cell' and ('part of' some 'microvascular endothelium')
rather than 
'microvascular endothelial cell’ and ('part of' some 'microvascular endothelium')
otherwise the link to blood vessel (contained in 'microvascular endothelial cell’) would be missing - also because ‘microvascular endothelium’ isn’t currently connected to blood vessel. Bottom line, feel free to leave your edit as is, and we can dig deeper into the above at a separate time.

You defined CL_4028001 (pulmonary capillary endothelial cell) as “Any capillary endothelial cell that is part of a pulmonary capillary.”. That’s not incorrect but slightly redundant: a capillary endothelial cell is already part of a capillary, so the differentia cane restricted to the lung, i.e. “Any capillary endothelial cell that is part of a lung.”. Same for the equivalence axiom, it should point to UBERON ‘lung’ rather than ‘pulmonary capillary’.

The asserted SubClassOf(obo:CL_4028001 obo:CL_0002144) is redundant, as per our discussion yesterday, and should be removed.

The asserted SubClassOf(obo:CL_4028001 obo:CL_2000016) isn’t incorrect, but wouldn’t be needed if ‘microvascular endothelium’ was connected to blood vessel in Uberon. So feel free to leave it, and we’ll dig deeper separately.

Re CL_4028002 (alveolar capillary type 1 endothelial cell): the synonyms you added should not be exact synonyms, but rather broad synonyms, as per our discussion yesterday. Also you made the term a simple subclass of a Uberon term, it should be SubClass Of ‘part of’ some ‘alveolus of lung’ instead.

You made CL_4028002 (alveolar capillary type 1 endothelial cell) disjoint from CL_4028003 (alveolar capillary type 2 endothelial cell). That’s not incorrect, but it’s a strong statement: if you can envision e.g. “transitions zones” where it’d be difficult to distinguish the two, could the axiom create issues? There should be some documentation on disjoints in the training material from the CL training course, if not we can look for some.

You defined CL_4028003 (alveolar capillary type 2 endothelial cell) as "An alveolar capillary endothelial cell that is located proximally to alveolar capillary type 1 endothelial cells and in close apposition to Alveolar Type 1 epithelial cells.”. Alveolar Type 1 epithelial cell should be in lower caps please. Also, in CL it’s a synonym for 'type I pneumocyte’, so for easier internal reference I’d suggest "An alveolar capillary endothelial cell that is located proximally to alveolar capillary type 1 endothelial cells and in close apposition to alveolar type 1 epithelial cells (also known as type I pneumocytes).”
Is aerocyte really an exact synonym? Is it ever used in the literature to refer to something else? If there’s any chance of ambiguity in usage, I’d make it broad. 
"capillary 2 cell" (not "capillary 2 cells”) should be a broad synonym, not exact.
Please make the term a subclass of part of Uberon, see above.

Long list, but consider it part of training :-) thanks again for contributing to CL!

paolaroncaglia:
I've opened https://github.com/obophenotype/uberon/issues/2186 to address 'microvascular endothelium'.

JoshuaFortriede:
lung microvascular endothelial cell - Equivalent To: Left alone for now. The equivalent class of this term is different from other subclasses of microvascular endothelial cell. Should these be cleaned up?

CL_4028001 (pulmonary capillary endothelial cell) - Fixed definition and equivalent to
asserted SubClassOf(obo:CL_4028001 obo:CL_0002144) - Removed
asserted SubClassOf(obo:CL_4028001 obo:CL_2000016) - Left for now.

Exact synonyms have been changed to Broad

Disjoint with designations: Removed until more information is known. 

Aerocyte synonym: I can only find literature on this related to this specific cell type. I added the PMID as a database cross reference on the synonym

shawntanzk:
Hi @JoshuaFortriede unfortunately I have very little knowledge of the lung and hence probably am not the best person to review this. The checks seem to pass so I don't think there is any issues. I'm not sure about the PMID xref sitting outside annotating something like the definition, but im sure @paolaroncaglia would comment on that if it is an issue. Other than that, it doesn't seem like it would cause any clashes

paolaroncaglia:
Thanks @shawntanzk .
@JoshuaFortriede , Shawn is right, the PMIDs used as definition dbxrefs should be added as I showed you on Monday. Instructions can be found at the bottom of page 32 in the Cell Ontology training doc, https://docs.google.com/document/d/1owDKQjZG4MDrhQDAU9zjFjcxk9M0WAtXjsx38hR3T4w/edit#heading=h.xcohup59d78. 
Everything else looks good, and checks have passed, so once definition PMIDs are in the right place for your new terms and checks have run again, I'll approve and you can merge. Thanks!

paolaroncaglia (approved):
@JoshuaFortriede 
Looks good to me. Please go ahead and merge; or tag me if you don't have permissions to. Thanks.

paolaroncaglia:
@JoshuaFortriede also, after merging, please delete the branch as you'll see at the bottom of the ticket.

PR #1363, opened 2021-12-09
Add `sed -E` flag for consistency

### LINKED ISSUE #1362: [Typo/Bug] Add '-E' flag in document 'Fixing ^^xsd:string Diffs'
**Description of typo, bug or error**
The command in the document `docs/Fixing_xsdstring_diffs.md`:

> If "make" is not installed, on MAC: ``` sed -i '' "s/Annotation[(]\(oboInOwl[:]hasDbXref [\"][^\"]*[\"]\)[)]/Annotation(\1^^xsd:string)/" cl-edit.owl ```

lacks the sed flag `-E` found in the `cl.Makefile` command:

>src/ontology/cl.Makefile:       `sed -i.bak -E "s/Annotation[(](oboInOwl[:]hasDbXref [\"][^\"]*[\"])[)]/Annotation(\1^^xsd:string)/" $<`

For consistency, I would change the command in the document and add the `-E` flag as follows:

``` sh
sed -i '' -E "s/Annotation[(](oboInOwl[:]hasDbXref [\"][^\"]*[\"])[)]/Annotation(\1^^xsd:string)/" cl-edit.owl
```

rays22:
This commit intends to update the documentation on fixing `^^xsd:string` issues. It adds an `-E` flag so that the command line method of normalisation is consistent with the one in the `cl.Makefile`.

If applied, this commit will fix #1362.

### DIFF
--- a/docs/Fixing_xsdstring_diffs.md
+++ b/docs/Fixing_xsdstring_diffs.md
@@ -13,14 +13,10 @@ When you make edits, sometimes there will be large amounts of unintended differe
 
   If you do not have **docker** installed: ```make normalise_xsd_string```
 
-<<<<<<< HEAD
   If **make** is not installed, on MAC:
 ``` sh
-sed -i '' -E "s/Annotation[(](oboInOwl[:]hasDbXref [\"][^\"]*[\"])[)]/Annotation(\1^^xsd:string)/" cl-edit.owl
+sed -i '' -E "s/Annotation[(](oboInOwl[:]hasDbXref [\"][^\"]*[\"])[)]/Annotation(\1^^xsd:string)/g" cl-edit.owl
 ```
-=======
-  If "make" is not installed, on MAC: ``` sed -i '' "s/Annotation[(]\(oboInOwl[:]hasDbXref [\"][^\"]*[\"]\)[)]/Annotation(\1^^xsd:string)/g" cl-edit.owl ```
->>>>>>> master
 
 
 This should resolve your `^^xsd:string` issue, after which, you can handle your pull request as per usual.
--- a/docs/Fixing_xsdstring_diffs.md
+++ b/docs/Fixing_xsdstring_diffs.md
@@ -1,5 +1,5 @@
-# Fixing ^^xsd:string Diffs
-When you make edits, sometimes there will be large amounts of unintended differences that show up that involves the removal of ^^xsd:string. If so, you can resolve them by following normalising your cl-edit.owl file.
+# Fixing `^^xsd:string` Diffs
+When you make edits, sometimes there will be large amounts of unintended differences that show up that involves the removal of `^^xsd:string`. If so, you can resolve them by following normalising your **cl-edit.owl** file.
 
 ## SOP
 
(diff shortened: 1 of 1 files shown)


dosumis:
Did you make sure it still works from the container (with ./run.sh)?

rays22:
> This does not work for me:
> 
> ```
> $ sed -i '' -E "s/Annotation[(]\(oboInOwl[:]hasDbXref [\"][^\"]*[\"]\)[)]/Annotation(\1^^xsd:string)/" hp-edit.owl
> sed: 1: "s/Annotation[(]\(oboInO ...": \1 not defined in the RE
> $ sed -i '' "s/Annotation[(]\(oboInOwl[:]hasDbXref [\"][^\"]*[\"]\)[)]/Annotation(\1^^xsd:string)/" hp-edit.owl
> $ 
> ```

Sorry @matentzn . I missed to fix the incorrectly escaped brackets `\(` and `\)` that cause the command copied from the original document to fail.
The `-E` does not seem to make a difference on macOS, so we can omit it if you want to make it Posix compliant. This is an excerpt from `man sed` on macOS:

>The -E, -I, -a and -i options, the special meaning of -f -, the prefixing “+” in >the second member of an address range, as well as the “I” flag to the address >regular expression and substitution command are non-standard FreeBSD >extensions and may not be available on other operating systems.
>macOS 12.0                                    June 10, 2020                                   macOS >12.0

matentzn:
What do you suggest - maybe have one instruction for MacOS, and another for other Lunix systems?

rays22:
> What do you suggest - maybe have one instruction for MacOS, and another for other Lunix systems?

I have checked on Linux and the GNU `sed` version also has the `-E` option. I have also read the GNU `sed` documentation and it says that the `-E` flag for extended regular expression syntax is already part of the Posix standard, so it should not cause compatibility issues with `sed` on Linux, Mac or other Posix compliant OS-es.

matentzn:
Do whatever you think is right, and keep @paolaroncaglia in the loop which command she should be using!

rays22:
1. The current stand-alone command suggested in the `Fixing ^^xsd:string Diffs` document will fail if you add the `-E` flag because some issues in the regular expression syntax versions.
    - **Explanation** The `-E` flag makes `sed` use the newer extended regular expression syntax rather than the older basic regular expressions.

It is hard to notice by eye, but the syntax differences are at `\(` and `\)`:

``` sh
..........................\(oboInOwl[:]hasDbXref [\"][^\"]*[\"]\)............................................
...........................(oboInOwl[:]hasDbXref [\"][^\"]*[\"]).............................................
```
2. The current command works without the `-E` flag:
``` sh
sed -i '' "s/Annotation[(]\(oboInOwl[:]hasDbXref [\"][^\"]*[\"]\)[)]/Annotation(\1^^xsd:string)/" cl-edit.owl
```

, but **it fixes only the first missing `^^xsd:string`** instance on every line, so you need to run it multiple times to fix all the missing `^^xsd:string` instances.
3. The cl.Makefile uses the modern regular expression syntax and [it has been updated to do a global fix in one go.](https://github.com/obophenotype/cell-ontology/pull/1366)
4. I would recommend the following stand-alone command that is basically identical to the one in the cl.Makefile and should also do the normalisation in a single pass:

``` sh
sed -i '' -E "s/Annotation[(](oboInOwl[:]hasDbXref [\"][^\"]*[\"])[)]/Annotation(\1^^xsd:string)/g" cl-edit.owl
```

matentzn:
I am all with you! Add that to the docs, push it and let everyone on slack know that they should be using that! :) Thanks

paolaroncaglia:
@rays22 @matentzn FYI, I just used the command in the latest PR and it works brilliantly. Thank you very much.

matentzn (approved):

Issue #1368, opened 2021-12-10
Typo in def of CL:0000359 vascular associated smooth muscle cell


paolaroncaglia:
**CL term**
CL:0000359 vascular associated smooth muscle cell

**Description of typo, bug or error**
"A smooth muscle cell _assocatiated_ with the vasculature."
=> please correct to associated.

bvarner-ebi:
@paolaroncaglia , @dosumis, after opening cl-edit.owl and running the reasoner (before making any changes), I see the following disjoint appear on all the classes:

![Screenshot 2021-12-16 at 05 13 38](https://user-images.githubusercontent.com/95643797/146352817-029ae897-ef8b-4437-b5a5-a1df9447f977.png)

owl:Nothing is also red in the class hierarchy.

Is this expected?

dosumis:
@shawntanzk  - is this the inconsistency coming in from GO which is only visible with ELK 0.5 ?  Should we be recommending that everyone switch back to [elk 0.43](https://github.com/liveontologies/elk-reasoner/releases) until this is fixed?

shawntanzk:
Yeah it is
Leaving this link so that this ticket gets linked there to it https://github.com/geneontology/go-ontology/issues/16524

paolaroncaglia:
As per discussion on cl-editors Slack: 
Please could all editors make sure they have ELK 0.4.3 installed for Protege (https://github.com/liveontologies/elk-reasoner/releases), and use this for editing CL until further notice. Thanks!

dosumis:
Sorry all - I got a slap on the wrist from @matentzn  for suggesting this.  He prefers fixes upstream that stop this and related issues from bleeding through.

bvarner-ebi:
@dosumis, would it be prudent to pause future pull requests until this is resolved?

matentzn:
Its resolved #1377 

dont stop yourself @bvarner-ebi

shawntanzk:
#1377 has been merged to master - once you update it, it should not show any unsats anymore :)

PR #1381, opened 2021-12-16
Fix typo #1368

bvarner-ebi:
@dosumis, @paolaroncaglia, would you mind reviewing this correction related to issue #1368?


bvarner-ebi:
Seeing many changes involving xsd:string like the following in the diff...

Screenshot 2021-12-16 at 08 23 12

Please confirm that these are expected.

shawntanzk:
@bvarner-ebi follow this: https://obophenotype.github.io/cell-ontology/Fixing_xsdstring_diffs/ to remove string diffs :)

dosumis:
@bvarner-ebi some background: this is an annoying bug that shows up sporadically.  It's been fixed in the OWL-API lib that causes it, but we're waiting for this fix to make it's way through to Protege. Until then, we need to use the fix @shawntanzk pointed to above.

bvarner-ebi:
Thank you, @shawntanzk and @dosumis.
I'm going to delete this branch/PR and start again after running the fix.

shawntanzk:
@bvarner-ebi you dont actually have to - you can just run the line and push that commit in - should work :) but up to your preference

PR #1405, opened 2022-01-14
Adding import documentation

bvarner-ebi:
Adding brief documentation on what to do if one needs to add a class/term to CL that has not yet been imported from another ontology

### DIFF
--- a/docs/Adding_classes_from_another_ontology.md
+++ b/docs/Adding_classes_from_another_ontology.md
@@ -1,4 +1,4 @@
-# How to add (import) a class to the Cell Ontology (CL) from another ontology
+# How to add (import) classes to the Cell Ontology (CL) from another ontology
 
 ## 1. Follow steps 1 - 5 under the heading [Protege-based declaration.](https://obophenotype.github.io/cell-ontology/odk-workflows/UpdateImports/#protege-based-declaration)
 
--- a/docs/adding_class_from_another_ontology.md
+++ b/docs/adding_class_from_another_ontology.md
@@ -1,0 +1,18 @@
+# How to add (import) a class to the Cell Ontology (CL) from another ontology
+
+## 1. Follow steps 1 - 5 under the heading [Protege-based declaration.](https://obophenotype.github.io/cell-ontology/odk-workflows/UpdateImports/#protege-based-declaration)
+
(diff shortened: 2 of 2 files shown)


bvarner-ebi:
Thank you very much for the feedback so far, @shawntanzk!

dosumis:
Looks good.  @shawntanzk - does this method no longer work?  https://obophenotype.github.io/cell-ontology/odk-workflows/UpdateImports/#using-term-files

Addition in Protege has a couple of minor drawbacks:
 - it adds to declarations in the editor's file (occasionally a cause or problems, e.g. declarations may need to be deleted if terms are no longer used.
 - prior to import it causes failure of the SPARQL label and def checks (or at least it used to) - I think this is connected to the presence of declarations.

I don't think this is a blocker though.  Maybe worth discussion in a future tech call.

dosumis (approved):

shawntanzk:
> does this method no longer work? https://obophenotype.github.io/cell-ontology/odk-workflows/UpdateImports/#using-term-files

unfortunately not for merged/base file import, from my understanding, it does not use the separate import files anymore. 

I will add the concerns for the tech call and get back to you on them.

@gouttegd I can't remember exactly what it is, but do we need to index the docs somewhere? also do we need to run a make the docs or something? I vaguely remember something about that but its been awhile since I've edited the docs. Thanks!

gouttegd:
> I can't remember exactly what it is, but do we need to index the docs somewhere?

Not that I am aware of. Please note, however, that the entire `obophenotype.github.io` website hosting the documentation has been set up long before I joined the party, and I was never involved with it. So I may not be aware of everything. :)

> also do we need to run a make the docs or something?

There’s a GitHub Action in place to do that automatically. It [doesn’t work in the ODK repository](https://github.com/INCATools/ontology-development-kit/issues/506) (I am testing a workaround for that right now), but in CL it should work as expected.

shawntanzk:
> There’s a GitHub Action in place to do that automatically.

omg the quality of life improvement on this is amazing! think you were the one who implemented this right? if so thanks so much!

gouttegd:
> think you were the one who implemented this right?

Nope. :) I implemented the actions to automatically build, test, and release the ODK.

I believe the actions to automatically build the docs were added by Nico.

PR #1407, opened 2022-01-18
update multipolar neuron def

### LINKED ISSUE #3: Terms with definitions that are too vertebrate specific for their names: Terms with definitions that are too vertebrate specific for their names:
```
For invertebrate neurons, we need basic morphological classification of neurons 
to be agnostic about axon vs dendrite.  But in CL we currently have:

CL:0000106 unipolar neuron: "Neuron with an axon and no dendrites." 
CL:0000104 multipolar neuron: "Neuron with an axon and two or more dendrites."
For both: remove mention of axons and dendrited.  Use neurite.

There are bipolar sensory neurons in insects, but bipolar neuron in CL is a 
SubClassOf interneuron:

CL:0000103 bipolar neuron: "A type of interneuron that has two neurites, 
usually an axon and a dendrite, extending from opposite poles of an ovoid cell 
body."

Given the definition, we may want to rename this term and create a new 
superclass that is purely morphologicall defined.


```

Original issue reported on code.google.com by `dosu...@gmail.com` on 24 Feb 2012 at 2:04

shawntanzk:
Fixes #3
That's pretty amazing, issue 3!

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -4052,7 +4052,7 @@ EquivalentClasses(obo:CL_0000103 ObjectIntersectionOf(obo:CL_0000099 ObjectSomeV
 
 # Class: obo:CL_0000104 (multipolar neuron)
 
-AnnotationAssertion(Annotation(oboInOwl:hasDbXref "FMA:67287"^^xsd:string) Annotation(oboInOwl:hasDbXref "GOC:tfm"^^xsd:string) obo:IAO_0000115 obo:CL_0000104 "Neuron with an axon and two or more dendrites."^^xsd:string)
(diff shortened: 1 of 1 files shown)


paolaroncaglia (commented):
Hi @shawntanzk ,
Looks ok to me, but I'd be happier if @dosumis and/or @gouttegd could please confirm that the new def would apply well across species. Thank you!

gouttegd:
The text definition looks good to me.

But `multipolar neuron` is logically defined as `neuron and ('has characteristic' some 'multipolar neuron morphology')`, and `multipolar neuron morphology` (PATO:0070026) is defined as “a cell morphology that inheres in neurons which possess single axon and many dendrites and dendritic branches”.

Should the definition of `multipolar neuron morphology` in PATO be updated as well so as to avoid explicitly referring to an axon? Otherwise there would be a discrepancy between the text definition of `multipolar neuron` and its logical definition.

shawntanzk:
made the PR for PATO here: https://github.com/pato-ontology/pato/pull/467

gouttegd (approved):

PR #1422, opened 2022-01-20
Update cl-idranges.owl

dosumis:
Adding Bradley + cleaning up some cruft in file.


shawntanzk (approved):

gouttegd (approved):

shawntanzk:
@bvarner-ebi - in case you need to do this for another ontology, see diff here :)

PR #1425, opened 2022-01-20
Closes #1409 - NTR choroidal melancoyte

### LINKED ISSUE #1409: [NTR]Choroidal melanocyte
**Preferred term label:** Choroidal melanocyte

**Synonyms**: -

**Definition**
Choroidal melanocytes are melanin-producing cells in the uvea of the human eye. Their cellular functions and extracellular interactions are distinct to skin melanocytes. They are involved in photoprotection, regulation of oxidative damage and immune responses. Choroidal melanocytes might have both neuroectodermal as well as mesectodermal origins. [PMID: 31819095, PMID: 34082079]

**Parent cell type term:**
Melanocyte (CL:0000148)

**What is the anatomical structure that the cell is a part of?**
Optic choroid (UBERON:0001776)

0000-0002-0819-0473

**Any additional notes or concerns** We are mapping cell types in Cellosaurus to CL. We found most of the terms, only few are missing. We  would be grateful if you could create them. This is one of them. Many thanks!

bvarner-ebi:
Added new class 'choroidal melancoyte'


src/ontology/cl-edit.owl
```diff
@@ -29524,6 +29525,16 @@ AnnotationAssertion(rdfs:label obo:CL_4028006 "alveolar type 2 fibroblast cell")
 SubClassOf(obo:CL_4028006 obo:CL_0002241)
 SubClassOf(obo:CL_4028006 ObjectSomeValuesFrom(obo:BFO_0000050 obo:UBERON_0002299))
 
+# Class: obo:CL_4030000 (choroidal melanocyte)
+
+AnnotationAssertion(<http://geneontology.org/formats/oboInOwl#created_by> obo:CL_4030000 "http://orcid.org/0000-0002-1773-2692")
```
shawntanzk, line 29530: would you mind switching this to dcterms:contributor please -> think there is an effort to switch (which reminds me I need to change mine too lol)

src/ontology/cl-edit.owl
```diff
@@ -29524,6 +29525,16 @@ AnnotationAssertion(rdfs:label obo:CL_4028006 "alveolar type 2 fibroblast cell")
...
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref "PMID:31819095"^^xsd:string) Annotation(oboInOwl:hasDbXref "PMID:34082079"^^xsd:string) obo:IAO_0000115 obo:CL_4030000 "Choroidal melanocytes are melanin-producing cells in the uvea of the human eye. Their cellular functions and extracellular interactions are distinct to skin melanocytes. They are involved in photoprotection, regulation of oxidative damage and immune responses. Choroidal melanocytes might have both neuroectodermal as well as mesectodermal origins.")
```
shawntanzk, line 29531: I think convention is to have "A [parent class] that..." and not to use the name in the first sentence (definiendum). So: "A melanocyte in the uvea of the human eye" (or should it be choroid?) - also is this human specific? 

Chris Mungall wrote a really great blog that might help (it is quite a bit of a read though) - https://douroucouli.wordpress.com/2019/07/08/ontotip-write-simple-concise-clear-operational-textual-definitions/

shawntanzk: I think "Their cellular functions and extracellular interactions are distinct to skin melanocytes." and "Choroidal melanocytes might have both neuroectodermal as well as mesectodermal origins." should be in rdfs:comments instead

src/ontology/cl-edit.owl
```diff
@@ -29524,6 +29525,16 @@ AnnotationAssertion(rdfs:label obo:CL_4028006 "alveolar type 2 fibroblast cell")
...
+AnnotationAssertion(<http://purl.org/dc/elements/1.1/contributor> obo:CL_4030000 "http://orcid.org/0000-0002-0819-0473")
+AnnotationAssertion(<http://purl.org/dc/elements/1.1/date> obo:CL_4030000 "2022-01-20T12:59:06Z"^^xsd:dateTime)
```
shawntanzk, line 29533: not sure we use date stamps in CL - at least I don't. Don't see it around much and I guess trying to follow convention is ideal.

bvarner-ebi: I see some terms with a "creation date" annotation. If I change the annotation to "creation date", would that be better? Or do you propose eliminating creation date going forward?

src/ontology/cl-edit.owl
bvarner-ebi: Thanks for the feedback, @shawntanzk. I used the definition provided by the requestor but can break it down per your suggestions. The references in the ticket both reference humans, but can do additional searching for non-human references.

src/ontology/cl-edit.owl
dosumis: Bit shocked we're still arguing over this.  Looks like Bradley has the [setup specified in the Mondo tutorial](https://oboacademy.github.io/obook/lesson/contributing_to_obo_ontologies/#setting-preferences-for-new-entities-metadata).  I'm still using creation_date - as this is what we taught in the last CL workshop.  I don't want to derail work by worrying about it, but we should aim for a consensus on the standard, we can then retrofit for consistency.  Perhaps ping Nicole and Nico for their opinion on this?

dosumis: But we shouldn't worry about it for this PR

src/ontology/cl-edit.owl
bvarner-ebi: Is there a standard for annotating who edits/creates the term vs who requests the term?
Does requestor = contributor and editor (myself) = dcterms:contributor OR should everyone be annotated with dcterms:contributor?

shawntanzk: I think we are trying to move everyone to dcterms:contributor (I know that's Nico's preference at least) last we talked about this at least. Sorry for this being super confusing and all, we have recently tried to unify all these stuff but getting consensus is SUPER difficult

bvarner-ebi: Thank you, @shawntanzk! For this PR and going forward I will set both the requestor and editor to dcterms:contributor. 
Since I have not found a workflow for it, I'll plan to draft the steps for creating a new term in CL and then share with the group in an effort to get consensus.

paolaroncaglia:
I believe that the SOP is to represent requestors via their ORCID as a definition dbxref (annotation of the definition), so, separately from whoever created the term. Sometimes they'll be the same person.
Re. creation date, I find it useful when archeological digging of terms is called for - the ID range and/or created_by etc. can tell you who created the term, but sometimes it's really helpful to see at a glance which of two terms is older, e.g. when considering merges.

src/ontology/cl-edit.owl
bvarner-ebi: I found references to this cell type in mice, cats and dogs, so I removed the reference to humans.

bvarner-ebi: @shawntanzk, valid point on uvea vs choroid location. 
PMID: 31819095 states "Choroidal melanocytes (HCMs) are melanin-producing cells in the vascular uvea of the human eye (iris, ciliary body and choroid)". 
The requestor assigned optic choroid as the anatomical location and the name itself does suggest it is part of the choroid. In these cases does it warrant a follow up with the requestor or should we decide based on our own research?

shawntanzk: from personal experience, I try to do what I can from research and then in the end check with requester or an expert if it is alright/what needs changing - tends to get things done a lot faster that way (so I basically write everything up how I think is right then ask) But perhaps more importantly, the textual definition should match up with your logical definition as much as possible.

Personally, I would use vascular uvea to be safe but also change the logical axiom as such (if you put choroid, then if cannot be in iris and ciliary body, but if you put vascular uvea, it is still in choroid) - really annoying given the name of the cell though lol. But probably also check with someone more experienced than me, maybe @paolaroncaglia could comment on how to handle this?

bvarner-ebi: Thank you for the guidance, @shawntanzk. The requester agreed to align the definition with the anatomical structure in EA, so both are set to 'uvea'.

shawntanzk: I think it would be useful to have a has_exact_synonym as 'uvea melanocyte' in this case - given the equivalent class axiom - sorry for being a pain >.<

dosumis: maybe "uveal"

shawntanzk: lol, yes, that >.<

bvarner-ebi: Not a pain, @shawntanzk. I appreciate the feedback and added the synonym.

shawntanzk (approved):

PR #1442, opened 2022-02-03
Closes #1438 NTR-oligodendrocyte_type_2_astrocyte_progenitor_cell

### LINKED ISSUE #1438: [NTR] Oligodendrocyte/type-2 astrocyte progenitor cell
**Preferred term label:** Oligodendrocyte/type-2 astrocyte progenitor cell

**Synonyms**
 O-2A/OPCs
oligodendrocyte type 2 astrocyte (O-2A) progenitor cells
oligodendrocyte-type-2 astrocyte progenitor cells

**Definition**: Oligodendrocyte/type-2 astrocyte progenitor cells can generate oligodendrocytes and type-2 astrocytes [PMID: 24133281, PMID: 2182078, PMID: 11756508, PMID: 9826671]

**Parent cell type term:** neuron associated cell (sensu Vertebrata) (CL:0000123)

**What is the anatomical structure that the cell is a part of?** brain (UBERON:0000955)

**Your nano-attribution (ORCID)**: 0000-0002-0819-0473

1) These cells can be purified from optic nerves and other regions of the embryonic, postnatal, and adult rat CNS and can be differentiated _in vitro_ to oligodendrocytes and type-2 astrocytes. I do not know if this must be indicated somewhere
2) the synoynm "oligodendrocyte precursor cells" is sometimes used in papers to refer to "oligodendrocyte/type-2 astrocyte progenitor cells". This "synonym" should not be included here because "oligodendrocyte precursor cell" CL:0002453 corresponds to a different cell type.

bvarner-ebi:
Closes #1438 NTR-oligodendrocyte_type_2_astrocyte_progenitor_cell


bvarner-ebi:
@paolaroncaglia, @shawntanzk, I will appreciate your feedback.

I noticed a similar term (CL:0002453 oligodendrocyte precursor cell) has EA:
'neuron associated cell (sensu Vertebrata)'
 and ('part of' some 'central nervous system')

I was considering to add that same axiom as a SubClass Of for this new term (CL:4030003). Is the EA above appropriate to be an EA for CL:0002453 since it is necessary, but not sufficient, to describe a oligodendrocyte precursor cell?

paolaroncaglia (commented):
@bvarner-ebi 
Briefly, only 1 rdfs:comment is allowed per class. Also, in the first one, please fix typo "synoynm" and replace cells with cell in "The synoynm "oligodendrocyte precursor cells...". Thanks!

paolaroncaglia:
@bvarner-ebi 

> I noticed a similar term (CL:0002453 oligodendrocyte precursor cell) has EA: 'neuron associated cell (sensu Vertebrata)' and ('part of' some 'central nervous system')
> 
> I was considering to add that same axiom as a SubClass Of for this new term (CL:4030003). Is the EA above appropriate to be an EA for CL:0002453 since it is necessary, but not sufficient, to describe an oligodendrocyte precursor cell?

Good catch! No, the EA should be removed; in fact, it's also present as subclass axioms. It may have been a genuine mistake. I'd recommend to remove it please.

paolaroncaglia:
@bvarner-ebi checks are failing very likely due to the 2 rdfs:comments.

paolaroncaglia:
@bvarner-ebi also, I'd consider relaxing the definition of CL:0002453 oligodendrocyte precursor cell slightly, from
"The cell type from which oligodendrocytes develop..." to "A cell type..." or similar, as your new term will now represent another, separate class from which oligod. may develop.

shawntanzk:
Wow, 100% the EA should be removed, amazing it didnt cause more trouble with that EA. Good catch :) 

Does this cell develop from glioblast too? or is it a type of glioblast? Currently glioblast (sensu Vertebrata) is reasoned to be a progenitor cell - which I assume the cell you are adding should be too. This could be done through adding an axiom that reasons that out, or asserted. (of course be conservative if research doesn't give a straightforward answer to this)

bvarner-ebi:
> Wow, 100% the EA should be removed, amazing it didnt cause more trouble with that EA. Good catch :)
> 
> Does this cell develop from glioblast too? or is it a type of glioblast? Currently glioblast (sensu Vertebrata) is reasoned to be a progenitor cell - which I assume the cell you are adding should be too. This could be done through adding an axiom that reasons that out, or asserted. (of course be conservative if research doesn't give a straightforward answer to this)

Thanks for the thoughtful suggestion, @shawntanzk. I saw it referred to as a 'glial progenitor cell' (PMID:24133281), but not specifically a glioblast. I'll do some more investigation.

bvarner-ebi:
With further investigation, I am not yet convinced that this newly requested cell type is unique from 'oligodendrocyte precursor cell'. It seems that the cell type is a biopotential precursor for both oligodendrocytes and type-2 astrocytes. I haven't found evidence referring to a separate precursor restricted oligodendrocyte-only proliferation .

For example, from PMID:10704434:

"Oligodendrocytes arise from a well-defined precursor cell that also can give rise to a specific astrocyte population (the type-2 astrocyte) in vitro ([Raff et al. 1983](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2174536/#Raffetal1983)), although, thus far, it appears that this astrocyte pathway is not used during development ([Skoff 1990](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2174536/#Skoff1990)). **Due to this disparity, this single precursor cell is alternatively called an oligodendrocyte type-2 astrocyte (O-2A) progenitor or an oligodendrocyte precursor cell (OPC)**."

From https://doi.org/10.1038/s41598-017-14207-0:

"In addition, we found that Venus-positive OPCs were able to differentiate to type II astrocytes. In vivo, OPCs and oligodendrocytes express Venus, and some of astrocytes were positive for Venus in the ventral cortex."

From https://doi.org/10.1242/dev.126409:

"Oligodendrocytes are the myelinating cells of the central nervous system (CNS), and the last major neural phenotype to form during development. They arise from oligodendrocyte progenitor cells (OPCs), which can generate either oligodendrocytes or astrocytes depending on context ([ffrench-Constant and Raff, 1986](); [Raff et al., 1983a]())."

I will follow-up with the requester and ask if she approves enhancing the current 'oligodendrocyte precursor cell' term instead of creating a separate term for 'oligodendrocyte type-2 astrocyte progenitor cell'.

shawntanzk:
Perhaps a narrow synonym would do the job then? (adding a comment on OPC might help too)

bvarner-ebi:
> Perhaps a narrow synonym would do the job then? (adding a comment on OPC might help too)

Yes, that sounds like a reasonable solution, @shawntanzk. I was going to give the requester a couple of days to respond before proceeding. Also, thank you for posing the question on whether this comes from a glioblast... that was the impetus to do some more investigation!

bvarner-ebi:
Hi, @shawntanzk, I added 'exact' synonyms, since it seems that these added terms are referring to the same entity as the the class label. Do you agree with these new revisions? Latest requester comments available at #1438.

shawntanzk:
@bvarner-ebi - perhaps narrow synonym would be more appropriate? 

The way I'm thinking is that while they refer to the same cell type, technically not all OPCs are "oligodendrocyte-type 2 astrocyte (O-2A) progenitor cell" - they are only that type if they do end up becoming astrocytes? sort of a weird existentialist "existence precedes essence" kindda thing? (not that I'm suggesting that O-2A are conscious beings that have the curse of freedom and responsibility lol)  

I might also be overthinking this as the name still has oligodendrocyte in it, but from my quick reading, "O-2A cells give rise to type-2 astrocytes in the presence of serum", which sounds O-2A refers to the subset of OPC that become type-2 astrocytes.

btws in case you need it, def for has_narrrow_synonym is "An alias in which the alias is narrower than the primary class name. Example: pyrimidine-dimer repair by photolyase is a narrow synonym of photoreactive " 

At least that's my thoughts - would be good to hear from others too though :)

bvarner-ebi:
Thanks for this perspective, @shawntanzk!

>which sounds O-2A refers to the subset of OPC that become type-2 astrocytes.

^ Actually, after searching some more, I found this diagram, which illustrates O-2A as a precursor to OPC...

From https://veteriankey.com/of-the-central-and-peripheral-nervous-system/:

B9780702028991000146_f4

What has made it less clear for me is that all the other references I've come across so far use the terms to refer to these cell types as the same entity with language like "also referred to as" and "alternatively called".

E..g., (https://jbiol.biomedcentral.com/articles/10.1186/jbiol50): 
"... oligodendrocyte-type-2 astrocyte (O-2A) progenitor cells (also referred to as oligodendrocyte precursor cells ..." 

From PMID:10704434:
"Oligodendrocytes arise from a well-defined precursor cell that also can give rise to a specific astrocyte population (the type-2 astrocyte) in vitro ([Raff et al. 1983](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2174536/#Raffetal1983)), although, thus far, it appears that this astrocyte pathway is not used during development ([Skoff 1990](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2174536/#Skoff1990)). Due to this disparity, this single precursor cell is alternatively called an oligodendrocyte type-2 astrocyte (O-2A) progenitor or an oligodendrocyte precursor cell (OPC)."

**Given the above, which option seems the best?**
- follow the original request to create a separate term for oligodendrocyte type-2 astrocyte (O-2A) progenitor cell and add axiom that states that oligodendrocyte precursor cell develops from this
- keep the original term oligodendrocyte precursor cell and use narrow/broad synonyms
- something else

shawntanzk:
@bvarner-ebi I think our jobs as curators is to present the biologists understanding of it, in which case, I don't think the way you suggest is wrong. looking at that diagram though, I might be convinced that it should be a broad synonym rather than a narrow synonym. I think broad synonym might be safer in this case. I would also leave in the comment a little about how OPCs might develop from O-2A but it is unclear if they are distinct cell types or reference to the same cell type and put some citations on them. 
Chances are, no one really knows lol, it happens a lot in neuro unfortunately, it always seems like nothing is ever clear cut in neuro >.<

bvarner-ebi:
Thanks for all the insight, @shawntanzk! I committed the changes.

src/ontology/cl-edit.owl
```diff
@@ -19054,13 +19054,20 @@ SubClassOf(obo:CL_0002451 obo:CL_0000048)
...
 AnnotationAssertion(oboInOwl:created_by obo:CL_0002453 "tmeehan"^^xsd:string)
 AnnotationAssertion(oboInOwl:creation_date obo:CL_0002453 "2010-11-10T02:51:34Z"^^xsd:string)
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref "PMID:24133281"^^xsd:string) oboInOwl:hasBroadSynonym obo:CL_0002453 "O-2A/OPC"^^xsd:string)
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref "PMID:11756508"^^xsd:string) oboInOwl:hasBroadSynonym obo:CL_0002453 "O2A/OPC"^^xsd:string)
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref "PMID:2253328"^^xsd:string) oboInOwl:hasBroadSynonym obo:CL_0002453 "oligodendrocyte-type 2 astrocyte (O-2A) progenitor cell"^^xsd:string)
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref "PMID:11756508"^^xsd:string) oboInOwl:hasBroadSynonym obo:CL_0002453 "oligodendrocyte-type-2 astrocyte progenitor cell"^^xsd:string)
```
shawntanzk, line 19063: this looks like a duplicate from the above except without (O-2A) - in which case i don't think there's a need. search stuff would be able to pick it up without having this synonym

src/ontology/cl-edit.owl
```diff
@@ -19054,13 +19054,20 @@ SubClassOf(obo:CL_0002451 obo:CL_0000048)
...
 AnnotationAssertion(oboInOwl:hasOBONamespace obo:CL_0002453 "cell"^^xsd:string)
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref "PMID:10704434"^^xsd:string) Annotation(oboInOwl:hasDbXref "PMID:2182078"^^xsd:string) Annotation(oboInOwl:hasDbXref "https://veteriankey.com/of-the-central-and-peripheral-nervous-system/"^^xsd:string) rdfs:comment obo:CL_0002453 "This cell type can be purified from optic nerves and other regions of the embryonic, postnatal and adult rat CNS and can be differentiated in vitro to oligodendrocytes and type-2 astrocytes. Some references use the terms 'oligodendrocyte precursor cell' and 'oligodendrocyte type-2 astrocyte (O-2A) progenitor' to refer to the same entity (e.g., PMID:10704434), while others describe an oligodendrocyte type-2 astrocyte (O-2A) progenitor cell as the precurosr to oligodendrocyte- and type-2 astrocyte- progenitor cells (e.g, https://veteriankey.com/of-the-central-and-peripheral-nervous-system/)."^^xsd:string)
```
shawntanzk, line 19066: Figure is from https://www.us.elsevierhealth.com/essentials-of-domestic-animal-embryology-9780702028991.html - can cite it as ISBN:9780702028991 instead of website

src/ontology/cl-edit.owl
```diff
@@ -19054,13 +19054,20 @@ SubClassOf(obo:CL_0002451 obo:CL_0000048)
...
 AnnotationAssertion(rdfs:label obo:CL_0002453 "oligodendrocyte precursor cell"^^xsd:string)
-EquivalentClasses(obo:CL_0002453 ObjectIntersectionOf(obo:CL_0000123 ObjectSomeValuesFrom(obo:BFO_0000050 obo:UBERON_0001017)))
-SubClassOf(Annotation(oboInOwl:is_inferred "true") obo:CL_0002453 obo:CL_0000123)
+SubClassOf(obo:CL_0002453 ObjectIntersectionOf(obo:CL_0000123 ObjectSomeValuesFrom(obo:BFO_0000050 obo:UBERON_0001017)))
+SubClassOf(obo:CL_0002453 ObjectIntersectionOf(obo:CL_0000123 ObjectSomeValuesFrom(obo:RO_0002203 obo:CL_0000127)))
```
shawntanzk, line 19069: I don't think you should have two develops into - this would mean that OPCs develop into both astrocytes and oligodendrocytes. (all statements must be true) - the opposite develops from might be a better way to axiomatize this (eg type-2 astrocytes develop from OPC and oligodendrocytes develop from OPCs)

shawntanzk (changes requested):
see in line comments

src/ontology/cl-edit.owl
bvarner-ebi: Removed SubClass Of axioms. 
Oligodendrocyte already has recommended axiom.
Type-2 astrocyte does not exist in CL. Create this new term now vs submit separate NTR ticket?

shawntanzk: 100% new NTR ticket - if not discussions about that will stop this ticket from going through lol

shawntanzk (approved):
Looks good to me - might also want to get a 2nd opinion on the broad synonym but I think its pretty safe. PS you might want to put in a dc:contributor for your own orcid, think it might help when you need to collate what you have done for job review when the time comes :)

PR #1445, opened 2022-02-09
Automated diff machine

shawntanzk:
tested it on a test branch and it should work nicely :)

don't know if its the most elegant solution but it's the way i got it to work

### DIFF
--- a/.github/workflows/diff.yml
+++ b/.github/workflows/diff.yml
@@ -29,7 +29,7 @@ jobs:
           path: edit-diff.md
   classify_branch:
     runs-on: ubuntu-latest
-    container: obolibrary/odkfull:v1.2.32
+    container: obolibrary/odklite:v1.2.32
     steps:
       - uses: actions/checkout@v2
       - name: Classify ontology
@@ -42,7 +42,7 @@ jobs:
           retention-days: 1
   classify_main:
     runs-on: ubuntu-latest
-    container: obolibrary/odkfull:v1.2.32
+    container: obolibrary/odklite:v1.2.32
     steps:
       - uses: actions/checkout@v2
         with:
@@ -60,7 +60,7 @@ jobs:
       - classify_branch
       - classify_main
     runs-on: ubuntu-latest
-    container: obolibrary/odkfull:v1.2.32
+    container: obolibrary/odklite:v1.2.32
     steps:
       - uses: actions/checkout@v2
       - name: Download master classification
--- a/.github/workflows/diff.yml
+++ b/.github/workflows/diff.yml
@@ -1,25 +1,50 @@
 name: 'Create ROBOT diffs on Pull requests'
 
 on:
+  # Triggers the workflow on pull request events for the master branch
   pull_request:
     branches: [ master ]
 
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
-  ROBOT_Diff:
+  classify_pr:
     runs-on: ubuntu-latest
-    container: obolibrary/odkfull:v1.2.26
-      
+    container: obolibrary/odklite:v1.2.32
     steps:
-      
       - uses: actions/checkout@v2
-      
-      - name: Run ROBOT diff
-        run: |
-          cd src/ontology && make rdiff
-          
-      - name: comment with diff
-        env:
-          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
-        uses: NejcZdovc/comment-pr@v1.1.1
+      - name: Upload PR reasoned.owl
+        uses: actions/upload-artifact@v2
+      # Checks-out main branch under "main" directory
+      - uses: actions/checkout@v2
         with:
-          file: "../../src/ontology/diff.md"
+          ref: main
+          path: main
+      - name: Diff classification
+        run: export ROBOT_JAVA_ARGS=-Xmx6G; robot diff --left main/src/ontology/cl-edit.owl --right src/ontology/cl-edit.owl -f markdown -o classification-diff.md
+      - name: Upload diff
+        uses: actions/upload-artifact@v2
+        with:
+          name: classification-diff.md
+          path: classification-diff.md
+      
+
+  post_comment:
+    needs: classify_pr
+    runs-on: ubuntu-latest
+    container: docker://dvcorg/cml-py3:latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Download diff
+        uses: actions/download-artifact@v2
+        with:
+          name: classification-diff.md
+          path: classification-diff.md
+      - name: Prepare comment
+        run: "echo \"Here's a diff of how these changes impact the classified ontology: \n\" >comment.md; cat classification-diff.md/classification-diff.md >>comment.md"
+      - name: Post comment
+        env:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+        run: cml-send-comment comment.md
--- a/.github/workflows/diff.yml
+++ b/.github/workflows/diff.yml
@@ -15,8 +15,6 @@ jobs:
     container: obolibrary/odklite:v1.2.32
     steps:
       - uses: actions/checkout@v2
-      - name: Upload PR reasoned.owl
-        uses: actions/upload-artifact@v2
       # Checks-out main branch under "main" directory
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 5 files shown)


src/ontology/cl.Makefile
```diff
@@ -20,6 +20,11 @@
 #tmp/pr_logical.owl: mirror/pr.owl
 #	echo "Skipped pr logical" && cp $< $@
 
+.PHONY: rdiff
+rdiff: 
+	curl https://raw.githubusercontent.com/shawntanzk/cell-ontology/master/src/ontology/cl-edit.owl -o rdifftmp.owl
```
matentzn, line 25: best do not curl (unnecessary dependency on network - always avoid), use `checkout master -- $(SRC)` instead.

src/ontology/cl.Makefile
```diff
@@ -20,6 +20,11 @@
...
+	$(ROBOT) diff -labels True -left rdifftmp.owl --right cl-edit.owl -format markdown --output diff.md
```
matentzn, line 26: For now this is fine @ubyndr it would be important that however you realised the final diff system we integrate in ODK, that it can be run on _any_ file in the ODK pipeline.

matentzn (changes requested):

src/ontology/cl.Makefile
shawntanzk: thanks! this was the part that I was quite unhappy about how inelegant it was, i usually use git show for this part but didnt know the best way to do this otherwise :)

matentzn: git show may be fine, maybe even better, I have not used it - just do not curl!

src/ontology/cl.Makefile
dosumis: Can we make this dependent on file extension?

src/ontology/cl.Makefile
shawntanzk: the struggle I had with that was that I couldn't get git commands to work with actions - trying this on test fork and it seems to be failing there too. but 100% agree that curl is not ideal

src/ontology/cl.Makefile
matentzn: exactly.. I agree. It should be something like:

```
rdiff-%: tmp/main-% tmp/branch-%*
	ROBOT diff tmp/main-$* tmp/branch-$* ....
```

matentzn: and if it is, say cl.owl, it should depend on cl.owl so that it is actually created by make first!

src/ontology/cl.Makefile
matentzn: I think the cleanest thing is to orchestrate this in the action itself, have multiple steps with different checkouts, each generating the file in question, then running the diff in a third step.

balhoff: @shawntanzk here's what I implemented for GO a while back, maybe it can be of use: https://github.com/geneontology/go-ontology/blob/master/.github/workflows/inference_diff.yml

matentzn: Oh YES, that is right you did that.. Sorry man I forgot about it. I think this is the best design!

shawntanzk: sweet! thanks so much @balhoff :D

src/ontology/cl.Makefile
shawntanzk: does this mean we want to compare $(SRC).owl that is made from $(SRC)-edit.owl?

src/ontology/cl.Makefile
ubyndr: Do we need 4 jobs, classify_pr, classify_master, diff_classification and post_comment? We are checking out our branch 3 times is that necessary? Can I use single job for, downloading branch, master, doing diff analysis and posting a comment? Or is that a bad design? @matentzn

shawntanzk:
@matentzn & @dosumis - current way from @balhoff creates a comment every commit - is this ideal or should we use a method that updates the first comment each time?

matentzn:
Creating a comment at every committ seems correct to me;

dosumis:
I think I'd prefer to update a single comment otherwise thread may get very long.  It has the downside that comments might make a bit less sense as updates happen.  But I think that's OK.

shawntanzk:
> I'd prefer to update a single comment 
personally this is my preference too, for that reason

ubyndr:
I have change the robot comment that @balhoff used in his action 
`robot diff --left ../../reasoned-master.owl/reasoned.owl --left-catalog catalog-v001.xml --right ../../reasoned-pr.owl/reasoned.owl --right-catalog catalog-v001.xml -f markdown -o classification-diff.md`, 
to 
`robot diff --left master/src/ontology/cl-edit.owl --right src/ontology/cl-edit.owl -f markdown -o classification-diff.md`. 
What is the purpose of using left and right catalog? Is that necessary? @matentzn & @dosumis

matentzn:
Ah sorry i didn't read that right. Single comment for sure.. I thought the question was how often to update, forget what I said above.

shawntanzk:
> What is the purpose of using left and right catalog?

This was my question when I saw it too lol

shawntanzk:
seems --update=TRUE either 1) doesnt work, or 2) is super confused cause there are so many previous comments - @ubyndr to test this out on another ontology in a new PR

balhoff:
If you don't provide the catalog, the imports won't be loaded correctly from local files.

src/ontology/cl.Makefile
balhoff: The separate jobs made it easier (to me) to work on two different branches. It could be 3 total jobs, but the diff and the comment used different docker images, so they are different jobs. But it sounds like you found a better comment posting action. Also with multiple jobs, the two classifications can happen simultaneously.

shawntanzk:
> If you don't provide the catalog, the imports won't be loaded correctly from local files.

@ubyndr - yeah we should put the catalog file then :) thanks @balhoff

shawntanzk:
Strategy: 
Change the ROBOT diff part to makefile and use github actions to call command in makefile
use variable for what files that the actions will look into (do this in ONT.makefile)

Strech: for ODK make a config that can define the files to go through diff

Thoughts - can the current commenting thing have 3 different comment? -> check if this is possible

shawntanzk:
minimum check these 2: 
ONT-edit.editformat 
ONT.owl

.github/workflows/diff.yml
```diff
@@ -0,0 +1,118 @@
...
+      - uses: actions/checkout@v2
+      - name: Classify ontology
+        run: cd src/ontology; make IMP=FALSE PAT=FALSE MIR=FALSE cl.owl
+      - name: Upload PR cl.owl
+        uses: actions/upload-artifact@v2
+        with:
+          name: cl-pr.owl
+          path: src/ontology/cl.owl
+          retention-days: 1
+  classify_master:
```
matentzn, line 43: We avoid `master` nomenclature now for political correctness reasons, use `classify_main` and `classify_branch` instead.

.github/workflows/diff.yml
```diff
@@ -0,0 +1,118 @@
...
+        run: export ROBOT_JAVA_ARGS=-Xmx6G; cd src/ontology; robot diff --left cl-master.owl/cl.owl --left-catalog catalog-v001.xml --right cl-pr.owl/cl.owl --right-catalog catalog-v001.xml -f markdown -o classification-diff.md
+      - name: Upload diff
+        uses: actions/upload-artifact@v2
+        with:
+          name: classification-diff.md
+          path: src/ontology/classification-diff.md
+  post_comment:
+    needs: [diff_classification, edit_file]
+    runs-on: ubuntu-latest
+    container: docker://dvcorg/cml-py3:latest
```
matentzn, line 86: Why is this needed? Out of curiosity.

matentzn (approved):
Awesome.. Looks impressive! I would change the nomenclature throughout (we don't use the term `master` anymore unless we have to, for example in the branches: section).

.github/workflows/diff.yml
shawntanzk: oh whoops this isn't, was testing with something and totally forgot to remove it. @ubyndr removed in here the last time, but I forgot to remove it on my test fork and copied and paste the whole thing - will remove

shawntanzk:
> Do we really need to use obolibrary/odkfull:v1.2.32

pretty sure we need it for the make step, the rest I think lite is fine

matentzn:
odklite is sufficient

shawntanzk:
If all is good, should I merged this in for CL or just wait for roll out with ODK?

matentzn:
Either way is fine. I will compile it soon and make a PR on CL

shawntanzk:
Will merge this in then :) no harm and all

PR #1449, opened 2022-02-16
2022-02-16 Release

shawntanzk:
Making a release to get the https://github.com/obophenotype/cell-ontology/pull/1446 into a release. See https://github.com/obophenotype/cell-ontology/issues/1359 for reason. 

@bvarner-ebi see https://obophenotype.github.io/cell-ontology/odk-workflows/ReleaseWorkflow/#review-the-release for how to review a release.

PR #1450, opened 2022-02-16
Added back EC axioms to CL_0000151

shawntanzk:
See https://github.com/obophenotype/cell-ontology/pull/1449


shawntanzk:
btws @dosumis - this is a super good example of the robot diff on edit file and reasoned file and a case for why we need both at all times :)

bvarner-ebi (approved):

Issue #1306, opened 2021-11-11
Dangling CARO term and CARO incompatibility with CL 


shawntanzk:
Currently, there is a dangling CARO term (CARO_0010000) that is used by a RO term (directly develops from) that is used in UBERON term (equiv class for maxillary artery) that used by another uberon term (deep auricular artery) that isn't used by any term except itself (I think). 
I tried to solve the dangling term by doing and import of that CARO term, however when trying to test a release, CARO seems to be incompatible with CL -> error saying that there are unsatisfiable classes. 
I'm not sure if it would be sensible, or for that matter how to, remove some terms in the uberon import? 
I'm also not sure if it is an issue that CARO isn't compatible with CL and do we have to chase down a reason? 
Or if it's just not worth it and we just leave the dangling term? 
Related to https://github.com/obophenotype/uberon/issues/1963

shawntanzk:
## [latissimus dorsi muscle](http://purl.obolibrary.org/obo/UBERON_0001112) SubClassOf [Nothing](http://www.w3.org/2002/07/owl#Nothing) ##

  - [latissimus dorsi muscle](http://purl.obolibrary.org/obo/UBERON_0001112) SubClassOf [has muscle antagonist](http://purl.obolibrary.org/obo/RO_0002568) some [deltoid](http://purl.obolibrary.org/obo/UBERON_0001476)
    - [deltoid](http://purl.obolibrary.org/obo/UBERON_0001476) SubClassOf [innervated_by](http://purl.obolibrary.org/obo/RO_0002005) some [axillary nerve](http://purl.obolibrary.org/obo/UBERON_0001493)
      - [axillary nerve](http://purl.obolibrary.org/obo/UBERON_0001493) SubClassOf [part_of](http://purl.obolibrary.org/obo/BFO_0000050) some [brachial nerve plexus](http://purl.obolibrary.org/obo/UBERON_0001814)
        - [brachial nerve plexus](http://purl.obolibrary.org/obo/UBERON_0001814) EquivalentTo [spinal nerve plexus](http://purl.obolibrary.org/obo/UBERON_0001813) and ([innervates](http://purl.obolibrary.org/obo/RO_0002134) some [forelimb](http://purl.obolibrary.org/obo/UBERON_0002102))
          - [brachial nerve plexus](http://purl.obolibrary.org/obo/UBERON_0001814) SubClassOf [neuron projection bundle](http://purl.obolibrary.org/obo/CARO_0001001)
[...]

PR #1461, opened 2022-02-21
Fixes issues suggested in #1306

shawntanzk:
see https://github.com/obophenotype/cell-ontology/issues/1306#issuecomment-1012220370


paolaroncaglia:
Perhaps @matentzn could kindly review? Thanks in advance. (@shawntanzk would like another set of eyes before pushing his pending PRs, but this is a tech issue so my contribution is limited. The PR does have a conflict though.)

matentzn (approved):
Formally this is all correct. I have _not_ looked at the object properties you used for the logical definitions!

Issue #1466, opened 2022-02-21
Please import UBERON:0007221 neonate stage into CL


paolaroncaglia:
Needed for https://github.com/obophenotype/cell-ontology/issues/979#issuecomment-1047003395. Thank you!

paolaroncaglia:
@shawntanzk , may I assign this to you please? No rush. Thanks in advance.

shawntanzk:
will do after uberon release and updating CL mirror

PR #1401, opened 2022-01-12
[DRAFT - DO NOT MERGE] glomerular visceral epithelial cell part of glomerular visceral epithelium

shawntanzk:
Fixes #1399 
@bvarner-ebi this branch has the term you need imported, once you added everything you can convert this draft PR into a PR and remove the DRAFT from the title and get it reviewed

PR #1469, opened 2022-02-21
imported UBERON:0007221

shawntanzk:
refreshed uberon mirror and imported UBERON:0007221
Happy for whoever reviews this to merge in
Fixes https://github.com/obophenotype/cell-ontology/issues/1466
Related to https://github.com/obophenotype/cell-ontology/pull/1401#issuecomment-1046883764


paolaroncaglia (commented):
@shawntanzk thank you, the import of the Uberon term I needed looks good. There's a ton of diffs to Uberon imported terms - I skimmed the list and the ones I looked at are ok; I presume that all Uberon diffs were looked at as part of the Uberon release, so no need to check all of them here, right?
@bvarner-ebi I see that you're tagged in this review too because one of your tickets was impacted, but not sure what we should check for, could you take a look please?
I'm happy to merge when both questions are addressed :-) Thanks!

shawntanzk:
The changes are due to 1) uberon mirror being refreshed (so all the changes are in there), and 2) I'm guessing the term I pulled it also pulled in other terms that it is dependent on.

> not sure what we should check for

mainly that the addition/change you want is in this ticket. I did a quick check in the reasoned diff and saw that there were changes to the term @bvarner-ebi wanted - just probably good to check that its the correct change lol. I also checked that uberon:0007221 is in there :)

paolaroncaglia:
@shawntanzk thanks, yes I checked and what I need is in there :-) I'll wait for Bradley to check too, no rush.

bvarner-ebi (approved):
I see the changes requested from https://github.com/obophenotype/uberon/issues/2230. Thank you, @shawntanzk!

PR #1483, opened 2022-03-04
Update resolving_merge_conflicts.md

bvarner-ebi:
Initial edits

### DIFF
--- a/docs/resolving_merge_conflicts.md
+++ b/docs/resolving_merge_conflicts.md
@@ -10,11 +10,14 @@ Occassionally non-trivial clashes will happen when two pull requests include edi
 
 1. Reserialise the Master file using the ODK commands normalize_src or normalize_obo_src. These commands require installation of ROBOT. If not already installed, follow [the instructions here](http://robot.obolibrary.org).
 
-2. At the line command (PC) or Terminal (Mac), use the cd (change directory) command to navigate to the repository's src/ontology directory.
+2. At the line command (PC) or Terminal (Mac), use the cd (change directory) command to navigate to the repository's src/ontology/ directory.
 For example,
+
 '''
-cd obophenotype/cell-ontology/src/ontology
+cd PATH_TO_ONTOLOGY/src/ontology/
 '''
+
+Replace "PATH_TO_ONTOLOGY" with the actual file path to the ontology. If you need to orient yourself, use the '''pwd''' (present working directory) or '''ls''' (list) line commands.
 
 3. If you are resolving a conflict in an .owl file, run:
 
--- a/docs/resolving_merge_conflicts.md
+++ b/docs/resolving_merge_conflicts.md
@@ -8,19 +8,36 @@ Occassionally non-trivial clashes will happen when two pull requests include edi
 
 ### SOP.
 
-1. Reserialise the Master file using ROBOT. This will fix any odd orderings that may have been introduced during previous merges.
-    ```sh
-    sh ./run.sh ROBOT convert --input cl-edit.owl -f ofn --output cl-edit.owl
-   ```
+1. Reserialise the Master file using the ODK commands normalize_src or normalize_obo_src. These commands require installation of ROBOT. If not already installed, follow [the instructions here](http://robot.obolibrary.org).
+
+2. At the line command (PC) or Terminal (Mac), use the cd (change directory) command to navigate to the repository's src/ontology directory.
+For example,
+'''
+cd obophenotype/cell-ontology/src/ontology
+'''
+
+3. If you are resolving a conflict in an .owl file, run:
+
+'''
+make normalize_src
+'''
+
+If you are resolving a conflict in an .obo file, run:
+
+'''
+make normalize_obo_src
+'''
+
+
 4. In GitHub Desktop:
 
-   * Checkout Master and pull to make sure your Master is up to date.
-   * Checkout the branch for the pull request & make sure it is up to date.
+   * Checkout Master and pull to make sure your Master branch is up to date.
+   * Checkout the branch for the pull request and make sure it is up to date.
    * Choose Branch > Update from master: ![image](https://user-images.githubusercontent.com/112839/112127621-89af9f00-8bbd-11eb-8613-f3a2b8166085.png)
 
-   * GitHub desktop should detect the clash and ask you if you want to open in your text editor of choice (Atom is a good choice)
-   * If clashes are due to trivial ordering problems, delete conflict marks, commit and push back to GitHub.
-   * Check the resulting diffs on the Pull Request on GitHub
-   * Once tests have run, if successful you can merge and delete the branch.
+   * GitHub desktop should detect the clash and ask you if you want to open in your text editor of choice (e.g., Atom).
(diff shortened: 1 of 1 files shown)


bvarner-ebi:
Once we have a new merge conflict, I will test these updates to confirm they resolve the conflict.

docs/resolving_merge_conflicts.md
```diff
@@ -1,26 +1,46 @@
...
-Occassionally non-trivial clashes will happen when two pull requests include edits to the same term or even the same axiom.  Ask an editor for help if you don't feel confident resolving these.  
+Occassionally non-trivial clashes will happen when two pull requests include edits to the same term or even the same axiom. Ask an editor for help if you don't feel confident resolving these.  
 
 ### SOP.
 
-1. Reserialise the Master file using ROBOT. This will fix any odd orderings, that may have been introduced during previous merges.
-    ```sh
-    sh ./run.sh ROBOT convert --input cl-edit.owl -f ofn --output cl-edit.owl
-   ```
+1. Reserialise the Master file using the ODK commands normalize_src or normalize_obo_src. These commands require installation of ROBOT. If not already installed, follow [the instructions here](http://robot.obolibrary.org).
```
shawntanzk, line 11: I think this should be "requires ODK". The commands are in the ODK container not in ROBOT.

docs/resolving_merge_conflicts.md
```diff
@@ -1,26 +1,46 @@
...
+
+
 4. In GitHub Desktop:
 
-   * Checkout Master and pull to make sure your Master is up to date.
-   * Checkout the branch for the pull request & make sure it is up to date.
-   * Choose Branch.update_from_master: ![image](https://user-images.githubusercontent.com/112839/112127621-89af9f00-8bbd-11eb-8613-f3a2b8166085.png)
+   * Checkout Master and pull to make sure your Master branch is up to date.
+   * Checkout the branch for the pull request and make sure it is up to date.
+   * Choose Branch > Update from master: ![image](https://user-images.githubusercontent.com/112839/112127621-89af9f00-8bbd-11eb-8613-f3a2b8166085.png)
```
shawntanzk, line 39: its ok for now, but when you move it to obook, could you download the picture and place it as an actual image file please

shawntanzk (approved):
See minor comment.

docs/resolving_merge_conflicts.md
bvarner-ebi: ODK deals with docker, right? I was able to run the commands without docker, like how I run the make normalise_xsd_string command.
When I initially tried this, 'make normalize_scr' would not run because my ROBOT path was not set up correctly. It gave an error that 'robot' was missing. Once that was fixed, the command ran. So it seems the minimum requirement might be to have ROBOT or ODK?

shawntanzk: normalise_xsd_string is a custom command for CL (see cl.Makefile):
```
normalise_xsd_string: $(SRC)
	sed -i.bak -E "s/Annotation[(](oboInOwl[:]hasDbXref [\"][^\"]*[\"])[)]/Annotation(\1^^xsd:string)/g" $<
	rm $<.bak
```

For noramlise_src it is ODK, but you are running it through Makefile 
Makefile is deployed through ODK into repos, however, noramlise_src requires ROBOT
```
.PHONY: normalize_src
normalize_src: $(SRC)
	$(ROBOT) convert -i $< -f ofn -o $(TMPDIR)/normalise && mv $(TMPDIR)/normalise $<
```
ROBOT is in docker odk, so if you run `sh run.sh` (which calls on docker, you can see this in the run.sh file) instead of `make` it should be there already (makes it easier for people). 

Does that make sense? took me forever to understand this whole concept, and im not even 100% sure i fully understand it lol

bvarner-ebi: Are these the pre-req steps to run these commands? https://oboacademy.github.io/obook/howto/odk-setup/

I try to stay away from docker when possible due to memory issues. If there is a way to limit these steps to the line command, we should include it as an alternate option at least.

When I open docker and run: sh ./run.sh ROBOT convert --input cl-edit.owl -f ofn --output cl-edit.owl, I get this message:

docker: Error response from daemon: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "ROBOT": executable file not found in $PATH: unknown.

I might not have the wrapper script set up properly and will check on that.

This illustrates the situation where there seem to be multiple prerequisites before following the instructions as written. Is the odk-setup link above the best place to start?

shawntanzk: > docker: Error response from daemon: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "ROBOT": executable file not found in $PATH: unknown.

ROBOT should not be caps > `sh ./run.sh robot convert --input cl-edit.owl -f ofn --output cl-edit.owl`

bvarner-ebi: Running 

sh ./run.sh robot convert --input cl-edit.owl -f ofn --output cl-edit.owl

did not give an error (robot in lower case). Thanks! I don't know how to confirm it works for solving merge conflicts without an example, but at least there was no error.

Of note, I have to run make normalise_xsd_string to resolve the xsd issue after running normalize_src.

shawntanzk: huh, I guess saving it as is using ROBOT strips the xsd:string stuff, wonder if this is the case in other ontologies - might need to consider doing a normalise step automatically too - let me go take a look

shawntanzk: tried of bspo which uses owl edit file (never noticed how most of the other ontologies i work with use obo edit files) and it seems fine, guess its a CL thing hence the custom cmd in CL - for this document you can add to do that given it is CL specific

bvarner-ebi: Thank you for checking!
So it's clear, is running...

sh ./run.sh robot convert --input cl-edit.owl -f ofn --output cl-edit.owl

... all that is needed, or does the editor have to run the normalize_(obo)_src command in addition to that?

shawntanzk: running `sh run.sh noramlize_src` (or `sh run.sh normalize_obo_src`) is all that is neeeded -> that command basically runs `robot convert --input cl-edit.owl -f ofn --output cl-edit.owl`

For CL, just note that you have to run normalise_xsd_string after I guess

bvarner-ebi: Thanks, @shawntanzk. For setting up docker, It looks like odk.sh needs to be in the src/ontology/ directory for this to work (per the instructions here: https://oboacademy.github.io/obook/howto/odk-setup/).

Is that correct? I'm surprised it already isn't there if that is the case.

bvarner-ebi: `sh run.sh normalize_src`

`docker: Error response from daemon: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "normalize_src": executable file not found in $PATH: unknown.`

shawntanzk: `sh run.sh make normalize_src`

shawntanzk:
@bvarner-ebi - dont forget to add a version of this that is generalisable to https://github.com/OBOAcademy/obook/blob/master/docs/howto/fixing-conflicts.md yeah

Issue #1481, opened 2022-03-03
Fixing non-RO relation


shawntanzk:
Related to https://github.com/obophenotype/cell-ontology/issues/1262

Currently there are cl specific relations which should be either 1) removed, or 2) moved to RO 

Level | Number of errors
-- | --
http://purl.obolibrary.org/obo/cl#has_completed | 1
http://purl.obolibrary.org/obo/cl#has_high_plasma_membrane_amount | 1
http://purl.obolibrary.org/obo/cl#has_low_plasma_membrane_amount | 1
http://purl.obolibrary.org/obo/cl#has_not_completed | 1
http://purl.obolibrary.org/obo/cl#lacks_part | 1
http://purl.obolibrary.org/obo/cl#lacks_plasma_membrane_part | 1
http://semanticscience.org/resource/SIO_000658 | 1

shawntanzk:
http://purl.obolibrary.org/obo/cl#lacks_plasma_membrane_part seems to be a shorthand, should this moved to RO? 
![Screenshot 2022-03-04 at 11 46 16](https://user-images.githubusercontent.com/76212760/156757876-0bd9c32c-5e00-41ba-a4f4-352ec6d8e343.png)

shawntanzk:
http://semanticscience.org/resource/SIO_000658 - not used directly in CL, should be removed

shawntanzk:
- [ ] RO release
- [ ] change 'has high plasma membrane amount' to  RO_0015015
- [ ] change 'has low plasma membrane amount' to RO_0015016

bvarner-ebi:
closing per @shawntanzk's direction

PR #1485, opened 2022-03-07
remove SIO_000658

shawntanzk:
Related to https://github.com/obophenotype/cell-ontology/issues/1481#issuecomment-1059093183

@matentzn - I know you don't do content PR, putting you as a reviewer as it is related to fixing non-RO relation


matentzn (approved):
If its not used I dont consider it content - just cleanup :) Great!

Issue #81, opened 2015-06-04
Add syns and xrefs for subtypes of glomerular visceral epithelial cell


GoogleCodeExporter:
```
{meso,pro} nephric podocyte

and add xref to ZFA

 ZFA:0001673 ! pronephric podocyte
 ZFA:0001674 ! mesonephric podocyte

```

Original issue reported on code.google.com by `cmung...@gmail.com` on 3 Jul 2013 at 9:32

paolaroncaglia:
@bvarner-ebi this is an old ticket about renal cell types with references to non-human vertebrates (ZFA = zebrafish anatomy). As you're going to look into the kidney domain for ASCT+B, it'd be helpful if you could address this ticket too please, to ensure consistency/avoid duplication. If you're ok with that please unassigned me. Thank you!

bvarner-ebi:
Hi, @cerivs, checking in to see if you find the plan above acceptable. Upon approval, I will proceed.

cc @paolaroncaglia, in case you are OK with approving

paolaroncaglia:
@bvarner-ebi your proposal sounds good to me. Thank you.

bvarner-ebi:
@paolaroncaglia, I also see several GO terms that use the original nomenclature, e.g., GO:0061257 'mesonephric glomerular visceral epithelial cell development'.

Should these terms be updated as well? E.g., the above term label would be updated to 'mesonephric podocyte development' and the original label would be added as an exact synonym. Once edited in GO, would those term label changes be reflected in CL after an import refresh?
Should I submit a ticket for a GO editor to make those edits?

dosumis:
I think worth submitting a GO ticket to see if they want to update their term names.

bvarner-ebi:
> I think worth submitting a GO ticket to see if they want to update their term names.

https://github.com/geneontology/go-ontology/issues/23000

raymond91125:
Shouldn't CL:0002525 metanephric glomerular visceral epithelial cell -> metanephric podocyte?

bvarner-ebi:
> Shouldn't CL:0002525 metanephric glomerular visceral epithelial cell -> metanephric podocyte?

Thanks for catching that, @raymond91125. I will update the commit.

PR #1496, opened 2022-03-11
Address #81 Add syns and xrefs for subtypes of glomerular visceral ep…

bvarner-ebi:
Address #81 Add syns and xrefs for subtypes of glomerular visceral epithelial cell
Before any potential GO edits


bvarner-ebi:
@paolaroncaglia, I know the comments in #81 instruct to use the narrow synonym type, but to align with CL:0000653 podocyte, I set these as exact synonyms. Do you agree?

PR #1497, opened 2022-03-11
Add human_reference_atlas subset

anitacaron:


matentzn:
I cant judge this! You are saying "subset", but you created a component.. I also don't see the related changes to the Makefile!

dosumis:
@matentzn wrote:
> I cant judge this! You are saying "subset", but you created a component.. I also don't see the related changes to the Makefile!

This file is created from a robot template via a build job on https://github.com/hubmapconsortium/ccf-validation-tools. (template is not cached).  It contains only subset and present_in_taxon axioms (to Homo sapiens).  I think it makes sense as a component as it needs to be present in the editors file and available for all release files.  RE:Makefile - i thought it would be sufficient to have an import statement on the editor's file, or does the component need to be explicitly named in the Makefile?

dosumis (approved):
I'm happy with it if @matentzn is.  Please also see https://github.com/obophenotype/uberon/pull/2346#pullrequestreview-917601946 - which does the same for Uberon.

matentzn:
Same concerns as uberon, otherwise fine

dosumis:
Looking good.  Is this ready to merge?

.github/workflows/docs.yml
```diff
@@ -5,6 +5,9 @@ name: Docs
 on:
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
+  push:
```
matentzn, line 8: dont commit

matentzn (approved):
Just dont commit the docs.yml file, else its fine

.github/workflows/docs.yml
anitacaron: done

PR #1499, opened 2022-03-16
Create GitHub.md

bvarner-ebi:
This is a proposed readme for an editor who is new to GitHub. Planning to bring it into OBOOK as we work on consolidating documentation.


shawntanzk:
don't think there is a need for this in CL at all, esp given it isn't CL specific. Think its better if this goes straight in obook.

bvarner-ebi:
> don't think there is a need for this in CL at all, esp given it isn't CL specific. Think its better if this goes straight in obook.

https://github.com/OBOAcademy/obook/pull/148

PR #1525, opened 2022-03-28
Fixes #1524 corneal epithelial cell should be part_of the corneal epi…

### LINKED ISSUE #1524: corneal epithelial cell should be part_of the corneal epithelium

bvarner-ebi:
Fixes # 1524 corneal epithelial cell should be part_of the corneal epithelium


bvarner-ebi:
@shawntanzk, @paolaroncaglia, could I get your guidance on an unexpected inference that I'm seeing?
After running the reasoner, I don't understand why the inferred 'squamous epithelial cell' EA appears. No explanations appear when I check.

Screenshot 2022-03-28 at 07 06 19

shawntanzk:
@bvarner-ebi - your change here seems to be on epithelial cell, making it part_of corneal epithelium

shawntanzk:
> inferred 'squamous epithelial cell' EA appears

No idea, can't replicate this on my side, but given that squamous epithelial cell is in equiv class, you should remove the subclassOf axiom

bvarner-ebi:
> @bvarner-ebi - your change here seems to be on epithelial cell, making it part_of corneal epithelium - which is 100% not right

@shawntanzk, thank you... I did not mean to do that, but now see that I made a change to an axiom under "SubClass Of (Anonymous Ancestor)" instead of adding under "SubClass Of". Thank you!

shawntanzk (approved):

PR #1529, opened 2022-04-01
Remove unmerge component

### LINKED ISSUE #1376: Reinstating range constraints on 'has characteristic' and 'regulates'
Currently to fix import issues that create unsat classes, we need to remove range constraints on 'has characteristic' and 'regulates'
This ticket is a reminder to remove this hack when the issues in GO are resolved

related to https://github.com/geneontology/go-ontology/issues/16524

@rays22 (or @dosumis) not sure if the above is the only ticket for this? or if there are tickets in GO to fix “has participant” and “has characteristic” range issues? - if so could you link them here please?

shawntanzk:
refreshed GO mirror, merged import, and removes unmerge component. Looks alright in protege - no unsats.
Fixes #1376


shawntanzk:
automated diff failed cause diff is too long to post

paolaroncaglia:
@shawntanzk looks like the checks were stalled, you may want to close the PR and open it again to trigger them.
Then you could ping the assigned reviewer ;-)

paolaroncaglia:
@matentzn actually could you please review this? @shawntanzk wrote that

> automated diff failed cause diff is too long to post

so should be ok to review? Thanks!

shawntanzk:
> so should be ok to review? Thanks!

yes :) its just the diff being too big, nothing can be done about it

matentzn:
hmmmm.. a bit worrysome that the diff is _that_ large, but I guess import refreshments will do that.

@shawntanzk would it be a good idea to actually, instead of diffing uberon.owl, diffing uberon-simple.owl? This way we get the reasoning consequences etc, but without changes to the imports, and only see stuff related to uberon classes..

shawntanzk:
yeah its 100% the reasoned diff - with import changes, the diff tends to end up too big. In these cases i usually just look at the actual diff (like we used to do before). 
Happy to change the diff to use simple instead - will do it in another PR :)

src/ontology/cl.Makefile
```diff
@@ -279,7 +279,6 @@ imports/merged_import.owl: mirror/merged.owl imports/merged_terms_combined.txt
 	if [ $(IMP) = true ]; then $(ROBOT) merge -i $< \
```
matentzn, line 279: Do we still need the imports/merged_import.owl goal in cl.Makefile? is it not identical with the one in `Makefile`?

shawntanzk: good point - just checked, makefile removes CLO in addition to what is removed, which I imagine is a good thing, will remove this goal for cl.makefile

PR #1530, opened 2022-04-04
Add obsolete replaced_by check

anitacaron:
Related to INCATools/ontology-development-kit#512

In CL, some of the `replaced_by` property is inside a comment, so this check won't work as expected.

Screenshot 2022-04-04 at 11 38 06


shawntanzk:
>  some of the replaced_by property is inside a comment

This is 100% wrong and should be replaced by a replaced_by lol - can we replace them accordingly (yes I know that's a lot of "replace" lol)

PR #1543, opened 2022-04-08
Change actions diff to use simple

shawntanzk:
Related to https://github.com/obophenotype/cell-ontology/pull/1529#issuecomment-1092670870


matentzn (approved):
Sweet, I agree!

Issue #1545, opened 2022-04-11
[Typo/Bug] Incomplete classification under animal cell


dosumis:
**CL term**

Incomplete classification under animal cell.  Terms should be moved manually:
![image](https://user-images.githubusercontent.com/112839/162786285-b7e55d1c-9ab9-4a2e-bfaf-b2116aa0f534.png)

PR #1546, opened 2022-04-12
Addresses #1545 Incomplete classification under animal cell #1545

bvarner-ebi:
Addresses #1545 Incomplete classification under animal cell #1545

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -2789,7 +2789,7 @@ AnnotationAssertion(rdfs:label obo:IAO_0000115 "definition"^^xsd:string)
 
 AnnotationAssertion(rdfs:label obo:IAO_0000424 "expand expression to"^^xsd:string)
 
-# Annotation Property: obo:IAO_0000700 (preferred_root)
+# Annotation Property: obo:IAO_0000700 (has ontology root term)
 
 AnnotationAssertion(rdfs:label obo:IAO_0000700 "preferred_root")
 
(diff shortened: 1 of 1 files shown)


bvarner-ebi:
'stem cell' is now a subclass of 'animal cell'.

The other two terms shown in #1545 seem to already fall under 'animal cell'.

From OLS:

Screenshot 2022-04-12 at 06 08 08

Screenshot 2022-04-12 at 06 06 06

Is additional modelling required, @dosumis?

src/ontology/cl-edit.owl
```diff
@@ -2789,7 +2789,7 @@ AnnotationAssertion(rdfs:label obo:IAO_0000115 "definition"^^xsd:string)
 
 AnnotationAssertion(rdfs:label obo:IAO_0000424 "expand expression to"^^xsd:string)
 
-# Annotation Property: obo:IAO_0000700 (preferred_root)
```
shawntanzk, line 2792: This is a super weird change, its not an issue I think cause rdfs label is still there, but just odd.

shawntanzk (approved):

shawntanzk:
btws @dosumis is there a reason why animal cell uses 'only in taxon' instead of 'in taxon'? That's the reason its not autoclassifying yeah?

src/ontology/cl-edit.owl
dosumis: That is weird.  It's in a commented line so will not cause any problems. I think these are written by serialiser for human readability.  I wonder if this is a sign that the IAO term has two labels - perhaps with one coming from an import.  Anyway - Definitely not something to worry about for this PR, but might be worth investigating.

dosumis:
>  @dosumis is there a reason why animal cell uses 'only in taxon' instead of 'in taxon'? That's the reason its not autoclassifying yeah?

We should probably move _en masse_ to using 'in taxon'

Issue #1557, opened 2022-04-24
Fix MeSH cross-references


bgyori:
Similar to UBERON (see https://github.com/obophenotype/uberon/issues/698), CL contains a number of cross-references to MESH that are represented as tree numbers rather than IDs. These should ideally be replaced where possible and I will open a PR to do so.

However, as also pointed out in https://github.com/obophenotype/cell-ontology/issues/146, many CL terms do not actually refer to a corresponding MESH term as a usual xref, but rather refer to it as part of the definition. Here is an example with CL:0000034 in OWL and OBO, respectively:
```
# Class: obo:CL_0000034 (stem cell)

AnnotationAssertion(Annotation(oboInOwl:hasDbXref "GOC:tfm"^^xsd:string) Annotation(oboInOwl:hasDbXref "MESH:A11.872"^^xsd:string) obo:IAO_0000115 obo:CL_0000034 "A relatively undifferentiated cell that retains the ability to divide and proliferate throughout life to provide progenitor cells that can differentiate into specialized cells."^^xsd:string)
AnnotationAssertion(oboInOwl:hasDbXref obo:CL_0000034 "CALOHA:TS-2086"^^xsd:string)
AnnotationAssertion(oboInOwl:hasDbXref obo:CL_0000034 "FMA:63368"^^xsd:string)
...
```

```
[Term]
id: CL:0000034
name: stem cell
namespace: cell
def: "A relatively undifferentiated cell that retains the ability to divide and proliferate throughout life to provide progenitor cells that can differentiate into specialized cells." [GOC:tfm, MESH:A11.872]
comment: This term applies to metazoan. For plant stem cells, consider using PO:0004011 ‘initial cell’ or its parent PO:0004010 ‘meristematic cell’.
xref: CALOHA:TS-2086
xref: FMA:63368
...
```

Is there some rationale behind putting references to MESH in the definition rather than the xref list? I would assume that most downstream code that aims to find xrefs would overlook these.

cthoyt:
It might be more specifically denoting that the text was borrowed from the MeSH record, but I agree, in this case it would be nice to have the xref also on the node level since these two places for xrefs can be interpreted differently.

dosumis:
> It might be more specifically denoting that the text was borrowed from the MeSH record

Correct.


>  but I agree, in this case it would be nice to have the xref also on the node level since these two places for xrefs can be interpreted differently.

+1.

[...]

PR #1558, opened 2022-04-24
Replace MeSH tree numbers with IDs

bgyori:
This PR replaces most MESH tree numbers with IDs appearing in definitions and xrefs. This mostly addresses #1557.

First, I mapped all tree numbers to MESH IDs where this was possible (i.e., the tree number is still valid in the current release of MeSH). Second, for cases where this was not possible, I used Gilda (https://github.com/indralab/gilda) to find an appropriate match in MeSH. Overall, the statistics are as follows:
- A total of 98 MeSH tree numbers appear in definitions or xrefs
- 54 of these could be mapped to IDs using a simple lookup in the current MeSH release
- 30 of the remaining ones could be mapped by Gilda
- 14 tree numbers remain unmapped, though it looks like these could be worked out with some manual digging

An overview spreadsheet to help review the changes is here: https://github.com/bgyori/cell-ontology/blob/329a9ce9b0a6b53feeea4ed984dfe503f6fd20cf/cl_mesh_mappings.csv

I also wanted to mention here as well the question in #1557, namely, whether it makes sense that these MESH references often (only) appear in the definition instead of the xref list. Should this be changed?


bgyori:
I found a few suspicious mappings where the original tree number (probably due to MeSH having changed over time) is not mapped to an appropriate ID currently. I will push some manual overrides for these cases.

bgyori:
After some manual improvements, the final tally is:
- A total of 98 MeSH tree numbers appear in definitions or xrefs
- 53 of these could be mapped to IDs using a simple lookup in the current MeSH release
- 31 of the remaining ones could be mapped by Gilda
- 8 of them, I mapped manually
- 6 tree numbers remain unmapped, where I wasn't confident enough in a good replacement

rays22 (approved):
Looks fine now, but
1. I have resolved 11 conflicts in `cl-edit.owl` manually.
2. CL:0002211 type I muscle cell --> I have changed 2 *broad synonym* to *exact synonym* annotations when `hasDbXref` [MESH:D018657 Muscle Fibers, Slow-Twitch](https://meshb.nlm.nih.gov/record/ui?ui=D018657)

PR #1559, opened 2022-04-25
Added CARO import

### LINKED ISSUE #1306: Dangling CARO term and CARO incompatibility with CL

### LINKED ISSUE #1473: Random unresolvable CARO term in CL hierarchy
http://www.ontobee.org/ontology/CL?iri=http://purl.obolibrary.org/obo/CARO_0000006

image

http://purl.obolibrary.org/obo/CARO_0000006 is itself unresolvable

shawntanzk:
Fixes #1473 
Fixes #1306

I think everything should run fine, will let the QC make sure of it (want to make sure esp that the no obsolete term issue is fixed in CARO)

Just a heads up, it does have a big of ugliness cause of CARO and CL having same terms: 
Screenshot 2022-04-25 at 09 53 08


shawntanzk:
`make: *** [cl.Makefile:83: oort] Error 255`

Will check up on this

shawntanzk:
Trying to figure out the issue: 

```
%       sh run.sh fastobo-validator cl-check.obo

     Parsing `cl-check.obo`
      Failed parsing `cl-check.obo`
           --> cl-check.obo:672:38
            |
        672 | synonym: "Acetylcholine" EXACT [KEGG COMPOUND:]␊
            |                                      ^---
            |
            = expected QuotedString
```

`KEGG COMPOUND` appears 86 times in cl-check.obo - guessing the space is the issue. 
Not sure where in cl-edit this is from, can't find it in cl-edit, which is strange given cl-check.obo comes from 
`sh run.sh robot merge -i cl-edit.owl convert -f obo --check false -o cl-check.obo`

Issue #573, opened 2019-07-18
Review loss of axioms


matentzn:
- SubClassOf(<http://purl.obolibrary.org/obo/CL_0000108> <http://purl.obolibrary.org/obo/CL_0000381>)
- SubClassOf(<http://purl.obolibrary.org/obo/CL_0000125> ObjectSomeValuesFrom(<http://purl.obolibrary.org/obo/RO_0002202> <http://purl.obolibrary.org/obo/CL_0000030>))
- SubClassOf(<http://purl.obolibrary.org/obo/CL_0008016> ObjectSomeValuesFrom(<http://purl.obolibrary.org/obo/RO_0002202> <http://purl.obolibrary.org/obo/CL_0008012>))

(Redundancy stripping?)

paolaroncaglia:
@shawntanzk @gouttegd  I'm reviewing a previous note by @matentzn investigating loss of axioms. The top one is about neurons:

> SubClassOf(http://purl.obolibrary.org/obo/CL_0000108 http://purl.obolibrary.org/obo/CL_0000381)

Should 'cholinergic neuron' be a subclass of 'neurosecretory neuron'? The latter is asserted as being an efferent neuron: is that correct across taxa? ('neurosecretory neuron' has an FBbt xref.) Are all cholinergic neurons efferent? I'm not convinced that we should add that subclass.

paolaroncaglia:
@dosumis 
The original request in this ticket was a note from Nico to review loss of 3 axioms. The last one pertains to 2 terms that you created. The lost axiom is 
'activated skeletal muscle satellite cell' develops from 'quiescent skeletal muscle satellite cell'
Their common parent 'skeletal muscle satellite cell' is textually defined as "...These cells are mostly quiescent, but upon activation they divide to produce cells that generate new muscle fibers." and 'activated skeletal muscle satellite cell''s def says "A skeletal muscle cell that has become mitotically active - typically following muscle damage.". Therefore, I believe that the lack/loss of that axiom is correct. Could you confirm please?

shawntanzk:
> SubClassOf(http://purl.obolibrary.org/obo/CL_0000108 http://purl.obolibrary.org/obo/CL_0000381)

[...]

PR #1570, opened 2022-04-28
Edit def of CL_0008016 "activated skeletal muscle satellite cell'

paolaroncaglia:
Addresses #573


paolaroncaglia:
@shawntanzk please feel free to merge after you've reviewed, if all is well with the RO terms, thanks!

src/ontology/cl-edit.owl
```diff
@@ -2728,6 +2728,8 @@ Declaration(ObjectProperty(obo:RO_0002202))
 Declaration(ObjectProperty(obo:RO_0002215))
 Declaration(ObjectProperty(obo:RO_0002292))
 Declaration(ObjectProperty(obo:RO_0013001))
+Declaration(ObjectProperty(obo:RO_0015015))
+Declaration(ObjectProperty(obo:RO_0015016))
```
shawntanzk, line 2732: @dosumis @matentzn do OPs have to be declared? I thought imports didnt need to be declared? Is it different for OPs? It's not a huge issue just curious.

matentzn: Dont worry about it. Redundant declarations are common and not harmful. To answer your question: they do not have to be declared, if they are declared in the imported ontology!

Issue #1128, opened 2021-06-21
[Class hierarchy] small, large and nest basket cells might be under neocortex basket cell


shawntanzk:
From the description of small, large and nest basket cells, they sound like they could possibly be cortex specific. 
If so, these terms should be under neocortex basket cell.
Would like some input on this @tgbugs, thanks.

Related to #1127

dosumis:
'large basket cell' def is clearly cortex specific: "A basket cell that is large, and typically ascends to give rise to many long horizontally and vertically projecting axon collaterals that traverse neighboring columns and can extend through all cortical layers."

Mention of layer here sounds cortex specific: "A basket cell with axonal arbors composed of frequent, short, curvy axonal branches that tend to be near their somata and within the same layer."

Wikipedia (& reference therein) has these 3 types as cortex specific: https://en.wikipedia.org/wiki/Basket_cell#Cortex


@tgbugs - this is consistent with Wikipedia https://en.wikipedia.org/wiki/Basket_cell#Cortex

tgbugs:
Sorry, just seeing this. The one area where the neocortex axioms would be problematic is that there are basket cells in hippocampus, so basket cell usually applies neocortex and hippocampus, whereas neocortex basket cell is the differentiated form for neocortex only. That said, all of these types are coming from neocortex only, and Rat area S1 specifically. So the change to include the part of neocortex axiom is ok.

PR #1574, opened 2022-05-03
added neocortex location to 3 basket cell types

shawntanzk:
Fixes https://github.com/obophenotype/cell-ontology/issues/1128
From reading, I think that these basket cells are in neocortex. There are basket cells in cerebellum too (and they already have a class in CL) but they seem to be another type that are multipolar. I didn't create a new class for them as they would be the only child on under cerebellum basket cells, so there isnt a need for it.


patrick-lloyd-ray (commented):
Looks good to me.

PR #1575, opened 2022-05-03
Added logical defs for spiral ganglion subtypes

### LINKED ISSUE #1319: Add logical defs for subtypes of spiral ganglion neurons
From @cmungall in relation to subtypes of spiral ganglion neuron:
" Judging by the text definition I would be tempted to make a logical def, but first a DP for cell types differentiated by what they innervate."
@shawntanzk will look into this
Related to closed PR
https://github.com/obophenotype/cell-ontology/pull/1237

shawntanzk:
Fixes #1319


shawntanzk:
note: I left the morphology out of the logical def as that is not technically what defines the neuron (sufficiently defined by what it innervates), happy to add it in though, don't think it makes a difference in this case

patrick-lloyd-ray (approved):
Looks good to me! Glad the morphology description is still in the text.

Issue #1577, opened 2022-05-10
[NTR] [ASCT+B table] proximal tubule epithelial cell (3 terms for segments 1 - 3)


bvarner-ebi:
This is part of the ASCT+B harmonisation initiative. New terms are being spawned from [the kidney table](https://docs.google.com/spreadsheets/d/1BihrOU9vWrsA_wYjcbdPUr42_Sb22CO3gkBDieiYQPs/edit#gid=1063693910).

Note - please check that the term does not already exist (check OLS: https://www.ebi.ac.uk/ols/ontologies/cl)

**Preferred term label:**
proximal tubule epithelial cell segment 1
proximal tubule epithelial cell segment 2
proximal tubule epithelial cell segment 3

**Synonyms:**
currently none

**Definition (free text, please give PubMed ID in format PMID:XXXXXX):**
respectively:
Epithelial cell that is part of segment 1 (S1) of the proximal tubule epithelium.
Epithelial cell that is part of segment 2 (S2) of the proximal tubule epithelium.
Epithelial cell that is part of segment 3 (S3) of the proximal tubule epithelium.
PMID:31249312

**Parent cell type term (check the hierarchy here https://www.ebi.ac.uk/ols/ontologies/cl):**
CL:0002306 'epithelial cell of proximal tubule'

**What is the anatomical structure that the cell is a part of? Please check Uberon: https://www.ebi.ac.uk/ols/ontologies/uberon**
UBERON:0004134 'proximal tubule'

**Your nano-attribution (ORCID)**
https://orcid.org/0000-0002-1773-2692
@emquardokus, kindly advise if other attributions should be added to these terms

**Any additional notes or concerns**
"parent GitHub ticket": #1482

bvarner-ebi:
The following terms have been added:

CL:4030009 'epithelial cell of proximal tubule segment 1'
CL:4030010 'epithelial cell of proximal tubule segment 2'
CL:4030011 'epithelial cell of proximal tubule segment 3'

PR #1578, opened 2022-05-10
Addresses #1577 [NTR] [ASCT+B table] proximal tubule epithelial cell …

bvarner-ebi:
Addresses #1577 [NTR] [ASCT+B table] proximal tubule epithelial cell (3 terms for segments 1 - 3)


src/ontology/cl-edit.owl
```diff
@@ -30074,6 +30077,30 @@ AnnotationAssertion(rdfs:label obo:CL_4030008 "pronephric podocyte")
 EquivalentClasses(obo:CL_4030008 ObjectIntersectionOf(obo:CL_0000653 ObjectSomeValuesFrom(obo:BFO_0000050 obo:UBERON_0002120)))
 SubClassOf(obo:CL_4030008 obo:CL_0000653)
 
+# Class: obo:CL_4030009 (epithelial cell of proximal tubule segment 1)
```
paolaroncaglia, line 30080: @bvarner-ebi 
I agree that it's hard to get these long labels 100% right. Personally I'd prefer 'epithelial cell of segment 1 of proximal tubule' for complete disambiguation, but it makes for a longer label. Another option is to leave 'epithelial cell of proximal tubule segment 1' but add an exact synonym "epithelial cell of segment 1 of proximal tubule". Same for the 2 sibling terms. Thanks.

src/ontology/cl-edit.owl
```diff
@@ -30074,6 +30077,30 @@ AnnotationAssertion(rdfs:label obo:CL_4030008 "pronephric podocyte")
...
+
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref "PMID:31249312"^^xsd:string) obo:IAO_0000115 obo:CL_4030009 "Epithelial cell that is part of segment 1 (S1) of the proximal tubule epithelium."^^xsd:string)
+AnnotationAssertion(<http://purl.org/dc/terms/contributor> obo:CL_4030009 <http://orcid.org/0000-0002-1773-2692>)
+AnnotationAssertion(<http://purl.org/dc/terms/date> obo:CL_4030009 "2022-05-10T13:14:00Z"^^xsd:dateTime)
+AnnotationAssertion(rdfs:label obo:CL_4030009 "epithelial cell of proximal tubule segment 1"^^xsd:string)
+SubClassOf(obo:CL_4030009 obo:CL_0002306)
```
paolaroncaglia, line 30086: @bvarner-ebi OLS is slow and I haven't had a chance at looking more closely in the ASCT+B table or recent Uberon commits: are new terms going to be created for proximal tubule segments, so that part-of axioms or equivalences can be added as appropriate?

src/ontology/cl-edit.owl
bvarner-ebi: Thank you, @paolaroncaglia!

There seems to be inconsistency among preexisting labels... some are structured

'% cell of %', 

whilst others are of the format 

'% cell'

Examples:
'epithelial cell of proximal tubule',
'brush border cell of the proximal tubule',
'kidney proximal convoluted tubule epithelial cell', 
'kidney inner medulla collecting duct epithelial cell'

I'll add synonyms as you suggest in the latter half of your comment.

src/ontology/cl-edit.owl
bvarner-ebi: At this moment I was not planning to add the more specific Uberon terms. I agree they would be valuable, but I don't think they would be strictly necessary. My thought process is that we go this first pass with getting the new terms in CL so they will be available for use and we can always to a second pass of refining if desired. Does that seem like a reasonable approach? CC: @emquardokus

src/ontology/cl-edit.owl
paolaroncaglia: @bvarner-ebi ok with me, thanks!

src/ontology/cl-edit.owl
paolaroncaglia: @bvarner-ebi seems reasonable to me, yes.

paolaroncaglia (approved):

PR #1603, opened 2022-05-24
Switch FBbt mappings to SSSOM

gouttegd:
Currently, mappings between CL and FBbt are managed the “classical way”, as cross-references stored on the CL side (though some of them also used to be stored on the FBbt side).

This PR switches from the “classical way” to FBbt-maintained mappings in the SSSOM format. To do that, it:

* removes all FBbt cross-references from the -edit file;
* imports a new dedicated `mappings.owl` component;
* updates the Makefile to automatically download the SSSOM file from FBbt and generate the `mappings.owl` component.

This is exactly the same approach as the one already in use in Uberon.

The initial plan was to have those mapping components managed directly by the ODK (to avoid duplication of code across repositories), but since we don’t have yet a roadmap in the ODK for that, I’d rather accept _some_ duplication for now instead of letting the CL/FBbt mappings go without updates.

### DIFF
--- a/src/ontology/Makefile
+++ b/src/ontology/Makefile
@@ -46,7 +46,7 @@ TODAY ?=                    $(shell date +%Y-%m-%d)
 OBODATE ?=                  $(shell date +'%d:%m:%Y %H:%M')
 VERSION=                    $(TODAY)
 ANNOTATE_ONTOLOGY_VERSION = annotate -V $(ONTBASE)/releases/$(VERSION)/$@ --annotation owl:versionInfo $(VERSION)
-OTHER_SRC =                 $(PATTERNDIR)/definitions.owl 
+OTHER_SRC =                 $(PATTERNDIR)/definitions.owl $(COMPONENTSDIR)/mappings.owl 
 ONTOLOGYTERMS =             $(TMPDIR)/ontologyterms.txt
 PATTERNDIR=                 ../patterns
 PATTERN_TESTER=              dosdp validate -i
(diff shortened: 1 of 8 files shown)


matentzn (approved):
Great. Not a greeeeaaaat fan of src/scripts/sssom2xrefs.awk over sssom-py, but I think its fine for now, while we are stabelising sssom-py convert.

gouttegd:
Not a great fan either (that’s akin to a Mungall-ism, but in a language even older than Perl :p ), but `sssom convert` does not satisfy the use case for now, does it? Unless I missed something, there is currently no way to customise the generated OWL to introduce the annotation we want (in this case, `oboInOwl:hasDbXref`) instead of the annotation as it appears in the mapping file (`skos:exactMatch`).

I certainly don’t want this AWK script to propagate outside of CL/Uberon, and I look forward to replace all of it with a proper solution implemented directly in the ODK. But until we can do that, this will do.

dosumis:
How does this relate to bridge file generation?

gouttegd:
@dosumis With this PR, CL will automatically pick up any changes that I do to the FBbt/CL mappings (that I manage on the FBbt side). Those mappings are converted to xrefs in the released version of CL, and after that the bridge generation process that is in place in the Uberon repository (based on Chris’ `make-bridge-ontologies-from-xrefs.pl` script) takes over as usual to create the bridge files from the xrefs.

So this PR does not actually change anything in the way the bridge files are generated (yet!). The goal here is to start introducing the use of SSSOM without breaking the bridge generation pipeline.

Later (dependent notably on updates to the SSSOM tooling), we hope to be able to skip the xrefs step altogether and generate the bridge files directly from the SSSOM (though we might still keep the xrefs for backward compatibility). But we are not there yet.

PR #1609, opened 2022-05-30
Fix nifext_55 prefix

### LINKED ISSUE #1607: Fix nifext_55 prefix
NIFEXT:55 expands to http://uri.neuinfo.org/nif/nifstd/nifext_55 and resolves to http://uri.interlex.org/base/ilx_0107375. The prefix rule is `@prefix NIFEXT: <http://uri.neuinfo.org/nif/nifstd/nifext_> .`.

_Originally posted by @tgbugs in https://github.com/obophenotype/uberon/issues/2354#issuecomment-1097164518_

anitacaron:
Fixes #1607

<s>Based on the same cross-reference prefix, fix this only case that is missing the prefix.</s>


src/ontology/cl-edit.owl
```diff
@@ -29538,7 +29538,7 @@ SubClassOf(obo:CL_4023075 ObjectSomeValuesFrom(obo:RO_0002292 obo:PR_P60041))
 AnnotationAssertion(Annotation(oboInOwl:hasDbXref "WikipediaVersioned:Martinotti_cell&oldid=9947732"^^xsd:string) obo:IAO_0000115 obo:CL_4023076 "An interneuron that has Martinotti morphology. These interneurons are scattered throughout various layers of the cerebral cortex, sending their axons up to the cortical layer I where they form axonal arborization.")
 AnnotationAssertion(<http://purl.org/dc/elements/1.1/creator> obo:CL_4023076 <http://orcid.org/0000-0001-7258-9596>)
 AnnotationAssertion(oboInOwl:hasDbXref obo:CL_4023076 "ILX:0107375")
-AnnotationAssertion(oboInOwl:hasDbXref obo:CL_4023076 "nifext_55")
+AnnotationAssertion(oboInOwl:hasDbXref obo:CL_4023076 "NIFSTD:nifext_55")
```
matentzn, line 29541: Why not `NIFSTD:55`?

anitacaron: Because there're others like this

```
AnnotationAssertion(oboInOwl:hasDbXref obo:CL_1001435 "NIFSTD:nifext_122"^^xsd:string)
```

matentzn: According to your ticket #1607 

```
@prefix NIFEXT: <http://uri.neuinfo.org/nif/nifstd/nifext_>
```

So your example expands to http://uri.neuinfo.org/nif/nifstd/nifext_nifext_55

Which is wrong. So maybe then, all NIFSTD curies are wrong!

anitacaron: They expand to nothing because they're all strings?

anitacaron: But if the prefix is `@prefix NIFSTD: <http://uri.neuinfo.org/nif/nifstd/>` would work.

matentzn: Yes, ignore that minor detail :D

matentzn: If you want an explanation ask again in Tech call

matentzn:
@cthoyt whats better throughout: `NIFSTD:nifext_55` or `NIFEXT:55`?

@anitacaron sorry I misread your prefix in the ticket, I thought it was NIFSTD

src/ontology/cl-edit.owl
matentzn: Yesh, I agree @anitacaron NIFSTD would work lets wait for @cthoyt to get his opinion.

cthoyt: neither of these are in bioregistry, therefore you can assume I don't have an opinion :p 

but as far as NIF's sprawl of namespaces, I think this is because nobody really knows what any of these do. Can you spot check that these references go to things that make sense if you use the NIFSTD or NIFEXT URI format strings to expand to URIs?

cthoyt:
I would highly prefer `NIFEXT:55` over `NIFSTD:nifext_55` since it doesn't embed one namespace in another

src/ontology/cl-edit.owl
cthoyt: Would one of you be willing to submit a bioregistry new prefix request then?

cthoyt: oh boy ☹️

anitacaron: Just to clarify and resolve this comment, we agree to use the prefix NIFEXT that is going to be added in the bioregistry via https://github.com/biopragmatics/bioregistry/issues/404.

matentzn: Lets wait a moment on @tgbugs to reply to that issue.

matentzn (approved):

cthoyt:
Thanks for doing this @matentzn and @anitacaron! If you're still curious how NIFSTD is being (often mis-)used, you can check https://cthoyt.com/oquat/unknowns/prefix/nifstd

Issue #1604, opened 2022-05-24
hepatic progenitor cell 


masciam:
Note - please check that the term does not already exist (check OLS: https://www.ebi.ac.uk/ols/ontologies/cl)

**Preferred term label:**
hepatic progenitor cell 

**Synonyms**
adult hepatic progenitor cell

**Definition (free text, please give PubMed ID in format PMID:XXXXXX)**
It is a progenitor cell that is part of the adult human liver and is capable of differentiating only in hepatocyte or cholangiocyte 

**Parent cell type term (check the hierarchy here https://www.ebi.ac.uk/ols/ontologies/cl):**
progenitor cell

**What is the anatomical structure that the cell is a part of? Please check Uberon: https://www.ebi.ac.uk/ols/ontologies/uberon**
adult liver

**Your nano-attribution (ORCID)**
0000-0003-1940-6740

**Any additional notes or concerns**

bvarner-ebi:
Thank you for this ticket, @masciam. Would you kindly add the reference (PMID, DOI, etc) to support adding this cell type?

emquardokus:
@bvarner-ebi @masciam 
Here are several DOI to support adding cell type; needed urgently for HuBMAP
https://doi.org/10.1038/s12276-020-0483-0
https://doi.org/10.7554/eLife.47313
https://doi.org/10.1016/B978-0-12-385233-5.00008-8
http://dx.doi.org/10.1136/gut.2007.143321
Thanks

bvarner-ebi:
This cell type appears to be the same as CL:0002196 'hepatic oval stem cell'.

In https://doi.org/10.1038/s12276-020-0483-0 it states: "... [biliary epithelial cells] are able to dedifferentiate into liver progenitor cells (LPCs), also known as oval cells.'

In [PMID:17901986](https://pubmed.ncbi.nlm.nih.gov/17901986) it states: "Hepatic stem/progenitor cells also known as “oval cells” in rodents have been implicated in liver tissue repair, ..."

The [comment here](https://github.com/obophenotype/cell-ontology/issues/1594#issue-1241737084) proposes adding this term as a synonym to 'hepatic oval stem cell'. There is a concern regarding the multipotent categorisation of stem cell, but based on the literature and in line with the definition of 'hepatic oval stem cell', this progenitor cell can differentiate into more than one cell.

Proposal:
[...]

PR #1625, opened 2022-06-08
Addresses #1604 hepatic progenitor cell

bvarner-ebi:
Addresses #1604 hepatic progenitor cell


shawntanzk:
scanning through the linked ticket, there seems to be a term that is used in humans and one used in rats - in which case I would add a comment on it and add taxon restrictions (mind you I just scanned through, so if this is wrong ignore me).

emquardokus:
You're spot on Shawn. Rat terminology is oval cell based on the morphology of nucleus; human is using adult human liver progenitor cell (there could also be embryonic, so specification is important).
Upon reading literature further, there also seems to be subsets associated with injury induced vs normal population without injury.

bvarner-ebi:
Comment added.

shawntanzk:
I have a bit of concert over the removal of 
[hepatic oval stem cell](http://purl.obolibrary.org/obo/CL_0002196) SubClassOf [hepatic stem cell](http://purl.obolibrary.org/obo/CL_0002195)
I don't disagree that it shouldn't be a subclass, but would it be right to say that it develops from? (eg is it a progenitor cell that develops from a hepatic stem cell after liver injury?) If so, that axiom should be added and it should be stated in the def - basically, I think it hepatic oval stem cell should still be linked to hepatic stem cell given the name. It would also be useful to add a curators/editors note on your decision, and maybe link the github ticket to it so future curators/editors will be able to understand why it hepatic oval stem cell isn't a subclass of hepatic stem cell
with this PR, hepatic oval stem cell will also lose its relation to liver which I think is problematic (I assume hepatic in the name means it has to be part of the liver?) - perhaps adding back the axiom would be a good idea
Hope that helps :)

bvarner-ebi:
> I don't disagree that it shouldn't be a subclass [of [hepatic stem cell](http://purl.obolibrary.org/obo/CL_0002195)], but would it be right to say that it develops from? (eg is it a progenitor cell that develops from a hepatic stem cell after liver injury?) If so, that axiom should be added and it should be stated in the def 

@shawntanzk, do you have a source to support this?

In PMID:12468240 : "Bone marrow stem cells have recently been shown to be a potential source of the hepatic oval cells..."
In PMID:12490302 : "Although bone marrow stem cells can generate oval cells and hepatocytes, transdifferentiation is very rare and inefficient."

If there is a source that has evidence of the step of 'hepatic stem cell' --> 'hepatic oval stem cell', the suggestion seems reasonable. The sources above suggest bone marrow stem cell --> hepatic oval (stem) cell. It is not clear to me if there is an intermediate step involving the 'hepatic stem cell'. A transition is noted [here](https://github.com/obophenotype/cell-ontology/issues/1604#issuecomment-1148459107), but I would like to have a source before adding the axiom you suggest.

emquardokus:
@bvarner-ebi @shawntanzk 
Please see this article--they discuss differences between hepatic stem cells and oval cells. Also older literature suggests the possibility for bone marrow as a source for hepatic stem cells, but newer literature seems to disprove that previous hypothesis and provides evidence for adult and fetal livers being sources for hepatic stem cells. This paper discusses the molecular markers that differentiate between all of these populations, so this is super useful. 
Bruno, Stefania, Maria Beatriz Herrera Sanchez, Giulia Chiabotto, Valentina Fonsato, Victor Navarro-Tableros, Chiara Pasquino, Marta Tapparo, and Giovanni Camussi. 2021. “Human Liver Stem Cells: A Liver-Derived Mesenchymal Stromal Cell-Like Population With Pro-Regenerative Properties.” Frontiers in Cell and Developmental Biology 9. https://doi.org/10.3389/fcell.2021.644088.

shawntanzk:
> @shawntanzk, do you have a source to support this?

lol no, hence it being a question, I know next to nothing about liver stuff, its been more than 10 years since I've studied anything not in the brain hahahaha, it was just my assumption from the names. 
Thanks so much for the reference @emquardokus, happy to read up more next week when im back at work (promised myself no heavy work while im on leave heh).

shawntanzk:
Ok read a little bit into this, and I think I understand what you mean @emquardokus with this from the paper you shared: 

"Although there are differences in the phenotypic markers between the HLSCs and oval cells, we can only speculate that HLSCs are a separate niche of stem cells in the liver, since it has not been confirmed in vivo." 

Therefore in the ontology we should represent them as separate cell types that aren't connected. 

I think to overcome this issue perhaps looking at this paper/figure might help:

https://www.sciencedirect.com/science/article/pii/S0925477302003386

1-s2 0-S0925477302003386-gr1

Perhaps then just adding these axioms would help (honestly its the part of some liver that I think is important): 
"develops from" some hepatoblast 
"part of" some liver

I don't know how liver stem cell links to the above but I think somehow getting hepatoblast, hepatocytes, hepatic liver stem cell, & hepatic oval cell all linked up properly would be nice, but I get that might take awhile 

adding the canal of hering stuff (I think its http://purl.obolibrary.org/obo/UBERON_0001282) would also be nice, but I guess also time issue

Basically given this is high priority, I think the minimal is to link it to the liver and maybe hepatoblast, my 2c anyway :D Hope that helps

bvarner-ebi:
> Therefore in the ontology we should represent them as separate cell types that aren't connected.

Which is what I'm proposing with this PR, correct?

shawntanzk:
> Which is what I'm proposing with this PR, correct?

yes,  not connected might be a bad way to say it - I think your modelling is right, but it currently lacks axioms and explanation and a comment on why you modelled it that way (as a naive person seeing the terms separate I would think its a mistake rather than intentional)

bvarner-ebi:
Additional text added to comment based on @emquardokus's [comment above](https://github.com/obophenotype/cell-ontology/pull/1625#issuecomment-1152229688). Ellen, I'll add your ORCiD to the comment if you like (kindly confirm what it is).

@shawntanzk, I was thinking of adding:

'develops into' some 'hepatocyte'
'develops into' some 'cholangiocyte'

but would the above suggest that every 'hepatic progenitor cell' must develop into both of these cells (which I do not believe is possible - a HPC cannot develop into both of the other cells) or does it mean that every 'hepatic progenitor cell' **may** develop into either one or neither (which would seem correct).

If you have other axioms in mind directly related to this cell type, kindly advise. Thank you for the suggestions above, but I believe those additions should be part of a separate ticket.

bvarner-ebi:
> Personally, I think adding these 2 axioms on hepatic oval stem cell in this PR is ideal:
> 
> ```
> "develops from" some hepatoblast
> "part of" some liver
> ```

Thanks, @shawntanzk! We discussed this during the CL call. I added `"part of" some liver`, but myself and @emquardokus have reservations on "develops from" some hepatoblast given lack of clarity about this in recent literature. There is obviously more work to do on remodelling these cell types in the liver and I propose it makes sense to put that in future tickets...

bvarner-ebi:
@shawntanzk, any idea why the checks are not running? Is it because there is a pending release?

bvarner-ebi:
> not sure, but maybe try fixing the clash/updating from master and see if that triggers the actions?

@shawntanzk, that seemed to fix it. Thanks!

shawntanzk:
@bvarner-ebi could you remove this
[hepatic oval stem cell](http://purl.obolibrary.org/obo/CL_0002196) [hasOBONamespace](http://www.geneontology.org/formats/oboInOwl#hasOBONamespace) "cell"
This was recently scrubbed from CL - I think cause you were using an old branch when this PR started, it is trying to add it back. 
Thanks

PR #1626, opened 2022-06-08
Add `gamete-nursing cell` and subclasses.

### LINKED ISSUE #1611: nurse cell

**Preferred term label:**

**Synonyms**

Cells accessory to egg and/or sperm formation in a wide variety of organisms. Usually thought to synthesize special substances and to export these to the developing gamete (BTO:0000953)

**Your nano-attribution (ORCID)**:0000-0003-2534-198X

gouttegd:
This PR adds a new term `gamete-nursing cell` to refer to any cell that provides cytoplasmic material to a developing gamete. Since such cells can have, depending on the species, either a somatic or a germinal origin, this term makes no assumption about the lineage. Instead, we also add two subclasses to specifically refer to the germline-derived and somatic nursing cells.

closes #1611


rays22 (approved):
Looks good.

src/ontology/cl-edit.owl
```diff
@@ -30066,6 +30069,38 @@ AnnotationAssertion(rdfs:label obo:CL_4028006 "alveolar type 2 fibroblast cell")
 SubClassOf(obo:CL_4028006 obo:CL_0002241)
 SubClassOf(obo:CL_4028006 ObjectSomeValuesFrom(obo:BFO_0000050 obo:UBERON_0002299))
 
+# Class: obo:CL_4029001 (gamete-nursing cell)
+
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref "doi:10.1016/bs.ctdb.2020.10.008"^^xsd:string) obo:IAO_0000115 obo:CL_4029001 "A cell that supports the development of a gamete by providing it cytoplasmic material (including entire organelles) by direct cross-membrane channels (del Pino, 2021).")
```
bvarner-ebi, line 30074: @gouttegd, is it standard to include citation in the definition as opposed to only in a dbxref annotation?

gouttegd: It is in the Drosophila Anatomy Ontology, so I just carry over that practice to Uberon and CL.

There doesn’t seem to be a similar policy in Uberon/CL. I see a few terms where it is done, though.

matentzn: I think it is much better to add a doi to the definition and _not_ to add xrefs to PMID. This is not what xrefs are for! If we want to discuss the practice, feel free to voice your opinions here: https://github.com/information-artifact-ontology/ontology-metadata/issues/94

dosumis: It's what they're for for now - until some process officially changes it.  But the downstream churn when that happens will be horrible.

dosumis: The point of textual minirefs inside the definitions is just to follow normal scientific convention for cases where a definition makes multiple assertions sourced from different papers.  These are *In addition* to links via an AP.  I'd be happy to add as best practice advice for CL.

matentzn: Sorry, I think I have not clearly said what I meant. I think it is much better to add a doi to the definition using the xref property and not to add xrefs to PMID _directly on the term_. This is not what xrefs are for! If we want to discuss the practice, feel free to voice your opinions here: https://github.com/information-artifact-ontology/ontology-metadata/issues/94

dosumis: > I think it is much better to add a doi to the definition using the xref property and not to add xrefs to PMID directly on the term.

* Agree links to pub should be on def or synonym, not term.   
* We have to allow for heterogeneous identifiers including PMID and DOI.  Not every historic paper has a DOI, but I think PUBMED did a pretty good job of historically cataloging.

bvarner-ebi: Thank you for the feedback.

Yes, to clarify, I was questioning the practice of including a citation in a text definition. For example, I would be inclined to **not** include "(del Pino, 2021)" in the text definition, especially if there is a dbxref annotation on the text definition. It seems redundant, although I understand if can be useful if the definition is expanded.

A cell that supports the development of a gamete by providing it cytoplasmic material (including entire organelles) by direct cross-membrane channels **(del Pino, 2021)**.

gouttegd: @matentzn So just to be clear, you’re saying you want

```
AnnotationAssertion(Annotation(oboInOwl:hasDbXref "PMID:0123456789") obo:IAO_0000115 obo:CL_9999999 "The term definition.")
```

(where `CL_9999999` is the term being defined)

instead of

```
AnnotationAssertion(obo:IAO_0000115 obo:CL_9999999 "The term definition.")
AnnotationAssertion(oboInOwl:hasDbXref obo:CL_9999999 "PMID:0123456789")
```

(or, in OBO Flatfile syntax, you want

```
id: CL:9999999
def: The term definition. [PMID:0123456789]
```

instead of

```
id: CL:9999999
def: The term definition.
xref: PMID:0123456789
```
)

?

If so, that’s what I‘ve done in this PR, so I don’t understand why this needs discussing here.

matentzn: Thats right, it does not need a discussion, all correct. 

I have no opinion on adding the Author citation into the definition. I don't think the redundancy hurts, and I love attribution. So I have a tiny inclination to adding too much attribution :) Can be ignored.

PR #1628, opened 2022-06-09
Remove `oboInOwl:hasOBONamespace`

anitacaron:
Fixes #531

Issue #1352, opened 2021-12-06
Unify all attribution to ORCID REF: pato#278


anitacaron:
Duplicating issue for tracking purposes.

REF: https://github.com/pato-ontology/pato/issues/278

PR #1630, opened 2022-06-15
CL 15 June release

matentzn:
@shawntanzk review with care. Many changes on first glance.


shawntanzk:
general removal of namespace:cell (in .obo format) or in owl: `<oboInOwl:hasOBONamespace rdf:datatype="http://www.w3.org/2001/XMLSchema#string">cell</oboInOwl:hasOBONamespace>`|
I assume this is intentional?

This is wrong, not sure why it didnt get picked up by qc before making its way here.  (It's my stuff so I will fix it next release >.<)
Screenshot 2022-06-15 at 21 21 57
This is the term
[Term]
id: CL:0000695
name: Cajal-Retzius cell
But I wouldn't let this stop a release - can be fixed later. 
Same thing in 
[Term]
id: CL:0002453
name: oligodendrocyte precursor cell 

(I think there are a few in there, not sure where this change came about)
In general there is a whole lot of use of dce in new terms -> so I'm guessing CL hasn't enforced the check yet.
Solution would be mass fix + qc check I would think

anitacaron:
> general removal of namespace:cell (in .obo format) or in owl: <oboInOwl:hasOBONamespace rdf:datatype="http://www.w3.org/2001/XMLSchema#string">cell</oboInOwl:hasOBONamespace>|
> I assume this is intentional?

Yes, I removed in [this PR](https://github.com/obophenotype/cell-ontology/pull/1628)

> In general there is a whole lot of use of dce in new terms

Maybe this is because of the default property in Protege?
No need to create another issue for this. I'm already working to fix [this one](https://github.com/obophenotype/cell-ontology/issues/1352)

Issue #1627, opened 2022-06-09
Create new GitHub issue template for new term requests coming in via cellxgene


bvarner-ebi:
As discussed in recent cellxgene meeting, a template for new CL term requests coming in via cellxgene will be created. 

This template should contain the standard information for a CL new term request plus additional fields specific to cellxgene (@dosumis, can you confirm fields?) and a GitHub label for 'cellxgene'.

dosumis:
I think minimal would be to just add the cellXGene label - maybe should also assign to you by default.  We could also add a bit about suggesting markers that uniquely identify the cell type (noting species).

PR #1633, opened 2022-06-16
Addresses #1627 Update a_adding_term_cellxgene.md

bvarner-ebi:
Template was not appearing in GitHub. It may have to do with the quotes I used in the label values. [h_typos_bugs.md](https://github.com/obophenotype/cell-ontology/files/8917596/h_typos_bugs.md) is a template that uses multiple labels with no quotes, so I updated a_adding_term_cellxgene.md to match the format.


bvarner-ebi:
Sorry, @dosumis, but can you approve this again. I think the format of using two sets of single quotes in the labels were blocking the template from being recognised.

bvarner-ebi:
@dosumis, can you approve this again? I think the format of using two sets of single quotes in the labels were blocking the template from being recognised.

Issue #1582, opened 2022-05-11
[NTR] [ASCT+B table] epithelial cell of distal convoluted tubule (2 terms for early and late segments) 


bvarner-ebi:
This is part of the ASCT+B harmonisation initiative. New terms are being spawned from [the kidney table](https://docs.google.com/spreadsheets/d/1BihrOU9vWrsA_wYjcbdPUr42_Sb22CO3gkBDieiYQPs/edit#gid=1063693910).

Note - please check that the term does not already exist (check OLS: https://www.ebi.ac.uk/ols/ontologies/cl)

**Preferred term label:**
epithelial cell of early distal convoluted tubule
epithelial cell of late distal convoluted tubule

**Synonyms:**
respectively:
DCT1 cell
DCT2 cell

**Definition (free text, please give PubMed ID in format PMID:XXXXXX):**
respectively:
Epithelial cell located in the early distal convoluted tubule.
Epithelial cell located in the late distal convoluted tubule.
PMID: 24855283
*note, I removed text regarding biomarkers from the definitions.

**Parent cell type term (check the hierarchy here https://www.ebi.ac.uk/ols/ontologies/cl):**
CL:1000849 ‘kidney distal convoluted tubule epithelial cell’

**What is the anatomical structure that the cell is a part of? Please check Uberon: https://www.ebi.ac.uk/ols/ontologies/uberon**
UBERON:0001292 ‘distal convoluted tubule’

**Your nano-attribution (ORCID)**
https://orcid.org/0000-0002-1773-2692
@emquardokus, kindly advise if other attributions should be added to these terms

**Any additional notes or concerns**
"parent GitHub ticket": #1482

bvarner-ebi:
FYI, @emquardokus,
The following terms have been added to CL:

CL:4030016 'epithelial cell of early distal convoluted tubule'
CL:4030017 'epithelial cell of late distal convoluted tubule'

PR #1644, opened 2022-06-21
Addresses #1582-NTR-epithelial_cells_DCT

bvarner-ebi:
Addresses #1582-NTR-epithelial_cells_DCT


shawntanzk:
Just to double - check early and late are segments and not time right? lol 
I think if that is the case, it might be worth adding it in uberon and doing that part of accordingly given this cell type is defined by its locality. Also initially when looking at the definition and terms, I thought it was to do with time (eg early stage of life, late stage of life) (might also come from me coming from a developmental psychology lab, then a lab that did Alzheimers related stuff >.< ) Either way some clarity might be useful - an Uberon term with the appropriate definition would do the trick, but if you think its not useful to have in uberon, then perhaps some clarify in the defs

bvarner-ebi:
> I think if that is the case, it might be worth adding it in uberon and doing that part of accordingly given this cell type is defined by its locality.

Thank you for the suggestion, @shawntanzk. I found that the UBERON terms already existed, but would need to be imported to CL.
This was my first attempt at an import, so I will especially appreciate you review. I see many files have changed in the process and I will have to resolve some conflicts. 
FYI, @matentzn

src/ontology/components/mappings.owl
```diff
@@ -142,6 +142,9 @@
   <Class rdf:about="http://purl.obolibrary.org/obo/CL_0000381">
     <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00005130</oboInOwl:hasDbXref>
   </Class>
+  <Class rdf:about="http://purl.obolibrary.org/obo/CL_0000850">
+    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00005133</oboInOwl:hasDbXref>
```
shawntanzk, line 146: @gouttegd - could you help make sure this is correct - I think it got updated with import updates

shawntanzk:
> This was my first attempt at an import, so I will especially appreciate you review. I see many files have changed in the process and I will have to resolve some conflicts.

It all looks good - the file changes I'm not sure about but its not anything on you, I've asked @gouttegd to look at them to make sure they are right as its fly stuff. But basically you did everything right with imports :)

bvarner-ebi:
@shawntanzk, I resolved 9 conflicts in the merged_import.owl file, but most were arbitrary decisions. I would not know another way to approach it and do not know why so many conflicts appeared. If we there is a way to test that I did not make any breaking changes, pls advise. 

FYI, @matentzn.

matentzn (approved):
All looks great - I am a bit puzzled but they fact that the components are refreshed although you just refresh the imports. What exact command did you run in the end?

shawntanzk:
@bvarner-ebi - I think you can make equivalent classes and in fact I think you should follow exactly the the textual definition states and let the reasoner do the work (of course making sure the reasoner works, if not figuring out why). so: 

epithelial cell of early distal convoluted tubule
def: Epithelial cell located in the early distal convoluted tubule. 
Equivalent class: epithelial cell and ("part of" some "early distal convoluted tubule")

This should automatically place it under "kidney distal convoluted tubule epithelial cell"

PS - sorry for being annoying and not bringing this up earlier >.< just realised.

matentzn:
@bvarner-ebi there should _never_ be any conflicts on imports. The only way this can happen is if you opened a new branch, and someone else came to update the import before you, and then you didn't update your branch from master first. Generally:

1. Always update from master before running an import
2. Then run the import

I suggest you rerun: `sh run.sh make no-mirror-refresh-merged` just to be sure (after updating the branch from master)

bvarner-ebi:
> All looks great - I am a bit puzzled but they fact that the components are refreshed although you just refresh the imports. What exact command did you run in the end?

`sh run.sh make PAT=false imports/go_import.owl -B` (I now know this was not correct)

`sh run.sh make IMP=true MIR=true PAT=false imports/go_import.owl -B` (I now know this was not correct)

`sh run.sh make imports/merged_import.owl`

matentzn:
OK in the end it makes sense! All good. But I would still recommend rerunning the command i suggest above - never resolve conflicts on imports, always rerun!

src/ontology/components/mappings.owl
gouttegd: That’s correct, yes. It’s a change that I made recently in the FBbt-to-CL mappings. That it got picked up during an import update is the expected behaviour, I believe.

gouttegd (approved):
All changes regarding the FBbt-CL mappings are normal and expected.

gouttegd:
> I am a bit puzzled but they fact that the components are refreshed although you just refresh the imports.

Unless I’m missing something, I believe this is normal.

Components are listed in `OTHER_SRC`:

```Makefile
OTHER_SRC = $(PATTERNDIR)/definitions.owl $(COMPONENTSDIR)/hra_subset.owl $(COMPONENTSDIR)/mappings.owl
```

`OTHER_SRC` is a prerequisite for `SRC_MERGED`, which itself is a prerequisite for `PRESEED` and ultimately for the `$(IMPORTDIR)/merged_terms_combined.txt`, which is updated when imports are refreshed.

bvarner-ebi:
> OK in the end it makes sense! All good. But I would still recommend rerunning the command i suggest above - never resolve conflicts on imports, always rerun!

Steps I'm now doing:

`git checkout 1582-NTR-epithelial-cells-of-distal-convoluted-tubule`
`git merge master`
"Already up to date."

`sh run.sh make no-mirror-refresh-merged`
`sh run.sh make imports/merged_import.owl`

src/ontology/cl-edit.owl
```diff
@@ -28080,6 +28084,27 @@ AnnotationAssertion(rdfs:label obo:CL_4030015 "collecting duct intercalated cell
...
 
+# Class: obo:CL_4030016 (epithelial cell of early distal convoluted tubule)
+
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref <http://orcid.org/0000-0003-2804-127X>) Annotation(oboInOwl:hasDbXref "PMID:24855283"^^xsd:string) obo:IAO_0000115 obo:CL_4030016 "Epithelial cell located in the early distal convoluted tubule."^^xsd:string)
+AnnotationAssertion(<http://purl.org/dc/terms/contributor> obo:CL_4030016 <https://orcid.org/0000-0002-1773-2692>)
+AnnotationAssertion(<http://purl.org/dc/terms/date> obo:CL_4030016 "2022-06-21T12:47:48Z"^^xsd:dateTime)
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref "PMID:24855283"^^xsd:string) oboInOwl:hasBroadSynonym obo:CL_4030016 "DCT1 cell"^^xsd:string)
+AnnotationAssertion(rdfs:label obo:CL_4030016 "epithelial cell of early distal convoluted tubule")
+EquivalentClasses(obo:CL_4030016 ObjectIntersectionOf(obo:CL_0000066 ObjectSomeValuesFrom(obo:BFO_0000050 obo:UBERON_0005101)))
+SubClassOf(obo:CL_4030016 ObjectSomeValuesFrom(obo:BFO_0000050 obo:UBERON_0005101))
```
shawntanzk, line 28095: The subclass axioms can be removed as they are in the equivalent class already

src/ontology/cl-edit.owl
```diff
@@ -28080,6 +28084,27 @@ AnnotationAssertion(rdfs:label obo:CL_4030015 "collecting duct intercalated cell
 SubClassOf(obo:CL_4030015 obo:CL_0005011)
 SubClassOf(obo:CL_4030015 obo:CL_1001432)
 
+# Class: obo:CL_4030016 (epithelial cell of early distal convoluted tubule)
+
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref <http://orcid.org/0000-0003-2804-127X>) Annotation(oboInOwl:hasDbXref "PMID:24855283"^^xsd:string) obo:IAO_0000115 obo:CL_4030016 "Epithelial cell located in the early distal convoluted tubule."^^xsd:string)
```
shawntanzk, line 28089: Minor detail, but ideally "An epithelial cell located..."

shawntanzk (changes requested):
see inline (just last bit of tidying up :))

matentzn:
> sh run.sh make no-mirror-refresh-merged
> sh run.sh make imports/merged_import.owl

No need: `sh run.sh make no-mirror-refresh-merged` is all you needed to do!

bvarner-ebi:
> > sh run.sh make no-mirror-refresh-merged
> > sh run.sh make imports/merged_import.owl
> 
> No need: `sh run.sh make no-mirror-refresh-merged` is all you needed to do!

OK. I was following the advice from Slack: "CL uses merged import so you want to run sh run.sh make imports/merged_import.owl"

I think this page could use a "refreshing" itself: https://oboacademy.github.io/obook/howto/update-import/#Refresh-imports

matentzn:
@bvarner-ebi I think in the general case, the instructions on obook are fine for the general case - `sh run.sh make imports/merged_import.owl` just downloads all the mirrors which seems wasteful (the no-mirror command skips downloading ontologies).

bvarner-ebi:
> @bvarner-ebi I think in the general case, the instructions on obook are fine for the general case - `sh run.sh make imports/merged_import.owl` just downloads all the mirrors which seems wasteful (the no-mirror command skips downloading ontologies).

Thanks for explaining. 
If there is a page explaining what mirrors are and how to check if they are updated, that would be helpful to link to the instructions.
I read "if your mirrors are updated... ", but did not know how to check if they were.

shawntanzk (approved):

Issue #1585, opened 2022-05-11
[NTR] [ASCT+B table] renal medullary fibroblast


bvarner-ebi:
This is part of the ASCT+B harmonisation initiative. New terms are being spawned from [the kidney table](https://docs.google.com/spreadsheets/d/1BihrOU9vWrsA_wYjcbdPUr42_Sb22CO3gkBDieiYQPs/edit#gid=1063693910).

Note - please check that the term does not already exist (check OLS: https://www.ebi.ac.uk/ols/ontologies/cl)

**Preferred term label:**
renal medullary fibroblast

**Synonyms**
kidney medullary fibroblast (exact)

**Definition (free text, please give PubMed ID in format PMID:XXXXXX)**
Fibroblast located in the medullary interstitium of the kidney. doi: https://doi.org/10.1101/2021.07.28.454201

**Parent cell type term (check the hierarchy here https://www.ebi.ac.uk/ols/ontologies/cl):**
CL:1000692 'kidney interstitial fibroblast'

**What is the anatomical structure that the cell is a part of? Please check Uberon: https://www.ebi.ac.uk/ols/ontologies/uberon**
UBERON:0005211 'renal medulla interstitium'

**Your nano-attribution (ORCID)**
https://orcid.org/0000-0002-1773-2692
@emquardokus, kindly advise if other attributions should be added to these terms

**Any additional notes or concerns**
"parent GitHub ticket": #1482

bvarner-ebi:
The following cell types have been added to CL:

CL_4030022 ‘renal medullary fibroblast’
CL_4030025 ‘renal cortical fibroblast’

PR #1650, opened 2022-07-05
Adresses #1585 NTR renal medullary fibroblast

bvarner-ebi:
Adresses #1585 NTR renal medullary fibroblast

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -2685,6 +2685,7 @@ Declaration(Class(obo:CL_4030018))
 Declaration(Class(obo:CL_4030019))
 Declaration(Class(obo:CL_4030020))
 Declaration(Class(obo:CL_4030021))
+Declaration(Class(obo:CL_4030022))
 Declaration(Class(obo:CL_4030023))
 Declaration(Class(obo:CL_4030024))
 Declaration(Class(obo:CP_0000000))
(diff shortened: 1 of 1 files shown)


emquardokus:
@shawntanzk I have a question for you. If there is more than simply location (kidney interaction) that differentiates this type of fibroblast, would your modeling still hold up? Do we always ignore molecular markers? (Immune cells relies on molecular markers for differentiation of T and B cell types).

emquardokus:
@bvarner-ebi yes, I understand; I think one think that is difficult is if they exist in a data table, but may not be explicitly stated in words.  I'm just feeling all of this out in terms of thoughts on how to deal with questions I also receive.  I think what Shawn stated "(one would assume the term renal medullary fibroblast would represent all fibroblasts in the renal medulla)," Please continue :-) and Thank-you!

bvarner-ebi:
> I think one think that is difficult is if they exist in a data table, but may not be explicitly stated in words. I

@emquardokus, yes, I've started looking at tables more closely after you found and shared some on previous tickets :) 
There are cases where some genes/markers appeared in tables, but other cell types also expressed them, but at reduced levels. In those cases I'm hesitant to classify them as defining biomarkers.
I think adding them in comments is reasonable in those cases.

bvarner-ebi:
As always, thanks for the review and suggestions, @shawntanzk.
 
> 1. is there other types of kidney interstitial fibroblasts? If so what are they and should they be added?

I added 'renal cortical fibroblast' as well. 

> 2. could we edit kidney interstitial fibroblast to have an equiv class, that way we should be able to better model the requested term as just `EquivalentTo fibroblast and (part_of some "renal medulla interstitium")`
>    My preference would be to have kidney interstitial fibroblast be `EquivalentTo fibroblast and ('part of' some 'kidney interstitium')`

I implemented your suggestions.

dosumis:
> can I just add how beautiful it is to see the difference between the reasoned and unreasoned diff :D

That warms my heart.

Issue #1651, opened 2022-07-08
[NTR] stromal cell of thymus, interstitial cell of thymus [HCA]


bvarner-ebi:
Original ticket: https://github.com/HumanCellAtlas/ontology/issues/114

**Preferred term label**
stromal cell of thymus

**Synonyms** (add reference(s), please)
thymic stromal cell (exact)
https://www.nature.com/articles/s41467-020-20082-7

**Definition** (free text, with reference(s), please. PubMed ID format is PMID:XXXXXX)
A stomal cell of the thymus.
https://www.nature.com/articles/s41467-020-20082-7

**Parent cell type term** (check the hierarchy here https://www.ebi.ac.uk/ols/ontologies/cl)
CL:0000499 'stromal cell'

**What is the anatomical structure that the cell is a part of?** (check Uberon: https://www.ebi.ac.uk/ols/ontologies/uberon)
UBERON:0002370 thymus

**Preferred term label**
interstitial cell of thymus

**Synonyms** (add reference(s), please)
thymic interstitial cell (exact)
https://www.nature.com/articles/s41467-020-20082-7

**Definition** (free text, with reference(s), please. PubMed ID format is PMID:XXXXXX)
A cell that makes up the loose connective tissue of the thymus.
https://www.nature.com/articles/s41467-020-20082-7

**Parent cell type term** (check the hierarchy here https://www.ebi.ac.uk/ols/ontologies/cl)
_Term 1 above_ 'stromal cell of thymus'

**What is the anatomical structure that the cell is a part of?** (check Uberon: https://www.ebi.ac.uk/ols/ontologies/uberon)
UBERON:0002370 thymus

bvarner-ebi:
Added:

CL:4030001 ‘stromal cell of thymus’
CL:4030003 ‘interstitial cell of thymus’

PR #1652, opened 2022-07-08
Addresses #1651 NTR thymus cells

bvarner-ebi:
Addresses #1651 NTR thymus cells

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -2668,7 +2668,9 @@ Declaration(Class(obo:CL_4029001))
 Declaration(Class(obo:CL_4029002))
 Declaration(Class(obo:CL_4029003))
 Declaration(Class(obo:CL_4030000))
+Declaration(Class(obo:CL_4030001))
 Declaration(Class(obo:CL_4030002))
+Declaration(Class(obo:CL_4030003))
 Declaration(Class(obo:CL_4030004))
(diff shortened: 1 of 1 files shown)


bvarner-ebi:
> are there other types of thymic stromal cell?
> totally understand if these need to be added and the others can wait kindda situation, just having single child parents is something to avoid (see https://douroucouli.wordpress.com/2022/01/04/ontotip-avoid-the-single-child-anti-pattern/)

@shawntanzk, https://www.nature.com/articles/s41467-021-21346-6:
"While recent evidence suggests that the thymic stroma is comprised of more functionally distinct subpopulations than previously appreciated, the extent of this cellular heterogeneity in the human thymus is not well understood.... The predominant stromal cells found in the postnatal thymus are thymic epithelial cells (TECs), mesenchyme, endothelium, and non-lymphoid hematopoietic cells (dendritic cells and macrophages)".

@ami-day, do you have any other thymic stromal cells you would like to add now?

bvarner-ebi:
@shawntanzk, the requester's response:

> Based on the articles you found, it looks like there are multiple thymic stromal cell types. Given this, and the aim of the HCA to identify previously unknown cell types, I think it makes sense to create the "stromal cell of thymus" node as a parent to "interstitial cell of thymus". I do not currently have any other thymic stromal cells in mind to add.

PR #1694, opened 2022-09-05
remove "not" in CL

### LINKED ISSUE #1693: find all "Not" in CL
grep "ObjectComplementOf" in cl-edit and find what they are 
Not should not be allowed in CL -> ELK doesn't deal with this

shawntanzk:
Fixes #1693
Changed taxon restrictions to annotation never_in_taxon - @matentzn not sure if CL pipelines deal with expansion of this or is that uberon specific? Removed some equiv classes
PS sorry for asking so many reviewers - thought it might be something that everyone on the reviewer list might want to go through :) 

PS I love how the automated diff in simple file vs unreasoned show how the "not" is not being used by a reasoner :D

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -3939,9 +3939,9 @@ SubClassOf(obo:CL_0000085 ObjectSomeValuesFrom(obo:RO_0002160 obo:NCBITaxon_7742
 
 # Class: obo:CL_0000086 (germ line stem cell (sensu Nematoda and Protostomia))
 
+AnnotationAssertion(obo:RO_0002161 obo:CL_0000086 obo:NCBITaxon_33511)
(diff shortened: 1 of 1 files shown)


matentzn:
- [ ] Do not merge this PR without a positive review from _both_ @cmungall and @dosumis

dosumis (approved):
Looks good to me.

cmungall:
I was confused looking at this then I realized the "reasoned" view is actually a composition of expanded and reasoned.

I was going to say this PR should update the editors docs on TCs too but looks like we don't have any!

go ahead and merge

cmungall (approved):

shawntanzk:
> I was going to say this PR should update the editors docs on TCs too but looks like we don't have any!

Can do a redirect here: https://oboacademy.github.io/obook/explanation/taxon-constraints-explainer/ and https://oboacademy.github.io/obook/howto/add-taxon-restrictions/

Issue #1695, opened 2022-09-06
[NTR-cxg] BEST4+ intestinal epithelial cell


rachadele:
**Preferred term label**
BEST4+ intestinal epithelial cell

**Synonyms** (add reference(s), please)

**Definition** (free text, with reference(s), please. PubMed ID format is PMID:XXXXXX)
An absorptive epithelial cell of the human intestine expressing Bestrophin 4 calcium-activated ion channels. 
PMID:24223998
PMID:35176508

**Parent cell type term** (check the hierarchy here https://www.ebi.ac.uk/ols/ontologies/cl) 
CL:0000677

**What is the anatomical structure that the cell is a part of?** (check Uberon: https://www.ebi.ac.uk/ols/ontologies/uberon)
intestinal epithelium UBERON:0001277

**Your ORCID**
0000-0002-7431-4139

**Additional notes or concerns** If there are markers that uniquely identify the cell type in a particular species, they can be added here.

bvarner-ebi:
Hi @rachadele, can you clarify if you intend for this class to only represent a cell type in **humans** as noted in your definition? I see the reference paper is focused on humans, but if this cell type exists in other species or genera, we can remove the word "human" from the definition and not worry about taxon restrictions.

bvarner-ebi:
Hi @rachadele, checking in to see if you have feedback on my comment immediately above. Once clarified, I can move forward with adding this new cell type.

lubianat:
@bvarner-ebi As far as I know, they are not known in mice. 

How about having 2 terms, one more neutral, without the taxonomic restriction, and a subclass restricted to humans?

rachadele:
Yes, it is intended to be a human cell type! Sorry for the late response.

bvarner-ebi:
> @bvarner-ebi As far as I know, they are not known in mice.
> 
> How about having 2 terms, one more neutral, without the taxonomic restriction, and a subclass restricted to humans?

Thanks for the feedback, @lubianat. My understanding is that it is discouraged to create new classes that would have only one child. (@shawntanzk, is my assumption correct)? Since there is also no evidence supplied that the cell type exists in other species, I am hesitant to create a taxon-agnostic class without supporting dbxrefs.

shawntanzk:
> My understanding is that it is discouraged to create new classes that would have only one child. (@shawntanzk, is my assumption correct)?

[...]

PR #1696, opened 2022-09-07
Addresses #1695 NTR-cxg BEST4 intestinal epithelial cell

bvarner-ebi:
Addresses #1695 NTR-cxg BEST4 intestinal epithelial cell


shawntanzk:
I would prefer it if the BEST4+ part was axiomatized given its a defining part of the cell type

shawntanzk:
Given the def says human, should you add a human taxon restriction? 
Also can we make this into an equiv class axiom? - not sure if the parent would make it too broad or if it is accurate to do so?

bvarner-ebi:
Hi, @shawntanzk, following the [comments](https://github.com/obophenotype/cell-ontology/issues/1695#issuecomment-1256313376) from the linked issue, can you advise if it makes sense to make this class species agnostic or specific to humans, as originally requested and as the current commit reflects? My concern is that a "middle-ground" solution would create ambiguity, but I'm open to remodelling. If we operate with the open-world assumption, I am thinking the current commit may make more sense for the requester.

shawntanzk:
after skimming the papers - I think its fair to leave this as human specific. 
I would add a (Hsap) at the back - we have done this with BDS stuff where we aren't sure that it exists outside mouse and evidence only suggest existence in mouse
see http://purl.obolibrary.org/obo/CL_4023023 for example

bvarner-ebi:
> after skimming the papers - I think its fair to leave this as human specific.
> I would add a (Hsap) at the back - we have done this with BDS stuff where we aren't sure that it exists outside mouse and evidence only suggest existence in mouse
> see http://purl.obolibrary.org/obo/CL_4023023 for example

Thank you, @shawntanzk. I see examples in CL where " ,human" is appended to the label, e.g., [CD16-positive, CD56-dim natural killer cell, human](http://purl.obolibrary.org/obo/CL_0000939). Does using that convention seem reasonable in this case, for consistency? 
If using (Hsap) is preferable, should all labels in CL be aligned?
Of note, proteins seem to be styled with " (human)" appended to the label.

shawntanzk:
Hmm good point. I'd say stick with human for now (thinking of people searching and all) but would be good to bring this up in a CL call - probably should standardise

shawntanzk (approved):
@bvarner-ebi - could you help bring up the inconsistency in taxon tags in the next CL call please? thanks :)

Issue #1698, opened 2022-09-13
[NTR-HCA] 'immune cell'; superseded by editing synonym for leukocyte


bvarner-ebi:
Context: https://github.com/HumanCellAtlas/ontology/issues/120

**Preferred term label**
immune cell

**Synonyms** (add reference(s), please)
n/a

**Definition** (free text, with reference(s), please. PubMed ID format is PMID:XXXXXX)
A cell that is part of the immune system and helps the body fight infections and other diseases.
https://www.cancer.gov/publications/dictionaries/cancer-terms/def/immune-cell

**Parent cell type term** (check the hierarchy here https://www.ebi.ac.uk/ols/ontologies/cl)
CL:0000988 'hematopoietic cell'

**What is the anatomical structure that the cell is a part of?** (check Uberon: https://www.ebi.ac.uk/ols/ontologies/uberon)
'immune system'

**Your ORCID**
n/a

**Additional notes or concerns**
Use case: Cell type ontology (HCA)

FYI @addiehl

addiehl:
Looking at the source definition, I see that it contains the gloss "These include neutrophils, eosinophils, basophils, mast cells, monocytes, macrophages, dendritic cells, natural killer cells, and lymphocytes (B cells and T cells)." That is pretty much the same subtypes that the existing CL class 'leukocyte' has.  How is 'immune cell' distinct from leukocyte?

I think that 'immune cell' should be added as a 'related synonym' to leukocyte. It's not an exact synonym, because there are non-leukocyte cells that play important roles in the functioning of the immune system (such as follicular dendritic cells, endothelial cells of the high-endothelial venues, platelets, and more), and these cells are important in the functioning of the immune system.

bvarner-ebi:
Per the comments above, "immune cell" will be changed from an exact synonym to a related synonym of the class [leukocyte](http://purl.obolibrary.org/obo/CL_0000738).

PR #1699, opened 2022-09-13
[DO NOT MERGE] Addresses #1698 NTR immune cell

bvarner-ebi:
Addresses #1698 NTR immune cell


addiehl:
I oppose the addition of this cell type as redundant with leukocyte.

bvarner-ebi:
> I oppose the addition of this cell type as redundant with leukocyte.

Thank you, @addiehl. I'll pose your objection to the requester and suggest they consider using 'leukocyte' and that we can add "immune cell" as a related synonym to that class.

bvarner-ebi:
> > I oppose the addition of this cell type as redundant with leukocyte.
> 
> Thank you, @addiehl. I'll pose your objection to the requester and suggest they consider using 'leukocyte' and that we can add "immune cell" as a related synonym to that class.

Actually, @addiehl, I see "immune cell" is already an exact synonym for ['leukocyte'](http://purl.obolibrary.org/obo/CL_0000738). I can propose we change it to a related synonym.

addiehl:
Yes, I think loosening the synonym type to "related" makes sense.

PR #1702, opened 2022-09-15
Addresses #1698 edit synonym for leukocyte

bvarner-ebi:
Addresses #1698 edit synonym for leukocyte

Originally "immune cell" was [requested to be a new term](https://github.com/HumanCellAtlas/ontology/issues/120).
Due to the points made [here](https://github.com/obophenotype/cell-ontology/issues/1698#issuecomment-1245641165), instead of creating a new term, the synonym "immune cell" for [leukocyte](http://purl.obolibrary.org/obo/CL_0000738) has been changed from exact synonym to related synonym.


shawntanzk:
could you check on the use case for immune cell - not sure if this would meet the requesters requirements (do as a follow up)
That being said fully agree with this change so will approve it

shawntanzk (approved):

bvarner-ebi:
> could you check on the use case for immune cell - not sure if this would meet the requesters requirements (do as a follow up)
> That being said fully agree with this change so will approve it

Great, thank you, @shawntanzk. The requester [has commented that this is an acceptable solution](https://github.com/HumanCellAtlas/ontology/issues/120#issuecomment-1248022995).

Issue #1712, opened 2022-10-13
[Typo/Bug] ttp://www.geneontology.org/formats/oboInOwl#created_by


acevedol:
**CL term**
`ttp://www.geneontology.org/formats/oboInOwl#created_by`

**Description of typo, bug or error**

Came across a typo in the cell ontology: `ttp://www.geneontology.org/formats/oboInOwl#created_by`

bvarner-ebi:
@shawntanzk,
does it make sense to update the instance with the created_by that has the typo (pictured below) to the other instance of created_by and then delete the former?

<img width="729" alt="Screenshot 2022-10-19 at 10 44 45" src="https://user-images.githubusercontent.com/95643797/196657108-8ed0ee05-db88-4f2c-84e8-9a27fcecbca6.png">

shawntanzk:
@acevedol thanks so much for reporting this, we really appreciate it. 

@bvarner-ebi yes, I'm guessing you can do this directly in a text editor, but use whichever way you are comfortable with :) thanks!

shawntanzk:
This typo is coming from CARO 
https://ontobee.org/ontology/CARO?iri=http://purl.obolibrary.org/obo/CARO_0020003
will move this ticket there instead

PR #1719, opened 2022-10-19
Addresses #1712 [Typo/Bug] ttp://www.geneontology.org/formats/oboInOwl#created_by

bvarner-ebi:
Addresses obophenotype/caro#36 [Typo/Bug] ttp://www.geneontology.org/formats/oboInOwl#created_by


shawntanzk:
wait, why is this in merged_import? doing changes in merged_import directly will not last, it will end up being changed back once we refresh it, we need to trace the source of the typo

shawntanzk:
This comes from CARO: https://ontobee.org/ontology/CARO?iri=http://purl.obolibrary.org/obo/CARO_0020003 <- you can see it here, will open a ticket there instead

bvarner-ebi:
> This comes from CARO: https://ontobee.org/ontology/CARO?iri=http://purl.obolibrary.org/obo/CARO_0020003 <- you can see it here, will open a ticket there instead

I made the changes in cl-edit.owl using Protégé. Thanks for finding the source of the typo.
Should this PR just be closed without merging?

shawntanzk:
Closing PR as this is a CARO issue, have moved the ticket there

PR #1703, opened 2022-09-15
CL 15 September release

bvarner-ebi:
Thank you for all the help, again, @anitacaron and @shawntanzk.

@shawntanzk, please see especially the changes in the BDS files.

PR #1720, opened 2022-10-28
CL 28 October release

bvarner-ebi:


anitacaron:
Is this right that there isn't any difference in the cl-simple.owl file?

bvarner-ebi:
> Is this right that there isn't any difference in the cl-simple.owl file?

It looks like there are changes....

Screenshot 2022-10-28 at 14 56 56

I don't think the github-actions messages work for release PRs. The same message appeared last month: https://github.com/obophenotype/cell-ontology/pull/1703#issuecomment-1248126922.

shawntanzk:
> Is this right that there isn't any difference in the cl-simple.owl file?

yes, the simple file builds it from edit file and compares it to the existing simple - given there shouldn't be any change in the edit file, there shouldn't be any change in diffs during release :)

shawntanzk (approved):
I assume the changes in `human_reference_atlas` are intentional - just double checking
Other than that looks good, thanks for making the release @bvarner-ebi

anitacaron:
Yes, the subset `human_reference_atlas` is in constant change 😄

Issue #1365, opened 2021-12-10
[Merge] CL:0000439 'prolactin secreting cell' and CL:0002311 'mammotroph'


paola-scibite:
CL:0000439 'prolactin secreting cell':
A peptide hormone cell that secretes prolactin.
adenohypophysis has part some prolactin secreting cell

CL:0002311 'mammotroph' (child of CL:0000439 'prolactin secreting cell'):
An acidophilic cell of the anterior pituitary that produces prolactin.
Subclass of: 
prolactin secreting cell
acidophil cell of pars distalis of adenohypophysis

Are there cells that secrete prolactin and are located in the adenohypophysis other than in its pars distalis (and/or don't stain well with acidic dyes)? Would CL:0000439 'prolactin secreting cell' be applicable to a wider range of species than its child? If not, should the two be merged?

The other child of CL:0000439 'prolactin secreting cell', CL:0002310 'mammosomatotroph', is "An acidophil cell of the anterior _pitiuatry_ gland that produces both prolactin and growth hormone." (note typo). Thanks.

(While looking into this, I also noticed that the spelling "basophillic" with 2 ls is used a few times in CL for "basophilic". Unless basophillic is ever used in literature, please consider correcting to basophilic. (I didn't have time to check PubMed for usage, but Google supports the typo hypothesis.) 

Paola

paolaroncaglia:
Hi CL editors, this ticket was automatically assigned to Nicole when I added the "merge" label, but it may be in scope for human-focused-funding projects. If so, it could also be used to further test @shawntanzk 's merge guidelines if that's still desired (I can't remember the link though!). Either way, thank you.

shawntanzk:
@bvarner-ebi - would you mind taking a look at this and using https://github.com/OBOAcademy/obook/pull/145 to merge the terms? thanks :)

paolaroncaglia:
@bvarner-ebi @shawntanzk thank you both.

bvarner-ebi:
@paola-scibite, thank you for this ticket.

Since there are references ([1](https://pubmed.ncbi.nlm.nih.gov/11015620/), [2](https://doi.org/10.1210/edrv-17-6-639)) that describe extrapituitary prolactin-secreting cells, I see an argument for not merging these terms.

Is it true that `adenohypophysis has part some 'prolactin secreting cell'` does **not** absolutely infer `'prolactin secreting cell' part of some adenohypophysis`? 
If so, I propose the following:

[...]

PR #1723, opened 2022-11-03
Addresses #1365 Fix typos for pituitary and basophilic

bvarner-ebi:
Addresses #1365 Fix typos for pituitary and basophilic


bvarner-ebi:
FYI @paola-scibite

shawntanzk (approved):
Thanks for fixing these typos, btws looking at these terms, we should really add `'has characteristic' some 'basophilic'` to these (PATO:0002094) not part of this ticket so won't hold it back, but something to look into

bvarner-ebi:
> Thanks for fixing these typos, btws looking at these terms, we should really add `'has characteristic' some 'basophilic'` to these (PATO:0002094) not part of this ticket so won't hold it back, but something to look into

Thanks for suggesting this, @shawntanzk. Feel free to create a ticket with your proposal.

Issue #1655, opened 2022-07-11
[NTR] peripheral blood lymphocyte


kalpanaps:
Note - please check that the term does not already exist (check OLS: https://www.ebi.ac.uk/ols/ontologies/cl)

**Preferred term label**: peripheral blood lymphocyte

**Synonyms** (add reference(s), please)

**Definition** (free text, with reference(s), please. PubMed ID format is PMID:XXXXXX): Lymphocyte subsets including T cells (CD3+), (B) B cells (CD19+) and (C) NK cells (CD3− CD56+) PMID:31092833

**Parent cell type term** (check the hierarchy here https://www.ebi.ac.uk/ols/ontologies/cl): Lymphocyte (CL:0000542)

**What is the anatomical structure that the cell is a part of?** (check Uberon: https://www.ebi.ac.uk/ols/ontologies/uberon)

**Your ORCID**:  https://orcid.org/0000-0003-2534-198X

**Additional notes or concerns**

bvarner-ebi:
Thank you for the new term request, @kalpanaps and for the feedback, @addiehl.

After skimming the reference, I am not sure if the text definition is appropriate since it seem to refer to specific cell types that were arbitrarily highlighted in the study. Could there be other subsets that are 'peripheral blood lymphocytes' that are not mentioned in this text definition? @kalpanaps, can you confirm or advise if I am misunderstanding?

@addiehl, would this EA also capture lymphocytes that are in blood but not part of acral / peripheral circulation? It seems like 'blood' == 'peripheral blood' in the ontology. Is that intended?

addiehl:
Peripheral blood is called peripheral presumably because it is taken most often from a peripheral location of the body, such as an arm.  Blood travels rather quickly through the circulatory system, and I believe the assumption is that the composition of blood extracted at a peripheral location is roughly similar to blood throughout the body, apart from oxygen content.  While there are mechanisms for lymphocytes to move into and out of the blood, most researchers seem to assume that the relative abundance and levels of lymphocyte subsets in blood taken at a peripheral location are representative of the blood as a whole.

bvarner-ebi:
[...]

PR #1722, opened 2022-10-31
Compound terms for HubMap

anitacaron:
Context: https://github.com/hubmapconsortium/ccf-ontology/issues/16

Use of the DOS-DP `cellPartOfAnatomicalEntity` to generate cells location specific.

PR #1725, opened 2022-11-04
Addresses #1655 NTR 'peripheral blood lymphocyte'

bvarner-ebi:
Addresses #1655 NTR 'peripheral blood lymphocyte'

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -2782,6 +2782,7 @@ Declaration(Class(obo:CL_4030025))
 Declaration(Class(obo:CL_4030026))
 Declaration(Class(obo:CL_4030027))
 Declaration(Class(obo:CL_4030028))
+Declaration(Class(obo:CL_4030029))
 Declaration(Class(obo:CL_4032000))
 Declaration(Class(obo:CL_4032001))
 Declaration(Class(obo:CP_0000000))
@@ -28943,6 +28944,15 @@ AnnotationAssertion(<http://purl.org/dc/terms/date> obo:CL_4030028 "2022-10-17T1
 AnnotationAssertion(oboInOwl:hasExactSynonym obo:CL_4030028 "amacrine cell, glycinergic"^^xsd:string)
 AnnotationAssertion(rdfs:label obo:CL_4030028 "glycinergic amacrine cell")
 EquivalentClasses(obo:CL_4030028 ObjectIntersectionOf(obo:CL_0000561 ObjectSomeValuesFrom(obo:RO_0002215 obo:GO_0061537)))
+
+# Class: obo:CL_4030029 (peripheral blood lymphocyte)
+
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref <https://orcid.org/0000-0003-2534-198X>) Annotation(oboInOwl:hasDbXref "PMID:31092833"^^xsd:string) obo:IAO_0000115 obo:CL_4030029 "A lymphocyte located in blood."^^xsd:string)
+AnnotationAssertion(<http://purl.org/dc/terms/contributor> obo:CL_4030029 <https://orcid.org/0000-0002-1773-2692>)
+AnnotationAssertion(<http://purl.org/dc/terms/date> obo:CL_4030029 "2022-11-04T14:40:00Z"^^xsd:dateTime)
+AnnotationAssertion(oboInOwl:hasRelatedSynonym obo:CL_4030029 "blood lymphocyte"^^xsd:string)
+AnnotationAssertion(rdfs:label obo:CL_4030029 "peripheral blood lymphocyte")
+EquivalentClasses(obo:CL_4030029 ObjectIntersectionOf(obo:CL_0000542 ObjectSomeValuesFrom(obo:BFO_0000050 obo:UBERON_0000178)))
 
 # Class: obo:CL_4032000 (club-like cell of the urethral epithelium)
 
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -2783,6 +2783,7 @@ Declaration(Class(obo:CL_4030026))
 Declaration(Class(obo:CL_4030027))
 Declaration(Class(obo:CL_4030028))
 Declaration(Class(obo:CL_4030029))
+Declaration(Class(obo:CL_4030030))
 Declaration(Class(obo:CL_4032000))
 Declaration(Class(obo:CL_4032001))
 Declaration(Class(obo:CP_0000000))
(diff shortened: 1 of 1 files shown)


shawntanzk:
I have my reservation about this, but I guess @addiehl is the expert here
Could I convince you to switch the label to be blood lymphocyte and have the related synonym be peripheral blood lymphocyte? seems more appropriate given what you mentioned in the related ticket @bvarner-ebi

addiehl:
'peripheral blood lymphocyte' matches usage in the literature better than 'blood lymphocyte'. If we were to make the term 'blood lymphocyte', I guarantee that someone will come along in a year or two and ask for 'peripheral blood lymphocyte'.  I don't mind making 'blood lymphocyte' an exact or related synonym.

bvarner-ebi:
> I don't mind making 'blood lymphocyte' an exact or related synonym.

Perhaps make 'blood lymphocyte' a broad synonym?

addiehl:
'blood lymphocyte' as a broad synonym works for me.

bvarner-ebi:
@addiehl, @shawntanzk, has the latest commit addressed your concerns?

shawntanzk:
> Perhaps make 'blood lymphocyte' a broad synonym?

By logical definition, this should be an exact synonym instead. PS sorry this slipped through the cracks for me

bvarner-ebi:
> By logical definition, this should be an exact synonym instead. 

Do we need to add 'peripheral blood' to Uberon? It seems like the terms 'blood' and 'peripheral blood' have been used without distinction in CL, likely due to the assumptions @addiehl notes [here](https://github.com/obophenotype/cell-ontology/issues/1655#issuecomment-1271747053). 

If CL assumes 'blood' == 'peripheral blood', it seems reasonable to make the label 'blood lymphocyte' to align with the logical definition and make 'peripheral blood lymphocyte' an exact synonym. If CL does not assume 'blood' == 'peripheral blood', then 'peripheral blood lymphocyte' could be a narrow synonym until 'peripheral blood' is added.

Of note, in BTO, ['peripheral blood'](http://purl.obolibrary.org/obo/BTO_0000553) is a subclass of ['blood'](http://purl.obolibrary.org/obo/BTO_0000089).

shawntanzk:
I'd be alright with either solution (exact synonym or adding term to uberon)
but my preference is exact synonym from what @addiehl said; im thinking about annotation - will having both confuse users?

bvarner-ebi:
> 'd be alright with either solution (exact synonym or adding term to uberon)
> but my preference is exact synonym from what @addiehl said; im thinking about annotation - will having both confuse users?

If this will be the solution, it seems to make sense to make "peripheral blood" an exact synonym to "blood". However, I think one should consider if peripheral blood includes blood in the liver, spleen, umbilical cord, placenta, bone marrow, etc. A cursory web search shows results suggesting the latter are not generally considered "peripheral blood".

shawntanzk:
perhaps youre right, creating the new term - even if users use a different term, queries can be done on blood and it would capture peripheral blood if we follow BTO hierarchy. I do like the cleanliness of creating terms and making equivs on them :)

addiehl:
Still working on my thoughts:
Is a 'peripheral blood lymphocyte' really different (more specialized) than a 'blood lymphocyte'?

Note that Uberon has both ['arterial blood'](http://purl.obolibrary.org/obo/UBERON_0013755) and a ['venous blood'](http://purl.obolibrary.org/obo/UBERON_0013756). Ontologically, both of these might be considered subclasses of 'peripheral blood'.  In practice, peripheral blood is usually collected from a vein in humans (clearly peripheral venous blood), but in experiments with animals (such as mice), it might be collected sub-orbitally, from a tail-bleed, or even directly from a heart puncture, in which case it may be a mix of venous and arterial blood or perhaps neither if taken from the heart.

Should Uberon have a 'peripheral blood term' as the logical union of 'arterial blood' and 'venous blood'?

bvarner-ebi:
> Is a 'peripheral blood lymphocyte' really different (more specialized) than a 'blood lymphocyte'?

CL seems to be going in the direction of creating a new class for a cell type when potentially the only difference is anatomical location (see https://github.com/obophenotype/cell-ontology/pull/1722). Based on  that, there could be justification for creating both  'peripheral blood lymphocyte' and 'blood lymphocyte' cell types even when there is not other distinguishing feature other than location.

addiehl:
I think that creating 'blood lymphocyte' with a 'peripheral blood lymphocyte' term is okay just to end this discussion, but we should avoid further subclassing, like 'peripheral blood T cell'. What annotators are really capturing with these terms is the experimental origin of a blood sample, which in an ideal data annotation world would be captured via co-annotation with an OBI term (or something). We don't even have to create a new Uberon term 'peripheral blood'; we can just have 'peripheral blood lymphocyte'
'blood lymphocyte' AND ('located in' some 'arterial blood' OR 'located in' some 'venous blood')

bvarner-ebi:
@addiehl, @shawntanzk, can you kindly review this PR so it gets captured in the next CL release? I believe all concerns have been addressed with the latest commit.

shawntanzk:
never handled a `or` before - hope ELK handles it >.< 
other than that looks alright to me, but I'll leave @addiehl to approve given he is the expert here :)

balhoff:
ELK doesn't support unions. Unions do work for certain purposes in Whelk, but it's not part of OWL EL.

shawntanzk:
awks, guess we need to handle the logic differently then. If you urgently need the term, I would suggest stripping the logical definition and leaving it as textual for now

bvarner-ebi:
> awks, guess we need to handle the logic differently then. If you urgently need the term, I would suggest stripping the logical definition and leaving it as textual for now

Is this going to break something if kept?

addiehl:
There is precedent for use of OR or union of in CL:
[CD34-positive, CD38-positive common myeloid progenitor OR CD34-positive, CD38-positive common lymphoid progenitor](http://purl.obolibrary.org/obo/CL_0000995)
[CD7-negative lymphoid progenitor OR granulocyte monocyte progenitor](http://purl.obolibrary.org/obo/CL_0001012)
[CD115-positive monocyte OR common dendritic progenitor](http://purl.obolibrary.org/obo/CL_0001019)
[CD117-positive common myeloid progenitor OR CD217-positive common lymphoid progenitor](http://purl.obolibrary.org/obo/CL_0001030)

These CL classes are based on DC_CL [PMID:19243617](https://pubmed.ncbi.nlm.nih.gov/19243617/) from 2009, and represent uncertainty in the knowledge of developmental progenitors for dendritic cells at the time. Improving the representation of dendritic cells has been on my radar for a while. Nevertheless, these union classes do not prevent overall reasoning with ELK, but are probably just ignored.

So I see no major issue with the equivalence axiom as proposed for 'peripheral blood lymphocyte', as Jim has pointed out.

bvarner-ebi:
> So I see no major issue with the equivalence axiom as proposed for 'peripheral blood lymphocyte', as Jim has pointed out.

Thank you, @addiehl. In that case, would you approve this PR so I can proceed with a merge?

bvarner-ebi:
After speaking with @dosumis, it seemed the best way to move forward was to replace the EA of 'peripheral blood lymphocyte' with subclass axioms. I also update the text definition.

@addiehl, if this is an agreeable way forward, I will appreciate your approval of the this PR.

bvarner-ebi:
@shawntanzk, @addiehl, @dosumis, unless there are grave concerns with approving this PR, would one of you mind approving it today so it can be incorporated in a CL release, which is planned for tomorrow AM?

shawntanzk:
was super confused for awhile cause the robot diff still shows EA, then I rmbr I just implemented #gogoeditdiff hence it didn't update lol >.< yeah a lot more comfortable with the subclass

shawntanzk (approved):

PR #1730, opened 2022-11-15
Go Go Release Diff Action

shawntanzk:
#gogoreleasediff

propose to add this to CL now so that when @bvarner-ebi does next release, we can test that it works well in real world situation etc.

### DIFF
--- a/.github/workflows/release-diff.yml
+++ b/.github/workflows/release-diff.yml
@@ -1,4 +1,4 @@
-name: 'Create ROBOT diff on release file'
+name: 'Go Go Release Diff'
 
 on:
   # Triggers the workflow on pull request events for the master branch
@@ -52,7 +52,7 @@ jobs:
           name: release-diff.md
           path: release-diff.md
       - name: Prepare release comment
-        run: "echo \"<details>\n <summary> Here's a diff of how these changes impact cl.owl: </summary> \n\" >comment.md; cat release-diff.md/release-diff.md >>comment.md"
+        run: "echo \"<details>\n <summary> Here's a diff of how this release impacts cl.owl: </summary> \n\" >comment.md; cat release-diff.md/release-diff.md >>comment.md"
         if: steps.check.outputs.triggered == 'true' 
       - name: Post reasoned comment
         if: steps.check.outputs.triggered == 'true' 
--- a/.github/workflows/release-diff.yml
+++ b/.github/workflows/release-diff.yml
@@ -46,6 +46,7 @@ jobs:
           GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
       - uses: actions/checkout@v3
       - name: Download release diff
+        if: steps.check.outputs.triggered == 'true' 
         uses: actions/download-artifact@v2
         with:
           name: release-diff.md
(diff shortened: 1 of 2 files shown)


.github/workflows/release-diff.yml
```diff
@@ -0,0 +1,64 @@
...
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  release_file:
+    runs-on: ubuntu-latest
+    container: obolibrary/odklite:v1.3.0
+    steps:
+      - uses: khan/pull-request-comment-trigger@v1.1.0
+        id: check
+        with:
+          trigger: '#gogoreleasediff'
```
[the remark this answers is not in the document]
shawntanzk: open to suggestions, but gogoreleasediff makes me feel like inspector gadget and i like it :p

.github/workflows/release-diff.yml
```diff
@@ -0,0 +1,64 @@
...
+        env:
+          GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
+      - uses: actions/checkout@v3
+      # Checks-out main branch under "main" directory
+      - uses: actions/checkout@v3
+        with:
+          ref: master
+          path: master
+      - run: export ROBOT_JAVA_ARGS=-Xmx6G; robot diff --labels True --left master/cl.owl --right cl.owl -f markdown -o release-diff.md
```
matentzn, line 29: I wouldnt create a dependency on GitHub here. Many ontologies don't check their releases in:

```
robot diff --labels True --left-iri http://purl.obolibrary.org/obo/cl.owl --right cl.owl -f markdown -o release-diff.md
```

You will also have to create the release first, i.e.

```
make cl.owl IMP=false MIR=false COMP=false
```

Right?

shawntanzk: > You will also have to create the release first, i.e.

I don't see why, for CL at least - we create our releases, create a branch with new files, and this is just to check the diff on it.
I guess if an ontology doesn't check the release in then maybe, but even uberon, we do check uberon.owl in right - was trying to avoid creating the release so it doesn't over use memory and all, and basically trying to minimise fails in actions which freaks people out

shawntanzk: not originally, this stemmed from @dosumis asking for a feature to have human readable diff for release (see https://github.com/obophenotype/cell-ontology/pull/1449#issuecomment-1041357288)

> You could simply ask ODK then to create releasediff.md during the release, and have an action w/o odk scan for this file an post it as a comment when they see it.

probably a smarter idea than this LOL - somehow did not cross my mind at all

>I thought the point of this execise was to so that when someone makes a PR, they can call gogoreleasediff and it will compare the release cl.owl version on the branch with the latest release? 

can easily make this from here, but big ontologies like uberon will struggle again cause of the build process >.<

matentzn: Yeah, there was never going to be a one size fits all - but simply adding your command to the existing diff action would help getting rid of it needlessly creating failures on PRs..

shawntanzk: > You could simply ask ODK then to create releasediff.md during the release, and have an action w/o odk scan for this file an post it as a comment when they see it.

will work on this, and happy to change this to this:

>I thought the point of this execise was to so that when someone makes a PR, they can call gogoreleasediff and it will compare the release cl.owl version on the branch with the latest release?

shawntanzk: @matentzn is there currently a make point that does releasediff.md during release already? I feel like I've seen something like it before, but can't quite find it

shawntanzk: think I found it, sneaky separate makefile heh

shawntanzk:
sorry for the spam >.< im bad with this and have to wait for errors to tell me where to fix things one at a time

PR #1732, opened 2022-11-16
Release diff poster

shawntanzk:
requirement:
Release should post a diff src/ontology/diffs/cl-diff.md

@matentzn - is there a set way to configure release to output cl-diff.md (think uberon does this)


matentzn:
- [ ] Blocked by https://github.com/INCATools/ontology-development-kit/issues/741

dosumis:
@shawntanzk to fix soon if poss.  Otherwise @anitacaron will take over.

ubyndr:
> * [ ]  Blocked by [Add simple diff makefile goal INCATools/ontology-development-kit#741](https://github.com/INCATools/ontology-development-kit/issues/741)

ODK issue should be blocked by this one, shouldn't it? We generally test actions on CL or Uberon then we add them to ODK? @matentzn

bvarner-ebi:
Hi, @anitacaron, is this draft PR still relevant?
If so, can it move forward?
If not, can it be closed?

anitacaron:
The diff file is always too big to fit in the limited characters commentary box (65536 characters). The current diff file has 405117 characters.

I suggest adding a link to the diff file instead.

matentzn (approved):
The only thing I can see that is odd at first sight is that the `comment.md` path is different in both subtasks, but I am sure you tested this so ok! Nice!

anitacaron:
Yeah, it works like this. The comment is [here](https://github.com/obophenotype/cell-ontology/pull/1732#issuecomment-1318422151)

PR #1735, opened 2022-11-17
Add gogoeditdiff trigger

shawntanzk:
#gogoeditdiff

This PR makes it so that the tag #gogoeditdiff needs to be on a comment for the automated diff to run
trial on CL to ensure social side of this works and all before full implementation in ODK

### DIFF
--- a/.github/workflows/diff.yml
+++ b/.github/workflows/diff.yml
@@ -70,6 +70,7 @@ jobs:
         env:
           GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
       - uses: actions/checkout@v3
+        if: steps.check.outputs.triggered == 'true'
         with:
           ref: master
       - name: Classify ontology
@@ -96,6 +97,7 @@ jobs:
         env:
           GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
       - uses: actions/checkout@v3
+        if: steps.check.outputs.triggered == 'true'
       - name: Download master classification
         if: steps.check.outputs.triggered == 'true'
         uses: actions/download-artifact@v2
--- a/.github/workflows/diff.yml
+++ b/.github/workflows/diff.yml
@@ -14,15 +14,25 @@ jobs:
     runs-on: ubuntu-latest
     container: obolibrary/odklite:v1.3.0
     steps:
+      - uses: khan/pull-request-comment-trigger@v1.1.0
+        id: check
+        with:
+          trigger: '#gogoeditdiff'
+        env:
+          GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
       - uses: actions/checkout@v3
+        if: steps.check.outputs.triggered == 'true'
       # Checks-out main branch under "main" directory
       - uses: actions/checkout@v3
+        if: steps.check.outputs.triggered == 'true'
         with:
           ref: master
           path: master
       - name: Diff classification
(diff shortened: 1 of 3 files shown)


.github/workflows/diff.yml
```diff
@@ -14,15 +14,25 @@ jobs:
...
+        if: steps.check.outputs.triggered == 'true'
       # Checks-out main branch under "main" directory
       - uses: actions/checkout@v3
+        if: steps.check.outputs.triggered == 'true'
         with:
           ref: master
           path: master
       - name: Diff classification
+        if: steps.check.outputs.triggered == 'true'
         run: export ROBOT_JAVA_ARGS=-Xmx6G; robot diff --labels True --left master/src/ontology/cl-edit.owl --left-catalog master/src/ontology/catalog-v001.xml --right src/ontology/cl-edit.owl --right-catalog src/ontology/catalog-v001.xml -f markdown -o edit-diff.md
```
ubyndr, line 33: I think you can do make steps for PR and master in here rather than other steps. And also you can add the other robot diff step in this one. You are downloading the master and current branch multiple times in the action. I don't think we need that. @gouttegd and @matentzn can comment on this one as well.

ubyndr (changes requested):

.github/workflows/diff.yml
ubyndr: Like we did in the Uberon diff action, https://github.com/obophenotype/uberon/blob/bd56184537d6503487d94c9103814a9876781acb/.github/workflows/diff.yml#L31-L47

shawntanzk: Oh I think the seperation of master and branch make step is intentional when we copied it over from @balhoff - this allows for simultaneous runs rather than waiting for one to finish to start the next:
Screenshot 2022-11-17 at 15 17 44

ubyndr (approved):

.github/workflows/diff.yml
matentzn, line 33: This does not cover changes to [components](https://github.com/obophenotype/cell-ontology/blob/master/src/ontology/cl-odk.yaml#L89). For the ODK version, it will be better to use the new _intermediate_ ontology which merges components into the edit file prior to further processing.

.github/workflows/diff.yml
```diff
@@ -31,10 +41,19 @@ jobs:
     runs-on: ubuntu-latest
     container: obolibrary/odklite:v1.3.0
     steps:
+      - uses: khan/pull-request-comment-trigger@v1.1.0
+        id: check
+        with:
+          trigger: '#gogoeditdiff'
```
matentzn, line 47: there could be a different trigger for the reasoned version to safe resources for larger ontologies?

.github/workflows/diff.yml
```diff
@@ -14,15 +14,25 @@ jobs:
     runs-on: ubuntu-latest
     container: obolibrary/odklite:v1.3.0
     steps:
```
matentzn, line 16: Given the total runtime, there should be check to ensure of the branch is up to date with master, and fail if it is not (or better yet, post a comment saying its failed because its not up to date).

matentzn (approved):

matentzn (changes requested):
Sorry acceidentally approved.

.github/workflows/diff.yml
shawntanzk: if we want a different trigger, i would suggest two separate actions, we can split it out

Issue #1736, opened 2022-11-17
[NTR] valve interstitial cell (heart)


emquardokus:
Note - please check that the term does not already exist (check OLS: https://www.ebi.ac.uk/ols/ontologies/cl)

**Preferred term label**
valve interstitial cell

**Synonyms** (add reference(s), please)
VIC
cardiac valve interstitial cell

**Definition** (free text, with reference(s), please. PubMed ID format is PMID:XXXXXX)

Ayoub, Salma, Chung-Hao Lee, Kathryn H. Driesbaugh, Wanda Anselmo, Connor T. Hughes, Giovanni Ferrari, Robert C. Gorman, Joseph H. Gorman, and Michael S. Sacks. “Regulation of Valve Interstitial Cell Homeostasis by Mechanical Deformation: Implications for Heart Valve Disease and Surgical Repair.” Journal of The Royal Society Interface 14, no. 135 (October 31, 2017): 20170580. https://doi.org/10.1098/rsif.2017.0580. (Table 1. Key Studies With Original Methodology Revealing Important Properties of Isolated, Cultured VICs)

Liu, Amber C, Vineet R Joag, and Avrum I Gotlieb. “The Emerging Role of Valve Interstitial Cell Phenotypes in Regulating Heart Valve Pathobiology.” The American Journal of Pathology 171, no. 5 (November 2007): 1407–18. https://doi.org/10.2353/ajpath.2007.070251.

Rutkovskiy, Arkady, Anna Malashicheva, Gareth Sullivan, Maria Bogdanova, Anna Kostareva, Kåre‐Olav Stensløkken, Arnt Fiane, and Jarle Vaage. “Valve Interstitial Cells: The Key to Understanding the Pathophysiology of Heart Valve Calcification.” Journal of the American Heart Association 6, no. 9 (n.d.): e006339. https://doi.org/10.1161/JAHA.117.006339.

This paper refers to them as "fibroblast-like"
Xu, Kang, Shangbo Xie, Yuming Huang, Tingwen Zhou, Ming Liu, Peng Zhu, Chunli Wang, et al. “Cell-Type Transcriptome Atlas of Human Aortic Valves Reveal Cell Heterogeneity and Endothelial to Mesenchymal Transition Involved in Calcific Aortic Valve Disease.” Arteriosclerosis, Thrombosis, and Vascular Biology 40, no. 12 (December 1, 2020): 2910–21. https://doi.org/10.1161/ATVBAHA.120.314789.

[...]

PR #1761, opened 2022-12-06
Addresses #1736 NTR 'valve interstitial cell'

bvarner-ebi:
#1736


bvarner-ebi:
#gogoeditdiff

src/ontology/cl-edit.owl
```diff
@@ -2797,6 +2798,7 @@ Declaration(Class(obo:CP_0000039))
 Declaration(Class(obo:CP_0000040))
 Declaration(Class(obo:CP_0000043))
 Declaration(Class(obo:D96882F1-8709-49AB-BCA9-772A67EA6C33))
+Declaration(Class(obo:GO_0070278))
```
shawntanzk, line 2801: I think not the greatest idea to declare a GO term, does protege do this?

shawntanzk:
hmmm strange that that didnt trigger actions... will just review without it now but probably should look into whats going on

src/ontology/cl-edit.owl
```diff
@@ -28983,6 +28984,19 @@ AnnotationAssertion(<http://purl.org/dc/terms/date> obo:CL_4030031 "2022-11-30T1
 AnnotationAssertion(rdfs:label obo:CL_4030031 "interstitial cell")
 EquivalentClasses(obo:CL_4030031 ObjectIntersectionOf(obo:CL_0000003 ObjectSomeValuesFrom(obo:BFO_0000050 obo:UBERON_0005169)))
 
+# Class: obo:CL_4030032 (valve interstitial cell)
+
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref <https://orcid.org/0000-0001-7655-4833>) Annotation(oboInOwl:hasDbXref "PMID:25311230"^^xsd:string) Annotation(oboInOwl:hasDbXref "https://doi.org/10.1098/rsif.2017.0580"^^xsd:string) Annotation(oboInOwl:hasDbXref "https://doi.org/10.1161/ATVBAHA.120.314789"^^xsd:string) Annotation(oboInOwl:hasDbXref "https://doi.org/10.1161/JAHA.117.006339"^^xsd:string) Annotation(oboInOwl:hasDbXref "https://doi.org/10.2353/ajpath.2007.070251"^^xsd:string) obo:IAO_0000115 obo:CL_4030032 "A cardiac valve cell that resides in a valve leaflet. Along with valve endothelial cells, a valve interstitial cell maintains tissue homeostasis for the function of cardiac valves through secreting biochemical signals, matrix proteins and matrix remodeling enzymes."^^xsd:string)
```
shawntanzk, line 28989: I would prefer the definiendum to be "An interstitial cell
that is part of the cardial valve" given the logical definition.

shawntanzk:
Just to double check @gouttegd  are the changes in the sssom file correct? 

@bvarner-ebi, not a big deal if you rather not change your protocol, and I don't think its bad to make those changes during the pipeline, but if you want to avoid pulling unrelated others to review this, you can undo the changes made to files you don't want to touch :) I'm sure there's some ODK way to not touch those, but I usually just let it do its thing and remove stuff i don't want lol.

gouttegd:
@shawntanzk Yes, the changed in the SSSOM file reflect changes that were made in the past two months on the FBbt side. We released a FBbt version last week, so the changes got picked up here when imports were refreshed.

PR #1772, opened 2022-12-15
CL 15 December 2022 release

bvarner-ebi:
CL 15 December 2022 release


bvarner-ebi:
Removing dbxref annotation on 'valve interstitial cell' (inadvertently added during term creation).

bvarner-ebi:
Cursory review: expected new terms and edits have been verified.

shawntanzk (approved):
All looks good
Would be good to revert the changes in docs.yml -> however @anitacaron suggests we remove it altogether. 
Something to double check -> the removal of `subset: human_reference_atlas` are intentional. They do look like they are.

cl-base.obo
```diff
@@ -23306,11 +24477,9 @@ relationship: BFO:0000050 UBERON:0002367
 id: CL:1000301
 name: fibroblast of subepithelial connective tissue of prostatic gland
 def: "A fibroblast that is part of the subepithelial connective tissue of prostatic gland." [GOC:tfm]
-subset: human_reference_atlas
```
bvarner-ebi, line 23309: @anitacaron, I believe you commented on edits to the human_reference_atlas before. I don't interact with this subset... do you know where these changes originate?

anitacaron: Yes, these changes are linked to the changes on the ASCT+B tables. No need to worry about them.

anitacaron:
@bvarner-ebi Please, remove the change in docs because the diff here was a manual change, which should not have happened because this file comes from ODK. 😄

anitacaron:
Sorry, it's correct this way.

anitacaron (approved):

PR #1787, opened 2023-01-09
Added branch status warning to diff report

### LINKED ISSUE #1738: Add check to ensure of the branch is up to date with master
Automated diff should check if branch is up to date with master before allowing the run

ubyndr:
Fixes #1738 

I have added `branch_status` step at the top. It checks if the current branch is behind compared to master branch, and make a comment to warn user. It fails after the comment process.

### DIFF
--- a/.github/workflows/diff.yml
+++ b/.github/workflows/diff.yml
@@ -203,4 +203,3 @@ jobs:
         with:
           file: "../../edit-comment.md"
           identifier: "UNREASONED"
-          file: "../../warncomment.md"
--- a/.github/workflows/diff.yml
+++ b/.github/workflows/diff.yml
@@ -1,13 +1,40 @@
 name: 'Create ROBOT diffs on Pull requests'
 
-on: 
+on:
 
   issue_comment:
         types: [created]
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
+  branch_status:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Check current branch status
+        id: info
+        run: |
+          echo "status=$(git rev-list --left-right --count origin/$DEFAULT_BRANCH...origin/${GITHUB_HEAD_REF} | awk '{print $1}')"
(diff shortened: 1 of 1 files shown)


.github/workflows/diff.yml
```diff
@@ -1,13 +1,40 @@
 name: 'Create ROBOT diffs on Pull requests'
 
-on: 
+on:
```
hrshdhgd, line 3: Shouldn't this workflow run after every commit made to a PR? I'm making the assumption that every open branch is associated with a PR. So with that in mind, shouldn't there be
```
on:
  pull_request:
    branches: [ main ]
```

matentzn: Ah no, this action fires when a specific command is spoken in a comment on the PR, like "kgcl" bot.

matentzn (approved):

.github/workflows/diff.yml
```diff
@@ -1,13 +1,40 @@
...
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+        uses: NejcZdovc/comment-pr@v1.1.1
+        with:
+          file: "../../warncomment.md"
+      - name: fail
+        if: steps.info.outputs.status >= 1
+        run: |
+          echo "Your branch is $STATUS commit/s behind, please update your branch.   "
+          exit 1
```
shawntanzk, line 35: does this actually hard fail? I think I tried this once with another strategy and it still continues. But I guess you've tested it :)

matentzn: Good point @shawntanzk I have assumed this was tested, but yea better be sure

ubyndr: @shawntanzk unless you use `continue-on-error: true` or `exit 0` it interrupts the run.

bvarner-ebi:
Hi, @ubyndr, checking to see if this PR can be merged. 
If there are pending items, could you list them so they can be addressed?

ubyndr:
> Hi, @ubyndr, checking to see if this PR can be merged. If there are pending items, could you list them so they can be addressed?

Hi @bvarner-ebi , it can be merged.

PR #1788, opened 2023-01-09
Change deploy_release to use release_assets instead of assets 

shawntanzk:
Not sure if this should stay in the end of the day, but for consistency, im making PR here since we have uploaded it this round. If we decide not to have it, can either revert or just not merge this.

### DIFF
--- a/src/ontology/cl.Makefile
+++ b/src/ontology/cl.Makefile
@@ -242,6 +242,7 @@ DEPLOY_GH=true
 cl:
 	$(MAKE) prepare_release IMP=false PAT=false 
 	$(MAKE) release-diff
+	cp imports/merged_import.owl ../../imports
 	if [ $(DEPLOY_GH) = true ]; then 	$(MAKE) deploy_release GHVERSION="v$(TODAY)"; fi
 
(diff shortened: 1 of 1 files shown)


anitacaron:
Ok, I looked into makefile, and it should change the variable in to RELEASE_ASSETS instead of ASSETS.

```
FILTER_OUT=../patterns/definitions.owl ../patterns/pattern.owl reports/cl-edit.owl-obo-report.tsv
MAIN_FILES_RELEASE = $(foreach n, $(filter-out $(FILTER_OUT), $(ASSETS)), ../../$(n))
```

shawntanzk:
ok tested it out, and i think you're right. Have remove the cp part in makefile and remove the merged_import - I think its ok to leave this release as it is , but next one will be fixed.

anitacaron:
Ok, I'll do this already for the UBERON release.

anitacaron:
Can you change the PR's title, please?

anitacaron:
What do you think about this change, @matentzn?

matentzn:
No objections

anitacaron (approved):

Issue #1785, opened 2023-01-05
[Synonym] Add syn to 'stromal cell of endometrium' 


bvarner-ebi:
Related to https://github.com/obophenotype/cell-ontology/issues/1760.

Per literature (e.g., [PMID:30309298](https://pubmed.ncbi.nlm.nih.gov/30309298/), https://doi.org/10.1038/s41591-020-1040-z), an endometrial stromal fibroblast differentiates into a decidual stromal cell during the process of decidualization.
This is currently reflected in CL with the following axiom under [CL:2000002 'decidual cell'](http://purl.obolibrary.org/obo/CL_2000002):

[develops from](https://www.ebi.ac.uk/ols/ontologies/cl/properties?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FRO_0002202) some [stromal cell of endometrium](https://www.ebi.ac.uk/ols/ontologies/cl/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FCL_0002255)

--

Given this context, I propose the following:

- [ ] Add synonym to 'stromal cell of endometrium':
"endometrial stromal fibroblast" (exact, PMID:30309298)

 - [ ] Add the following two axioms to 'stromal cell of endometrium':
['has potential to directly develop into'](http://purl.obolibrary.org/obo/RO_0002388) some ['decidual cell'](http://purl.obolibrary.org/obo/CL_2000002)
['part of'](http://purl.obolibrary.org/obo/BFO_0000050) some ['endometrial stroma'](http://purl.obolibrary.org/obo/UBERON_0002337)

- [ ] Revise the text definition of 'stromal cell of endometrium':
CURRENT: Connective tissue cells of the endometrium, these cells are fibroblastic in appearance. During decidualization they tranform into decidual (stromal) cells, swellling up and adopt an epithelial cell-like appearance due to the accumulation of glycogen and lipid droplets.
REVISED: A stromal cell of the endometrium that is fibroblastic in appearance. During decidualization this cell may differentiate into a decidual stromal cell. (Add dbxrefs [PMID:30309298](https://pubmed.ncbi.nlm.nih.gov/30309298/) and https://doi.org/10.1038/s41591-020-1040-z)

PR #1794, opened 2023-01-13
Addresses #1785 Edits to 'stromal cell of endometrium'

bvarner-ebi:
Addresses #1785 Edits to 'stromal cell of endometrium'


aleixpuigb (approved):
Everything looks good

PR #1803, opened 2023-01-16
Fixes #1389

### LINKED ISSUE #1389: Create a DP for has-soma-location pattern
So far we have a DP documented for cells part of anatomy:

https://obophenotype.github.io/cell-ontology/patterns/cellPartOfAnatomicalEntity/

But none for has-soma-location(HSL)

We need to add these, and make the ontology consistent (which @shawntanzk has been doing -see #1337) 

We need to document whether

 1. We *only* use HSL for neurons->uberon
 2. We use a mixture of part-of and HSL as appropriate
     - typically use HSL for equivalence axioms (principle: weaker properties in EC -> stronger inferences)
     - add supplemental part-of hidden GCIs if we know the neuron is not projecting outside the anatomical area

While 1 is both safer and easier to implement, I worry we will lose crucial inferences in CL

In particular I worry about loss in:

 - gene expression queries that propagate over part-of
 - entailments in ontologies that use CL for logical definitions

To restore these inferences we will need to add a lot of hidden GCIs

## Alternative proposal: radical mereological inclusivity

We abandon HSL, we go back to using simple powerful part-of relations.

We conceive of uberon brain regions such as hippocampus as having 'classic boundaries' at one level of granularity, but when zoomed in to the subcellular level the hippocampus itself has synaptic parts that interleave with other brain regions. These could be described using simple projects-into relationships.

This would require some minor changes in uberon: the existing adjacency relations (not that useful IMO) would need to be changed to something like 'grossly adjacent to'.

Gene expression queries would be more inclusive but arguably more correct since expression is in the soma.

dosumis:
Fixes #1389

- #1389

### DIFF
--- a/src/patterns/dosdp-patterns/neuronHasSomaLoc
+++ b/src/patterns/dosdp-patterns/neuronHasSomaLoc
@@ -32,13 +32,13 @@ data_list_vars:
 name:
   text: "%s of %s"
   vars:
-    - cell
+    - neuron
     - anatomical_entity
 
 def:
   text: "Any %s whose soma is located in a(n) %s."
   vars:
-    - cell
+    - neuron
     - anatomical_entity
   annotations:
     - annotationProperty: xref
@@ -47,5 +47,5 @@ def:
 equivalentTo:
   text: "%s and ('has soma location' some %s)"
   vars:
-    - cell
+    - neuron
     - anatomical_entity
--- a/src/patterns/dosdp-patterns/neuronHasSomaLoc
+++ b/src/patterns/dosdp-patterns/neuronHasSomaLoc
@@ -1,0 +1,51 @@
+pattern_name: neuronHasSomaLoc
+pattern_iri: http://purl.obolibrary.org/obo/cl/cellHasSomaLoc
+
+description: "Use this template to define a class of neurons by a combination of
+  gross type and soma locataion. e.g. cortical interneuron."
+#  examples:
+#    - http://purl.obolibrary.org/obo/CL_0008031  # cortical interneuron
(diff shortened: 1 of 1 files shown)


src/patterns/dosdp-patterns/neuronHasSomaLoc
```diff
@@ -0,0 +1,51 @@
...
+  anatomical_entity: "'anatomical entity'"
+
+data_list_vars:
+  pubs: "'xref'"
+  # Use this for recording reference using PMID/DOIS
+
+name:
+  text: "%s of %s"
+  vars:
+    - cell
```
shawntanzk, line 35: Should this var be 'neuron'?

dosumis: Yes.  Please fix.  Thanks!

bvarner-ebi:
Checking in on this PR before upcoming CL release.
No description, no action for a month and referenced ticket is already closed.
Can this PR be merged or closed?

anitacaron:
The change in this PR is not relevant to the CL release.

src/patterns/dosdp-patterns/neuronHasSomaLoc
anitacaron: All other 'cell' should be changed to 'neuron'?

anitacaron: done

anitacaron (approved):

bvarner-ebi:
@dosumis, can this be merged and closed?

Issue #1784, opened 2023-01-03
[Merge] 'neuroendocrine cell' and 'neurosecretory neuron'


bvarner-ebi:
**CL terms**
CL:0000165 'neuroendocrine cell' 

to subsume:

CL:0000381  'neurosecretory neuron' (has 1 child class, no text definition)

**Reason for deprecation**
'neuroendocrine cell' (which has exact synonym "neurosecretory cell") and 'neurosecretory neuron' may all refer to the same cell type.
https://doi.org/10.1093/icb/7.1.161

shawntanzk:
Given that the EA of neurosecretory neuron is
```
neuron
 and ('capable of' some 'hormone secretion')
```
I think this merge is sensible.
However, I think we need to be careful of the use of neurosecretory as basically all neurons "secretes" neurotransmitters right? and not all NTs are hormones. My suggestion would be to make neurosecretory neuron a related or broad synonym

bvarner-ebi:
>However, I think we need to be careful of the use of neurosecretory as basically all neurons "secretes" neurotransmitters right? and not all NTs are hormones. My suggestion would be to make neurosecretory neuron a related or broad synonym

Thanks for the feedback.
After a cursory search, the term "neurosecretion" appears to have inconsistencies in how it is defined and used (the release of any substance from a neuron vs the release of hormones from a neuron.)

In addition to the reference in the original comment, these additional references suggest that this term is used specifically in association with hormone release rather than release of any substance (neurotransmitters, etc):

https://link.springer.com/chapter/10.1007/978-3-030-22989-4_1
https://www.sciencedirect.com/topics/neuroscience/neurosecretory-cell
https://encyclopedia2.thefreedictionary.com/Neurosecretion
https://www.britannica.com/science/neurosecretory-cell

bvarner-ebi:
Given the "evolution" in meaning and potential ambiguity of the term, there are sources that treat it as a broader term and others that treat it as an exact synonym. With this in mind, I would annotate it as a related synonym. 

Consider adding a rdfs:comment or editor's note regarding "The modern definition of neurosecretion has evolved ..."

aleixpuigb:
Thank you for your input.

PR #1804, opened 2023-01-18
Addresses #1784 merge

aleixpuigb:
Addresses #1784 merge neuroendocrine cell and neurosecretory neuron


bvarner-ebi (changes requested):
There appear to be unintended changes to mappings.owl.

anitacaron:
@gouttegd, can you verify the changes in the `components/mappings.owl`, please?

gouttegd:
Almost all changes to `mappings.owl` are formatting changes, apart from this axiom that was added:

```
AnnotationAssertion(<http://www.geneontology.org/formats/oboInOwl#hasDbXref> <http://purl.obolibrary.org/obo/CL_0000165> "FBbt:00005130"^^xsd:string)
```

Have you manually added this cross-reference to `CL:0000165`? If so, that’s not how cross-references to FBbt are handled. They are managed on the FBbt side, and the `mappings.owl` file is automatically generated from what FBbt provides.

Currently, `FBbt:00005130` is mapped to `CL:0000381`. Since the CL term is now merged with `CL:0000165`, I will update the mapping to make it point to `CL:0000165`, and this will be automatically picked up by CL at the next import refresh.

bvarner-ebi:
> Have you manually added this cross-reference to CL:0000165? I

@aleixpuigb, can you clarify if you manually made any changes to the mappings.owl file.
@gouttegd, thank you for checking this.

aleixpuigb:
Thank you @anitacaron and @gouttegd for looking into this.

> Have you manually added this cross-reference to CL:0000165? 

I have not editted the `mapping.owl` file. However, Protégé didn't allow me to set the new synonym CL:0000165 'neurosecretory neuron' as xsd:string type, so I did it manually using Atom on the `cl-edit.owl` file.

src/ontology/cl-edit.owl
```diff
@@ -4849,12 +4849,16 @@ SubClassOf(obo:CL_0000164 obo:CL_0000163)
 # Class: obo:CL_0000165 (neuroendocrine cell)
 
 AnnotationAssertion(Annotation(oboInOwl:hasDbXref "MESH:D055099"^^xsd:string) obo:IAO_0000115 obo:CL_0000165 "An endocrine cell that has the specialized function to produce and secrete hormones in response to neuronal signals."^^xsd:string)
+AnnotationAssertion(oboInOwl:hasAlternativeId obo:CL_0000165 "CL:0000381")
 AnnotationAssertion(oboInOwl:hasDbXref obo:CL_0000165 "BTO:0002691"^^xsd:string)
 AnnotationAssertion(oboInOwl:hasDbXref obo:CL_0000165 "FMA:83810"^^xsd:string)
 AnnotationAssertion(oboInOwl:hasExactSynonym obo:CL_0000165 "neurosecretory cell"^^xsd:string)
+AnnotationAssertion(oboInOwl:hasExactSynonym obo:CL_0000165 "neurosecretory neuron"^^xsd:string)
```
bvarner-ebi, line 4856: Consider adding dbxref as noted in ticket.

src/ontology/cl-edit.owl
```diff
@@ -4849,12 +4849,16 @@ SubClassOf(obo:CL_0000164 obo:CL_0000163)
...
 AnnotationAssertion(rdfs:label obo:CL_0000165 "neuroendocrine cell"^^xsd:string)
+EquivalentClasses(obo:CL_0000165 ObjectIntersectionOf(obo:CL_0000540 ObjectSomeValuesFrom(obo:RO_0002215 obo:GO_0046879)))
```
bvarner-ebi, line 4858: It looks like this was carried over from the obsoleted term, which came from the FBbt definition.

Consider adding comment from the reference in the ticket:

"The neurosecretory cell is neither an ordinary neuron nor an endocrine cell, but a combination of both. Its neuronal features resemble those of ordinary neurons concerning both structure and function. The production of a visible secretory material marks the neurosecretory neuron as a gland cell, and the fact that extractable cellular products act in the manner of hormones places it in the realm of endocrine elements."

src/ontology/cl-edit.owl
```diff
@@ -7798,7 +7804,6 @@ AnnotationAssertion(oboInOwl:inSubset obo:CL_0000548 <http://purl.obolibrary.org
 AnnotationAssertion(rdfs:label obo:CL_0000548 "animal cell"^^xsd:string)
 EquivalentClasses(obo:CL_0000548 ObjectIntersectionOf(obo:CL_0000003 ObjectSomeValuesFrom(obo:RO_0002160 obo:NCBITaxon_33208)))
 SubClassOf(obo:CL_0000548 obo:CL_0000255)
-SubClassOf(obo:CL_0000548 ObjectSomeValuesFrom(obo:RO_0002160 obo:NCBITaxon_33208))
```
bvarner-ebi, line 7801: Was this intended?

bvarner-ebi (commented):
If you confirm line 7801 was intentionally deleted, I will approve.
Otherwise, please see other inline comments.

src/ontology/cl-edit.owl
aleixpuigb: Thank you for noticing @bvarner-ebi , I have fixed it

gouttegd:
> I have not edited the mapping.owl file.

OK, sorry to bother you with that but I’d like to understand how that file got modified.

One way this could have happened is if you cut the `FBbt:00005130` cross-reference from the `CL:0000381` class and pasted it into the `CL:0000165` class (in Protégé, not by manually editing `cl-edit.owl`). Is that what you did?

This would explain the changes to `mappings.owl` because, since the cross-reference on `CL:0000381` is defined in `mappings.owl`, when you cut-and-paste it Protégé would keep it in its original file.

aleixpuigb:
> One way this could have happened is if you cut the FBbt:00005130 cross-reference from the CL:0000381 class and pasted it into the CL:0000165 class (in Protégé, not by manually editing cl-edit.owl). Is that what you did?

I don't think I could do that because `FBbt:00005130` does not appear on Protégé and it has not been modified on `cl-edit.owl`. I haven't done it manually either.

src/ontology/cl-edit.owl
aleixpuigb: Added with the comment

aleixpuigb:
> Have you followed the [merging procedure](https://oboacademy.github.io/obook/howto/merge-terms/) described in the OBOOK?

Yes, that's the procedure I followed

gouttegd:
Perfect, then it’s normal that the `mappings.owl` file got modified (it’s a side-effect of that procedure), even though it’s unfortunate.

We need better documentation to explain to editors what to do when something like that happen (basically, that they should _not_ commit the changes to the `mappings.owl` file).

Thanks for helping me sorting that out.

aleixpuigb:
Thank you @gouttegd. I will not do it again!

gouttegd:
Not your fault! There was nothing in the documentation to warn you about this problem.

bvarner-ebi (approved):

bvarner-ebi:
Thanks for the follow up, @gouttegd.
@aleixpuigb, after merging, can you please announce the obsoleted term in the cl-editors Slack channel?
If you update the [merging procedure](https://oboacademy.github.io/obook/howto/merge-terms/) documentation to reflect @gouttegd's comments, you can assign me the PR to review.

gouttegd:
> If you update the [merging procedure](https://oboacademy.github.io/obook/howto/merge-terms/) documentation

Not sure the OBOOK is the right place. The `mappings.owl` component is specific to Uberon/CL. Other ontologies don’t have have such a component – or even worse, they could have a similarly named component that is used for other purposes and that is subject to different rules.

I think what we need is some documentation directly in Uberon’s `docs/` folder, describing what this component is, how it is produced, and what to do when it is inadvertently modified (then CL can simply point to the Uberon document, to avoid duplicating it).

For better or for worse, the mapping component is my thing, so I’ll take care of that.

src/ontology/cl-edit.owl
```diff
@@ -4850,10 +4850,10 @@ SubClassOf(obo:CL_0000164 obo:CL_0000163)
 
 AnnotationAssertion(Annotation(oboInOwl:hasDbXref "MESH:D055099"^^xsd:string) obo:IAO_0000115 obo:CL_0000165 "A neuron that is capable of some hormone secretion in response to neuronal signals."^^xsd:string)
 AnnotationAssertion(oboInOwl:hasAlternativeId obo:CL_0000165 "CL:0000381")
-AnnotationAssertion(oboInOwl:hasBroadSynonym obo:CL_0000165 "neurosecretory cell"^^xsd:string)
-AnnotationAssertion(oboInOwl:hasBroadSynonym obo:CL_0000165 "neurosecretory neuron"^^xsd:string)
 AnnotationAssertion(oboInOwl:hasDbXref obo:CL_0000165 "BTO:0002691"^^xsd:string)
 AnnotationAssertion(oboInOwl:hasDbXref obo:CL_0000165 "FMA:83810"^^xsd:string)
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref "https://www.sciencedirect.com/topics/biochemistry-genetics-and-molecular-biology/neurosecretion"^^xsd:string) Annotation(rdfs:comment "The modern definition of neurosecretion has evolved to include the release of any neuronal secretory product from a neuron.") oboInOwl:hasRelatedSynonym obo:CL_0000165 "neurosecretory cell"^^xsd:string)
```
bvarner-ebi, line 4855: Recommend setting the dbxref to the exact reference (it looks like it is ISBN:9780323711548). I think the current URL goes to a collection of several sources and those results can change over time.

bvarner-ebi (approved):
Of note, clinicalkey.com dbxref behind a paywall.

src/ontology/cl-edit.owl
aleixpuigb: > those results can change over time.
Good point! 

I chose that reference because I'm referencing the synonym, not the comment. However, it is better to provide the source of the comment

bvarner-ebi: I think the reason QC checks are failing is because the comment should not be an annotation of an annotation. Consider consolidating this comment with your other comment or try adding a second comment that is not under the synonym annotation.

aleixpuigb: Thank you, I thought there was a problem with the reference but I couldn't find it.

PR #1808, opened 2023-01-19
Add symbols for BDS terms

shawntanzk:
drafting first, probably need a look through to standardise things a bit more

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -3010,6 +3010,7 @@ Declaration(ObjectProperty(<http://purl.obolibrary.org/obo/cl#has_low_plasma_mem
 Declaration(ObjectProperty(<http://purl.obolibrary.org/obo/cl#has_not_completed>))
 Declaration(ObjectProperty(<http://purl.obolibrary.org/obo/cl#lacks_part>))
 Declaration(ObjectProperty(<http://purl.obolibrary.org/obo/cl#lacks_plasma_membrane_part>))
+Declaration(AnnotationProperty(obo:IAO_0000028))
 Declaration(AnnotationProperty(obo:IAO_0000115))
 Declaration(AnnotationProperty(obo:IAO_0000116))
 Declaration(AnnotationProperty(obo:IAO_0000231))
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -3064,7 +3064,7 @@ AnnotationAssertion(rdfs:label obo:IAO_0000115 "definition"^^xsd:string)
 
 AnnotationAssertion(rdfs:label obo:IAO_0000424 "expand expression to"^^xsd:string)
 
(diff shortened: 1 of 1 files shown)


shawntanzk:
#gogoeditdiff

shawntanzk:
Not sure why, but it isnt importing IAO_0000028 (symbol) in despite it being present in the term file. 
Don't think its a bit deal for now, but we should defs look into it down the road if people are going to be using that AP

anitacaron:
IAO annotations are all on OMO. It needs to add to the import list.

shawntanzk:
hmmm @anitacaron I've added OMO and all but doesn't seem to want to import IAO_0000028 (symbol) still

dosumis:
Does the import system work for APs?  Has never been clear to me.

shawntanzk:
> Does the import system work for APs? Has never been clear to me.

that is a good question, don't think I've imported an AP before hmmm - if we cant import it, how do we use it? just with a label?

anitacaron:
I'm updating this issue to try to merge it if it's still pertinent.

1. Symbol is not in OMO
2. Symbol is not an annotation property, but a class https://www.ebi.ac.uk/ols4/ontologies/iao/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FIAO_0000028
3. If still needed, should import IAO instead of OMO.

anitacaron:
I can fix the many conflicts if this is still relevant.

shawntanzk:
I think these are still useful, though honestly at this point, I think its easier to just reopen a new PR and use the diff here to add those symbols in. Happy to do it, but ODK is hard for me to access here, so would much prefer it if someone could create a branch with symbol being imported from IAO, and I can work from there. Thanks :)

dosumis:
Would like symbol added ASAP.  Anyone know if we can legally pun class -> AP?

hkir-dev:
class-AP punning is allowed https://www.w3.org/TR/owl2-new-features/#F12:_Punning
class-datatype punning and punning among different property types are not allowed.

anitacaron:
@shawntanzk, we can always update the PR, no need to create another one. I'll update this PR.

dosumis:
> class-AP punning is allowed https://www.w3.org/TR/owl2-new-features/#F12:_Punning

Seems like we need a dedicated make goal that adds AP declaration for IAO classes (although I think OMO was meant to solve this)

anitacaron:
Can it be declared in the `cl-edit.owl` file?

anitacaron:
@shawntanzk, we need to redefine it as an annotation property because it's officially a class.

anitacaron (approved):

shawntanzk:
> @shawntanzk, we need to redefine it as an annotation property because it's officially a class.

oh yeah, that totally makes sense :) Thanks @anitacaron, youre the best!

shawntanzk:
Ok really didnt have to gogoedit diff lol, looks good, merging in to make next release :) thanks all!

PR #1818, opened 2023-01-20
Addresses #1816 Add synonym "metazoan cell" to "animal cell"

### LINKED ISSUE #1816: [Question] Are all animals part of Metazoa?
**Background**
The logical definition of the term CL:0000548 "animal cell" is the following:

'native cell'
 **and** ('only in taxon' **some** Metazoa)

By definition, Metazoan are multicellular animals, what suggests: a) There are unicellular animals that are not Metazoan, or b) all animals are multicellulars, and therefore Metazoan.

**Question**
Is there any animal that is not part of Metazoa?

aleixpuigb:
Resolves #1816 Add synonym "metazoan cell" to "animal cell"

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -7790,6 +7790,7 @@ SubClassOf(obo:CL_0000547 ObjectSomeValuesFrom(obo:RO_0002202 obo:CL_0000038))
 
 # Class: obo:CL_0000548 (animal cell)
 
+AnnotationAssertion(oboInOwl:hasExactSynonym obo:CL_0000548 "metazoan cell")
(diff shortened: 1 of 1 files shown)


bvarner-ebi (commented):
Unsure of the source for all the RO and GO changes...

Optional:  consider adding an editor's note for context on using "metazoa" and "animal" interchangeably (CL uses NCBI taxonomy as standard).

aleixpuigb:
There was an automatic rearrangement when I added the synonym, but it was not intended.

anitacaron:
Protege sorts the content differently than using a text editor. If there isn't a fundamental change, this is good.

bvarner-ebi (approved):

Issue #1726, opened 2022-11-06
[NTR]perineurial cell


emquardokus:
Note - please check that the term does not already exist (check OLS: https://www.ebi.ac.uk/ols/ontologies/cl)

**Preferred term label**
perineurial cell

**Synonyms** (add reference(s), please)

**Definition** (free text, with reference(s), please. PubMed ID format is PMID:XXXXXX)
A cell that belongs to the supporting tissue surrounding a nerve fiber bundle. It has thin long bipolar cytoplasmic processes, pinocytotic vesicles, fragments of external lamina and/or external lamina-like material, attachment plaques, and desmosome-like junctions. [ NCI ] 

NCIT:C41442 http://purl.obolibrary.org/obo/NCIT_C41442

**Parent cell type term** (check the hierarchy here https://www.ebi.ac.uk/ols/ontologies/cl)
perineural cell http://purl.obolibrary.org/obo/CL_0002576
based on placement in NCIT perineural cell http://purl.obolibrary.org/obo/NCIT_C41416

**What is the anatomical structure that the cell is a part of?** (check Uberon: https://www.ebi.ac.uk/ols/ontologies/uberon)
perineurium 

**Your ORCID**
0000-0001-7655-4833

**Additional notes or concerns**
HuBMAP Skin ASCT+B table term

bvarner-ebi:
@emquardokus provided review article [PMID:36244793](https://pubmed.ncbi.nlm.nih.gov/36244793/), which states:

"Previous studies—including this one— largely accept the idea that the perineurial cell is a modified fibroblast with myofibroblastic features (Bunge et al. 1989; Vega et al. 1996; Du Plessis et al. 1997)."

I do not think this justifies the EA:

```
 'myofibroblast cell'
 and ('part of' some perineurium)
```
Looking forward to additional input on this and the proposed next steps above.

bvarner-ebi:
Hi @emquardokus, @dosumis, any additional thoughts on the proposed next steps above?

bvarner-ebi:
@emquardokus, thank you for this term request.

As commented above, the class CL:0002576 'perineurial cell' has been updated.

PR #1832, opened 2023-02-03
Addressess #1726 Edit perineurial cell

bvarner-ebi:
Addressess #1726 Add / Edit 'perineurial cell'


src/ontology/cl-edit.owl
```diff
@@ -19317,14 +19317,16 @@ AnnotationAssertion(rdfs:label obo:CL_0002575 "central nervous system pericyte"^
...
 SubClassOf(Annotation(oboInOwl:is_inferred "true") obo:CL_0002575 obo:CL_0000669)
 
-# Class: obo:CL_0002576 (perineural cell)
+# Class: obo:CL_0002576 (perineurial cell)
 
-AnnotationAssertion(Annotation(oboInOwl:hasDbXref "GO:tfm"^^xsd:string) Annotation(oboInOwl:hasDbXref "http://en.wikipedia.org/wiki/Perineurium"^^xsd:string) obo:IAO_0000115 obo:CL_0002576 "A myofibroblast that lies in the connective tissue of the spinal cord that has a distinctly lamellar arrangement."^^xsd:string)
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref "GO:tfm"^^xsd:string) Annotation(oboInOwl:hasDbXref "NCIT:C41442"^^xsd:string) Annotation(oboInOwl:hasDbXref "PMID:25818566"^^xsd:string) Annotation(oboInOwl:hasDbXref "http://en.wikipedia.org/wiki/Perineurium"^^xsd:string) Annotation(rdfs:comment <https://orcid.org/0000-0001-7655-4833>) obo:IAO_0000115 obo:CL_0002576 "A glial cell that is part of the perineurium. This cell type has thin long bipolar cytoplasmic processes, pinocytotic vesicles, fragments of external lamina and/or external lamina-like material, attachment plaques, and desmosome-like junctions."^^xsd:string)
```
aleixpuigb, line 19322: Consider adding that perineurial cells (unlike fibroblasts) have double basement membranes (os basal lamina).

Note: Probably more for perineurium, but an important function of perineurial cells is being "a component of the blood–nerve barrier, protecting the interior of the nerve from toxins, infection, and ionic flux" and are connected via tight junctions.

bvarner-ebi: Thanks for the feedback, @aleixpuigb.

1- Can you provide a ref to support the first statement?

2- Agreed. Would you kindly update perineurium in Uberon so it is part of the ['blood nerve barrier'](http://purl.obolibrary.org/obo/UBERON_0003209)? Then the latter term can be imported to CL. Ideally 'perineurial cell' will be inferred to be part of the 'blood nerve barrier' as well.

aleixpuigb: 1- It's from PMID:25818566
"Individual perineurial cells within this structure are extremely thin, have a double basal lamina, and are fitted together by overlapping border zones"
Additionally, perineurial cells have double basement membranes, whereas fibroblasts only have a single basal laminae
2- Ok, I will do that

aleixpuigb (approved):

src/ontology/cl-edit.owl
dosumis: Content looks great, but spotted some grammatical issues with this text:

> Perineurial cells historically have been referred to as fibroblasts because of shape; however, unlike fibroblasts, a perineurial cell: does not have a compact nucleus and large endoplasmic reticulum; does have a double basement membrane opposed to a single basal lamina; are carefully joined by tight junctions into a single sheet as opposed to arranged in a large mass; and finally, can surround a small axon bundle at a nerve terminal whereas a fibroblast cannot."^^xsd:string)

"**_a_** perineurial cell... **_are_** carefully joined by tight junctions"

Issue #1817, opened 2023-01-20
[NTR] endometrial ciliated cell


s-carney:
Note - please check that the term does not already exist (check OLS: https://www.ebi.ac.uk/ols/ontologies/cl)

**Preferred term label**
endometrial ciliated cell

**Synonyms** (add reference(s), please)
ciliated cell of the endometrium
uterine ciliated cell
ciliated cell of the uterus

**Definition** (free text, with reference(s), please. PubMed ID format is PMID:XXXXXX)
A ciliated cell that lines the endometrium where the cilia help move mucus from the glandular tissue [PMID: 31290979]

**Parent cell type term** (check the hierarchy here https://www.ebi.ac.uk/ols/ontologies/cl)
http://purl.obolibrary.org/obo/CL_0000064

**What is the anatomical structure that the cell is a part of?** (check Uberon: https://www.ebi.ac.uk/ols/ontologies/uberon)
http://purl.obolibrary.org/obo/UBERON_0004811

bvarner-ebi:
Thanks for the new term request, @s-carney.

In another ticket (#1760), it is planned to add 3 new terms (from https://www.nature.com/articles/s41591-020-1040-z):

‘endometrial ciliated epithelial cell’
with new child classes:
‘luminal endometrial ciliated epithelial cell’
‘glandular endometrial ciliated epithelial cell’

If you agree, I will merge the details from this request with creating the new term ‘endometrial ciliated epithelial cell’.

s-carney:
Hello, thanks for your response. I agree the other tickets describes well the term suggested.

bvarner-ebi:
Thanks for the reply, @s-carney. 
The reference provided also uses organoids in the study, which is outside the domain of CL. 
If you'd like to have your requested synonyms added to the new term, a non-organoid reference or your ORCID could be added to support their inclusion.

s-carney:
My primary source was the Human Protein Atlas:
https://www.proteinatlas.org/humanproteome/single+cell+type/Glandular+epithelial+cells#endometrialciliatedcells
Perhaps if this is not sufficient, I can find another reference.

bvarner-ebi:
@s-carney,

Thank you for this ticket. The following have been added to CL and should appear in the next release:

CL:4030040 ‘endometrial ciliated epithelial cell’
CL:4030041 ‘luminal endometrial ciliated epithelial cell’
CL:4030042 ‘glandular endometrial ciliated epithelial cell’

[...]

Issue #1760, opened 2022-12-05
Add terms for endometrium


dosumis:
Potential new terms for uterine muscle and epithelial cell types: https://www.nature.com/articles/s41588-021-00972-2

dosumis:
Hmmm - yes.  We need to be careful about that.  We need references for these type _in vivo_.  IIRC there are other single cell studies on endometrium.  Please take a look at CxG & maybe search pubmed.  IIRC Sarah Teichman's group has published on this.

bvarner-ebi:
This may be a place to start: https://www.nature.com/articles/s41591-020-1040-z
Of note, cell types found in the endometrium will be especially impacted by cell state given the cyclical remodelling of the endometrium.

bvarner-ebi:
After a cursory review of https://www.nature.com/articles/s41591-020-1040-z, here are some potential new cell types. Inviting feedback from @dosumis, @emquardokus and other interested parties.

----

term from paper: 
**“stromal fibroblast”** 
-- appears to be same as [CL:0002255 ‘stromal cell of endometrium’](http://purl.obolibrary.org/obo/CL_0002255)
action:
**- suggested revisions in https://github.com/obophenotype/cell-ontology/issues/1785**

----

term from paper:
**“endothelium”**
-- assume the paper uses this term to refer to endothelial cells
-- main text does not appear to explore this cell type in depth (other than differentially expressed genes)
-- [CL:2000091 ‘endometrial microvascular endothelial cell’](http://purl.obolibrary.org/obo/CL_2000091) exists, but this may be a different cell type
action:
**- seek additional references to support potential new term ‘endothelial cell of endometrium’**

----

term from paper:
**"macrophage"**
-- main text does not appear to explore this cell type in depth (other than differentially expressed genes)
action:
**- seek additional references to support potential new term ‘endometrial macrophage’**

----

term from paper:
**"lymphocyte"**
-- main text does not appear to explore this cell type in depth (other than differentially expressed genes)
action:
**- seek additional references to support potential new term ‘endometrial lymphocyte’**

----

term from paper:
**“ciliated epithelial cell”** 
action:
**- add 3 new classes: 
‘endometrial ciliated epithelial cell’
with new child classes:
‘luminal endometrial ciliated epithelial cell’
‘glandular endometrial ciliated epithelial cell’**

----

term from paper:
**“unciliated epithelial cell”** 
action:
**- add 2 new classes:
‘endometrial unciliated epithelial cell’
[...]

PR #1843, opened 2023-02-09
Addresses #1817 NTR endometrial ciliated cells

bvarner-ebi:
Addresses #1817 NTR endometrial ciliated cells (3)
Also related to #1760


aleixpuigb (approved):
The new terms and imports are good. The edits in the run.sh file and cl-odk.yaml I believe they are in line with what you were discussing with @anitacaron.

bvarner-ebi:
> The new terms and imports are good. The edits in the run.sh file and cl-odk.yaml I believe they are in line with what you were discussing with @anitacaron.

Yes, the edits regarding memory were made to allow importing the Uberon term. Before the edits, the import was failing due to memory constraints.

PR #1852, opened 2023-02-20
Document policy on delegating import refresh to the tech support group.

### LINKED ISSUE #1766: Import terms workflow
Off topic for this PR but: adding IRIs directly has the great advantage that it can be used instantly without having to wait for someone to help or disrupting your editing flow run a slow import update.  I wonder if we should tolerate it and have a clean-up job that removes stray declarations from the edtor's file. Alternatively maybe we could have some super minimal temp import update (say  IRI + label only) that editors can use rapidly while working on a PR, postponing a full update to later  - Maybe add to agenda for Monday tech call?

_Originally posted by @dosumis in https://github.com/obophenotype/cell-ontology/issues/1763#issuecomment-1344427397_

gouttegd:
Add a section to the `contributing` document explaining that refreshing the imports _may_ be delegated to someone from the "tech support" group, by tagging a PR with the dedicated `update-imports-required` label.

closes #1766


matentzn (approved):
Looks good!

Issue #1855, opened 2023-02-22
enteroendocrine cell


dosumis:
Text definition: An endocrine cell that is located in the epithelium of the gastrointestinal tract or in the pancreas.

Potential formal definition via GCIs:

endocrine cell that part_of some pancreas subClassOf 'enterodendocrine cell'
endocrine cell that part_of some stomach subClassOf 'enterodendocrine cell'
endocrine cell that part_of some intestine subClassOf 'enterodendocrine cell'

I'm assuming that gastrointestinal tract here should include stomach and intestine but not eosphagus

aleixpuigb:
Given this text definition, ['endocrine cell of pancreas' CL:0008024](https://www.ebi.ac.uk/ols4/ontologies/cl/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FCL_0008024) should be an enterodendocrine cell of pancreas.

dosumis:
> Given this text definition, ['endocrine cell of pancreas' CL:0008024](https://www.ebi.ac.uk/ols4/ontologies/cl/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FCL_0008024) should be an 'enterodendocrine cell of pancreas'.

The formalization should infer this.  + fits with classifiction of subclasses of 'endocrine cell of pancreas'

PR #1860, opened 2023-02-24
Addresses #1855 endocrine cell

aleixpuigb:
Addresses #1855 endocrine cell with formal definition via GCIs.


aleixpuigb:
A new PR will be created with edits from protege 5.6.1

PR #1864, opened 2023-02-28
Fixes #1855 enteroendocrine cell

### LINKED ISSUE #1855: enteroendocrine cell

aleixpuigb:
Fixes #1855 enteroendocrine cell

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -30846,5 +30846,8 @@ AnnotationAssertion(rdfs:label obo:D96882F1-8709-49AB-BCA9-772A67EA6C33 "obsolet
 AnnotationAssertion(owl:deprecated obo:D96882F1-8709-49AB-BCA9-772A67EA6C33 "true"^^xsd:boolean)
 
 
+SubClassOf(ObjectIntersectionOf(obo:CL_0000163 ObjectSomeValuesFrom(obo:BFO_0000050 obo:UBERON_0000160)) obo:CL_0000164)
+SubClassOf(ObjectIntersectionOf(obo:CL_0000163 ObjectSomeValuesFrom(obo:BFO_0000050 obo:UBERON_0000945)) obo:CL_0000164)
+SubClassOf(ObjectIntersectionOf(obo:CL_0000163 ObjectSomeValuesFrom(obo:BFO_0000050 obo:UBERON_0001264)) obo:CL_0000164)
 SubClassOf(Annotation(oboInOwl:hasDbXref <https://github.com/obophenotype/cell-ontology/issues/757>) ObjectIntersectionOf(obo:CL_0000540 ObjectSomeValuesFrom(obo:RO_0002162 obo:NCBITaxon_7742)) ObjectSomeValuesFrom(obo:RO_0002202 obo:CL_0000031))
 )


dosumis:
#gogoeditdiff

dosumis:
This one is interesting:  

> neuroendocrine cell of epithelium of crypt of Lieberkuhn http://purl.obolibrary.org/obo/CL_4033002
> Removed
> [neuroendocrine cell of epithelium of crypt of Lieberkuhn](http://purl.obolibrary.org/obo/CL_4033002) SubClassOf [intestinal epithelial cell](http://purl.obolibrary.org/obo/CL_0002563)
> Added
> [neuroendocrine cell of epithelium of crypt of Lieberkuhn](http://purl.obolibrary.org/obo/CL_4033002) SubClassOf [intestinal enteroendocrine cell](http://purl.obolibrary.org/obo/CL_1001516)

I think it makes sense, but I wonder if it would be counter-intuitive to experts.

bvarner-ebi (commented):
Is it intentional to drop all the epithelial cell inferences?

The GCIs themselves seem reasonable.

Regarding ‘neuroendocrine cell of epithelium of crypt of Lieberkuhn’, what sticks out to me is that the level of specificity in anatomical location is not aligned with the other terms — this term is at the histological level while all other classes in the ‘enteroendocrine cell’ branch are at the gross level, at least as described in the labels.

Screenshot 2023-03-02 at 09 23 18

I also have not found compelling evidence that this cell type exists in normal tissue, especially with the current EA of ‘neuroendocrine cell’ defining it as a neuron.

Kindly consider the comments above, and I will defer to the original requestor's review/approval for this.

aleixpuigb:
Thank you for your comments, @bvarner-ebi 

> Is it intentional to drop all the epithelial cell inferences?

My understanding is that if a parent term (even if it is inferred) has a class inferred, all the descendant terms will no get that class inferred. This might be to avoid having terms with a large number of inferences. In this case, pancreatic endocrine cell is now a subclassOf `epithelial cell of pancreas`, and therefore, all the subclasses drop all the subclasses drop it.

> I also have not found compelling evidence that this cell type exists in normal tissue, especially with the current EA of ‘neuroendocrine cell’ defining it as a neuron.

Enterochromaffin cells are neuroendorin and enteroendocrin cells that is located in intestinal epithelium (not restricted to the crypt) in homeostatic conditions.

dosumis (approved):
I think that the issues around [neuroendocrine cell of epithelium of crypt of Lieberkuhn](http://purl.obolibrary.org/obo/CL_4033002), while interesting, are not blockers here. Happy to merge and move this discussion eleswhere.  

@aleixpuigb - can you follow up with @emquardokus & gut experts about whether 'neuroendocrine cell of epithelium of crypt of Lieberkuhn' refers to EC cells in the crypt?

aleixpuigb:
> IIRC - two terms were merged recently. I suspect that was a bad idea.

Yes, [I merged neuroendocrine cell and neurosecretory neuron](https://github.com/obophenotype/cell-ontology/issues/1784). Looking at he [literature](https://academic.oup.com/icb/article/7/1/161/168831?login=true):
> The neurosecretory cell is neither an ordinary neuron nor an endocrine cell, but a combination of both.

refers to them as `neurosecreotry neurons` or `neurosecreotry cells` indistinctly, which the latest was already a synonym on neuroendocrine cells. Looking at [pulmonary neuroendocrine cells](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC7774893/), they are considered epithelial cells.

The usage of `neuroendocrine cell` and `neuroendocrine neuron` as synonym is also used in wikipedia to define [Magnocellular neuroendocrine cells](https://en.wikipedia.org/wiki/Magnocellular_neurosecretory_cell) (not magnocellular cells, which are neurons).

Therefore, it seems that the three terms have been used as synonyms, although sometimes they describe epithelial cells and   others to describe neurons. This term will need a bit of work in a separate ticket.

Issue #1870, opened 2023-03-02
[NTR] Retinal astrocyte


aleixpuigb:
Note - please check that the term does not already exist (check OLS: https://www.ebi.ac.uk/ols/ontologies/cl)

**Preferred term label**
retinal astrocyte

**Synonyms** (add reference(s), please)

**Definition** (free text, with reference(s), please. PubMed ID format is PMID:XXXXXX)
A star-shaped glial cell that is part of some retina.
PMID:25236977

**Parent cell type term** (check the hierarchy here https://www.ebi.ac.uk/ols/ontologies/cl)
astrocyte (CL:0000127)

**What is the anatomical structure that the cell is a part of?** (check Uberon: https://www.ebi.ac.uk/ols/ontologies/uberon)
retina (UBERON:0000966)

**Your ORCID**
http://orcid.org/0000-0001-6677-8489

**Additional notes or concerns**
In the human retina, retinal astrocytes might have the following markers: GFAP-positive, SOD3-positive and GYPC-positive. PMID:32555229

PR #1871, opened 2023-03-02
Addresses #1870 NTR retinal astrocyte

aleixpuigb:
Addresses #1870 NTR retinal astrocyte

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -2936,6 +2936,7 @@ Declaration(Class(obo:CL_4033011))
 Declaration(Class(obo:CL_4033012))
 Declaration(Class(obo:CL_4033013))
 Declaration(Class(obo:CL_4033014))
+Declaration(Class(obo:CL_4033015))
 Declaration(Class(obo:CP_0000000))
 Declaration(Class(obo:CP_0000025))
 Declaration(Class(obo:CP_0000027))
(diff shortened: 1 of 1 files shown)


src/ontology/cl-edit.owl
```diff
@@ -30775,6 +30776,16 @@ AnnotationAssertion(rdfs:label obo:CL_4033014 "peg cell")
 SubClassOf(obo:CL_4033014 obo:CL_0000066)
 SubClassOf(obo:CL_4033014 ObjectSomeValuesFrom(obo:BFO_0000050 obo:UBERON_0004804))
 
+# Class: obo:CL_4033015 (retinal astrocyte)
+
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref "PMID:25236977") Annotation(oboInOwl:hasDbXref "PMID:33796062") obo:IAO_0000115 obo:CL_4033015 "A star-shaped glial cell that is part of some retina. This cell links neurons to blood vessels and provides structural and physiological support to optic nerve head axons. In response to elevated intraocular pressure, a retinal astrocyte modulates extracellular matrix remodelation.")
```
bvarner-ebi, line 30781: Consider:

"... This cell links neurons to blood vessels and may provide structural and physiological support to optic nerve head axons. In response to elevated intraocular pressure, a retinal astrocyte may modulate extracellular matrix remodeling."

The final sentence could also be placed in a comment if you'd like to have a leaner text definition.

bvarner-ebi (changes requested):
See inline comments.

bvarner-ebi (approved):
Regarding the comment, the genes listed have been noted to be differentially expressed in larger amounts in the cell type. Unsure if that means the same as "might express".

Otherwise OK.

PR #1903, opened 2023-03-21
Remove CHEBI import from ODK config

anitacaron:
Also updating imports to removed not used CHEBI classes.


bvarner-ebi (commented):
@anitacaron, is the removal of the many GO terms from merged_import.owl intended?
If so, I will approve.

bvarner-ebi (approved):
Approving after offline conversation with @anitacaron.

PR #1904, opened 2023-03-21
Release-2023-03-21b

ghost:
Release-2023-03-21b

Replacing https://github.com/obophenotype/cell-ontology/pull/1902

status post PR #1903 merge.


aleixpuigb:
I could not use the diff file as it is 5 times bigger. However, I've had a look with protege and everything looks good.

anitacaron:
To clarify, the diff file was more significant because of the removal of all axioms from more than 200 CHEBI terms.

Issue #1768, opened 2022-12-12
Metazoan cell types are not consistently subclasses of 'animal cell'


cmungall:
```
alias cl='runoak -i sqlite:obo:cl'
cl info .desc//p=i cell .not .desc//p=i 'animal cell'
```

340 cell types
[non-animal-cl.txt](https://github.com/obophenotype/cell-ontology/files/10211790/non-animal-cl.txt)

some are correctly species-general (although the value of some of these groupings is questionable). E.g.

CL:0011026 ! progenitor cell
CL:0000064 ! ciliated cell

but many clearly should be animal cells

CL:0010007 ! His-Purkinje system cell
CL:0009003 ! larval midgut cell
CL:1000147 ! heart valve cell

confirmed by looking at OLS:

<img width="1133" alt="image" src="https://user-images.githubusercontent.com/50745/207149358-20eacd32-970d-45de-89d8-e91e5c765263.png">

discussing solutions with @dosumis 

1. change patterns to use animal cell as genus
    - would still need additional assertions?
2. use TC in logical def and auto-classify
3. abandon 'animal cell as grouping'

dosumis:
I'd like to keep animal cell as a grouping.  This might be easy to fix manually by moving grouping classes.  If not, I suggest SPARQL to change all using of 'cell' in genus position (subClassOf or EquivalentTo) and curation to switch back the relatively small number that are incorrect.  @ubyndr  should be able to design Sparql update.  @bvarner-ebi  could work on curation.

@bvarner-ebi   -  I think that based on this, we should not use the animal_cell tag in CAP.  I think just switch this to cell.

ubyndr:
Adding this term list,  [output.csv](https://github.com/obophenotype/cell-ontology/files/10236431/output.csv), upon @bvarner-ebi request. He will curate this list to whether to add them as `animal cell` or not.

dosumis:
@bvarner-ebi Tracking sheet [here](https://docs.google.com/spreadsheets/d/1io4vDR4pvQhNiWsMZt45SAvR-duxmOc14aLEC4OhTe0/edit#gid=0) @bvarner-ebi - I see lot of terms here flagged as animal cell.  Can all the non-controversial ones just be moved, or if they have already, could you link the relevant PR?

bvarner-ebi:
Thanks for adding the link above.

Below are the cell type classes that are probably considered non-controversial to add as subclasses of 'animal cell'. PR to be linked. 

[...]

PR #1906, opened 2023-03-23
Addresses #1768 Metazoan cell types are not consistently subclasses of 'animal cell'

bvarner-ebi:
Added "less controversial" cell types as subclasses under 'animal cell'.

Thank you to @ubyndr and @ar-ibrahim for helping do this using a "ROBOT template".


ar-ibrahim (approved):

bvarner-ebi:
@ar-ibrahim, it looks like some labels were added as duplicates on some of the terms. Should the template include the LABEL column if the IRI is already there?

src/ontology/cl-edit.owl
```diff
@@ -22030,8 +22135,10 @@ AnnotationAssertion(<http://purl.org/dc/terms/contributor> obo:CL_0007021 <https
 AnnotationAssertion(oboInOwl:hasExactSynonym obo:CL_0007021 "alarm substance secreting cell")
 AnnotationAssertion(Annotation(oboInOwl:hasDbXref "https://github.com/obophenotype/cell-ontology/issues/583") oboInOwl:hasExactSynonym obo:CL_0007021 "club cell (zebrafish)")
 AnnotationAssertion(Annotation(oboInOwl:hasDbXref "ISBN-10:0412429403") Annotation(oboInOwl:hasDbXref "https://github.com/obophenotype/cell-ontology/issues/583") oboInOwl:hasRelatedSynonym obo:CL_0007021 "club cell")
+AnnotationAssertion(rdfs:label obo:CL_0007021 "alarm substance cell")
 AnnotationAssertion(rdfs:label obo:CL_0007021 "alarm substance cell"@en)
```
ubyndr, line 22139: It seems like label with language tag caused the duplication

ar-ibrahim:
I usually use the robot template to create new EFO terms so I do add the label column. I don't think you need to do that in your case. I suggest you remove that column, run the robot template command again and then check to see if the duplicated terms have been removed.

ubyndr (approved):

PR #1921, opened 2023-04-04
Remove deprecated class from count

shawntanzk:
I assume we do not want to count deprecated terms, so adding filters to the sparql
feel free to close if we actually do want to count deprecated terms for some reason 
probably should do this on uberon too
dont know if this is an ODK shipped out sparql (did not find it on dynamic file jinja, but if it is, might want to do this there too. 

Thanks :)


shawntanzk:
No idea lol, I just use it for stats when I needed before and recently. Happy for you to close this without merging. Just thought since I made the mistake with the assumption, others might. But I have my own query saved that I will use so up to you :)

PR #1932, opened 2023-04-13
Obsolete 'primary pigment cell'.

### LINKED ISSUE #1930: [Obsolete] subclasses of _visual pigment cell (sensu Nematoda and Protostomia)_
[visual pigment cell (sensu Nematoda and Protostomia)](http://purl.obolibrary.org/obo/CL_0001658) has three subclasses:

* [primary pigment cell](http://purl.obolibrary.org/obo/CL_0000727)
* [secondary pigment cell](http://purl.obolibrary.org/obo/CL_0000728)
* [tertiary pigment cell](http://purl.obolibrary.org/obo/CL_0000729)

They have no definitions and are not referenced from anywhere. Given that the parental term is already mapped to its _Drosophila_ counterpart ([FBbt:00004230](http://purl.obolibrary.org/obo/FBbt_00004230)), I don’t think there is a need in CL for those subclasses. People who need to refer to a precise subtype of pigment cell can use one of the FBbt terms (if they’re fly people), or equivalent terms in other nematode or arthropod ontologies.

gouttegd:
The term _primary pigment cell_, which refers to one type of visual pigment cells in protostomes, is deemed to be unnecessary in CL, and is therefore obsoleted. Suggestion is to use either the closest parental term (_visual pigment cell (sensu Nematoda and Protostomia)_) to refer to any kind of visual pigment cell regardless of their exact type, or a more precise term from a taxon-specific ontology such as FBbt.

Likewise for _secondary_ and _tertiary pigment cell_.

closes #1930

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -9485,21 +9485,21 @@ SubClassOf(obo:CL_0000726 obo:CL_0000605)
 # Class: obo:CL_0000727 (obsolete primary pigment cell)
 
 AnnotationAssertion(obo:IAO_0000233 obo:CL_0000727 "https://github.com/obophenotype/cell-ontology/issues/1930")
-AnnotationAssertion(rdfs:comment obo:CL_0000727 "Obsoleted  as this level of precision is deemed unnecessary for CL. Consider using either the more general CL:0001658 or, if you need to refer to primary pigment cells specifically, a term from a taxon-specific ontology such as FBbt:00004231.")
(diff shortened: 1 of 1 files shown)


bvarner-ebi (approved):
Of note, there does not seem to be a standard for how to structure text definitions for obsoleted terms in CL. Some start using "OBSOLETE: ", "Obsolete: ", "OBSOLETE.", "OBSOLETED." no indication, etc.

Instead of creating another variation, ("Obsoleted" followed by 2 white spaces), I recommend selecting one of the options above.

gouttegd:
The use of two spaces after “Obsoleted” was accidental anyway, I was not trying to invent a new convention. :) I’ll fix that before merging.

Of note, I am waiting a bit to see if someone can weigh in on the use on `oboInOwl:seeAlso` to link to the original issue, since there are (at least) 3 different properties that can be used for that purpose (`oboInOwl:seeAlso`, `rdfs:seeAlso`, and `IAO:0000233`) and I am not aware of an OBO-wide consensus on that topic.

Issue #1926, opened 2023-04-06
[Typo/Bug] Equivalence axiom of 'Schwann cell' too broad


bvarner-ebi:
**CL term**
CL:0002573 'Schwann cell'

**Description of typo, bug or error**
[Whilst reviewing a PR](https://github.com/obophenotype/cell-ontology/pull/1920#discussion_r1159492931), it was realised that the current equivalence axiom for '[Schwann cell](http://purl.obolibrary.org/obo/CL_0002573)' would incorrectly classify some glial cells as Schwann cells.

**CURRENT EQUIVALENCE AXIOM:**

`'glial cell' and ('part of' some 'peripheral nervous system')`

This causes an 'enteroglial cell' to be inferred as a 'Schwann cell', which seems to be inaccurate.

See also https://doi.org/10.1002/9781118402061.ch8:
There are four major types of glia in the PNS. They are myelinating and non-myelinating Schwann cells, satellite glial cells (SGCs), enteric glial cells (EGCs), and olfactory ensheathing cells (OECs). 

Based on the above, not all glial cells of the PNS are 'Schwann cells'.

**PROPOSED FIX:**
- Remove current equivalence axiom
- add subclass axioms:
   - 'glial cell'
   - 'part of' some 'peripheral nervous system'
- Fix grammar in text definition

aleixpuigb:
@bvarner-ebi, can this issue be closed?

PR #1936, opened 2023-04-14
Addresses #1926 fix equivalence axiom of 'Schwann cell'

bvarner-ebi:
Addresses #1926 fix equivalence axiom of 'Schwann cell'

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -5737,8 +5737,8 @@ AnnotationAssertion(owl:deprecated obo:CL_0000260 "true"^^xsd:boolean)
 
 # Class: obo:CL_0000261 (obsolete anterior cell)
 
-AnnotationAssertion(oboInOwl:consider obo:CL_0000261 "false"^^xsd:boolean)
 AnnotationAssertion(oboInOwl:consider obo:CL_0000261 "http://purl.obolibrary.org/obo/DDANAT_0000002"^^xsd:anyURI)
+AnnotationAssertion(oboInOwl:consider obo:CL_0000261 "false"^^xsd:boolean)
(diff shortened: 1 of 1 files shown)


src/ontology/cl-edit.owl
```diff
@@ -19367,14 +19367,15 @@ SubClassOf(Annotation(oboInOwl:is_inferred "true") obo:CL_0002572 obo:CL_0000134
 
 # Class: obo:CL_0002573 (Schwann cell)
 
-AnnotationAssertion(Annotation(oboInOwl:hasDbXref "GOC:tfm") obo:IAO_0000115 obo:CL_0002573 "A glial cell that ensheathes axons of neuron in the peripheral nervous system and are necessary for their maintainance and function.")
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref "GOC:tfm") obo:IAO_0000115 obo:CL_0002573 "A glial cell that ensheathes an axon in the peripheral nervous system, which is necessary for the maintenance and function of the ensheathed neurons.")
```
aleixpuigb, line 19370: After a cursory review, it seems that this definition is incomplete:

`One of the most important functions of the Schwann cell is to myelinate the axons of the PNS. Myelin, which is a fatty layer that insulates the axon, helps to increase the saltatory conduction of the neuron. A myelinating Schwann cell wraps around a single axon. Myelination occurs in larger diameter axons (≳1 μm), while a process called ensheathment occurs on smaller diameter axons in which a single cell wraps around multiple small, unmyelinated axons separating them with a thin layer of cytoplasm.` [PMID:16807057](https://doi.org/10.1016/j.biocel.2006.05.007)

`Axons regulate the Schwann cell phenotype, determining whether they myelinate individual axons or ensheathe multiple, small axons in Remak bundles.`  [PMID:23279426](https://pubmed.ncbi.nlm.nih.gov/23279426)

The current definition seems to be focused on non-myelinating Schwann cells.

aleixpuigb (changes requested):
Please, see the above comment. The PR is not connected to the original issue, is that intended?

src/ontology/cl-edit.owl
bvarner-ebi: Agreed. Text def updated, comment added and dbxref added.

aleixpuigb (approved):

PR #1950, opened 2023-04-19
Obsolete 'primary neuron' and related terms.

### LINKED ISSUE #1931: Intended meaning of _primary neuron_ is unknown
We have the following terms under _neuron_:

* [primary neuron](http://purl.obolibrary.org/obo/CL_0000530)
  * [primary interneuron](http://purl.obolibrary.org/obo/CL_0000534)
  * [primary motor neuron](http://purl.obolibrary.org/obo/CL_0000533)
  * [primary sensory neuron](http://purl.obolibrary.org/obo/CL_0000531)
* [secondary neuron](http://purl.obolibrary.org/obo/CL_0000535)
  * [secondary motor neuron](http://purl.obolibrary.org/obo/CL_0000535)

Because none of them has a definition, it is unknown what is the intended meaning of _primary_ and _secondary_ here.

In FBbt, _primary neuron_ and _secondary neuron_ have a clear meaning: a primary neuron is a neuron that arises during the first wave of neurogenesis (at the embryonic stage), while a secondary neuron is a neuron that arises during the second wave of neurogenesis (at the larval stage).

The concept of “two-stages neurogenesis” is not specific to flies, so it makes sense for CL to have terms referring to neurons born during the “first neurogenesis” and neurons born during the “second neurogenesis” — _if_ that is indeed the meaning that was intended when the above terms were added in CL.

[primary motor neuron](http://purl.obolibrary.org/obo/CL_0000533) is a parent to [CAP motoneuron](http://purl.obolibrary.org/obo/CL_0000532), which again has no definition but is probably the [CaP motoneuron](https://zfin.org/action/ontology/term-detail/ZFA:0009243) of zebrafish (the FZA term has a cross-reference back to the CL term). In ZFA, [primary motor neurons](https://zfin.org/action/ontology/term-detail/ZFA:0009244) are neurons that arise during the first day of development, while [secondary motor neurons](https://zfin.org/action/ontology/term-detail/ZFA:0009247) are neurons that “develop several hours later than primary motor neurons”, which at first sight seems compatible with the general idea that “primary“ and “secondary” refer to the _time_ at which the neurons appear.

On the other hand, _primary motor neuron_ could also easily be understood (especially by people with a “vertebrate mindset”) as “a neuron of the primary motor cortex”, and likewise _secondary motor neuron_ could be “a neuron of the secondary motor areas”. Of note, both _primary motor neuron_ and _secondary motor neuron_ are mapped to equivalent FMA terms ([FMA:83619](http://purl.obolibrary.org/obo/FMA_83619) and [FMA:83620](http://purl.obolibrary.org/obo/FMA_83619), respectively), but since the FMA terms _also_ have no definitions (😩), that does not help a bit…

So:

Does anyone know or remember what was the intended meaning of _primary_ and _secondary_ in the label of the terms above? Was it about timing (where “primary” ≈ “early” and “secondary” ≈ “late”), as in flies and zebrafish?

gouttegd:
Obsolete 'primary neuron' and its generic subclasses 'primary motor neuron', 'primary sensory neuron', and 'primary interneuron'. Subclasses corresponding to precise neurons ('CAP motoneuron' and 'Rohon-Beard neuron') are re-classified so that they do not depend on those terms.

Likewise for 'secondary neuron'.

closes #1931

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -5683,7 +5683,6 @@ AnnotationAssertion(Annotation(oboInOwl:hasDbXref "VHOG:0001484") oboInOwl:hasEx
 AnnotationAssertion(oboInOwl:hasExactSynonym obo:CL_0000247 "Rohon-Beard cell")
 AnnotationAssertion(Annotation(oboInOwl:hasDbXref "VHOG:0001484") oboInOwl:hasRelatedSynonym obo:CL_0000247 "RB neuron")
 AnnotationAssertion(rdfs:label obo:CL_0000247 "Rohon-Beard neuron")
-SubClassOf(obo:CL_0000247 obo:CL_0000531)
 SubClassOf(obo:CL_0000247 obo:CL_4023168)
 
 # Class: obo:CL_0000248 (obsolete microsporocyte)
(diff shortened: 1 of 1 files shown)


gouttegd:
obophenotype/uberon#2869 needs to be fixed before this PR can be merged.

cerivs:
These terms are in use by the ZFA.  Before obsoleting terms in use they should be discussed with the entire community.

src/ontology/cl-edit.owl
cerivs, line 1: These terms are in use and should not be obsoleted without community feedback

gouttegd:
But isn’t ZFA using its own terms (ZFA:0009242, ZFA:0009244, etc)? Why would ZFA need those terms in CL?

gouttegd:
In any case, the issue _has_ been discussed in the monthly CL meeting, which AFAIK is open to the entire community, and the consensus was for obsoletion.

I don’t mind putting this issue on hold for the time being, but we won’t keep undefined terms in CL for ever.

cerivs:
The ZFA terms are the zebrafish instance of the CL term.   The first mention on Github of any problem with the term was only last week.  The meeting was one week later.  CL was originally unifying the cell components of the various species specific ontologies similar to how UBERON works for gross anatomy.    If CL is acting as the source for all cells then sufficient time should be given to allow the community to review proposals and come up with definitions.     One week notice is not sufficient for terms to be made obsolete.

gouttegd:
> The ZFA terms are the zebrafish instance of the CL term.

I don’t understand what you mean here. The ZFA terms are _mapped_ to the CL terms (on the ZFA side), that does not make them _instances_ of them. The ZFA terms can still exist, in their current position in the ZFA hierarchy, independently of the CL terms.

Exactly as the similar FBbt terms will continue to exist in the FlyBase ontology.

> If CL is acting as the source for all cells

Precisely, CL is _not_ acting as the source for all cells. Maybe it has started as such, but it has been agreed for quite some time (at least since I started contributing, two years ago) that for some not-generic-enough cell types, CL should delegate to other, more appropriate ontologies, such as FBbt or ZFA.

bvarner-ebi:
Hi, @cerivs, @gouttegd, are there still pending concerns for this PR?
If no, can you kindly merge?
If yes, can you detail the pending concerns and add them to the next CL meeting agenda if necessary?

gouttegd:
@cerivs There will be a CL meeting this Wednesday at 8am PDT (UTC-7) (see agenda in the top-level README). If you want to make a case that [primary neuron](http://purl.obolibrary.org/obo/CL_0000530) and related terms should _not_ be obsoleted, I suggest you either attend this meeting or expose your case in a comment below (I’ll attend the meeting and convey your points to the other editors).

Otherwise, I’ll recommend going forward with the obsoletion.

dosumis:
In discussion in CL meeting @cerivs made clear that this term has its origins in TAO - which was merged into CL.  It is therefore likely that there are users of this term who would be affected by an obsoletion.  To support these users, we could more narrowly define  it - and make any taxon restriction clear in the name and axiomatization.  Potential sources for definition as fish specific term https://docs.google.com/document/d/1XvMbNvr0FEsdqGhg79BYCYEHSqUxRHMcvhbGizEAht8/edit

image

gouttegd:
Superseded by #1984 (the terms are to be made specific to teleosts, rather than obsoleted).

Issue #1961, opened 2023-04-30
'taste receptor cell' is loosely defined but tightly classified


gouttegd:
(This is another instance of the same type of problem highlighted in #1960.)

[Taste receptor cell](http://purl.obolibrary.org/obo/CL_0000209) is logically defined as a [neuron](http://purl.obolibrary.org/obo/CL_0000540) that is [capable of](http://purl.obolibrary.org/obo/RO_0002215) some [detection of chemical stimulus involved in sensory perception of taste](http://purl.obolibrary.org/obo/GO_0050912). That definition is not taxon-restricted and will therefore catch any such neuron in any species. This includes, for example, the gustatory receptor neurons of _Drosophila melanogaster_, which are similarly defined by reference to GO:0050912.

But [taste receptor cell](http://purl.obolibrary.org/obo/CL_0000209) is _also_ classified as being [part of](http://purl.obolibrary.org/obo/BFO_0000050) some [stratum basale of epidermis](http://purl.obolibrary.org/obo/UBERON_0002025), which is ultimately (through a chain of relationships going back to Uberon’s [neural crest](http://purl.obolibrary.org/obo/UBERON_0002342)) taxon-restricted to chordates.

This causes ~200 gustatory receptors neurons from FBbt to be unsatisfiable when merging Uberon, CL, and FBbt.

As in #1960, we should decide whether [taste receptor cell](http://purl.obolibrary.org/obo/CL_0000209) should be taxon-neutral, in which case it should not be linked to the epidermis (beyond chordates, taste receptor cells may be found elsewhere than in the epidermis), or whether it is really intended to be chordate-specific, in which case the logical definition should be tightened.

I favour the first solution, for consistency with the neighbouring terms [olfactory epithelial cell](http://purl.obolibrary.org/obo/CL_0002167) and [auditory epithelial cell](http://purl.obolibrary.org/obo/CL_0002491), which are defined similarly to [taste receptor cell](http://purl.obolibrary.org/obo/CL_0000209) but do not have an additional, taxon-restricted classification.

Of note, if we opt for the first solution of making [taste receptor cell](http://purl.obolibrary.org/obo/CL_0000209) a true taxon-neutral term, its _text definition_ should also be amended, as it currently says that those cells are “mainly found in the epithelium of the tongue”, which is very much _not_ taxon-neutral.

dosumis:
[...]

PR #1963, opened 2023-05-02
Make 'taste receptor cell' taxon-neutral

gouttegd:
This PR implements the first solution discussed in #1961: it makes [taste receptor cell](http://purl.obolibrary.org/obo/CL_0000209) taxon-neutral by removing the reference to the [stratum basale of epidermis](http://purl.obolibrary.org/obo/UBERON_0002025).


aleixpuigb (approved):
I agree with the changes. Since the majority of fish do not have a tongue, I was wondering if substituting 'vertebrates' with 'tetrapods' in the definition was more precise. However, there are some aquatic species that have tongue, so I suppose it is ok to leave 'vertebrates'.

ybradford:
@gouttegd zebrafish, a vertebrate, has taste receptors on more structures than just the tongue. https://rdcu.be/dbbA9 'The taste buds allow fish to discriminate between palatable and noxious food items, and are found throughout the oropharyngeal epithelium, on the branchial arches, and on the head and other regions of the body in some species[7](https://www.nature.com/articles/s41598-017-14042-3#ref-CR7)

Issue #114, opened 2015-06-04
new term proposal catecholaminergic neuron


GoogleCodeExporter:
```

Need term catecholaminergic neuron
Def: Neurons that produce dopamine, norephinephrine or adrenaline.  CA 
synthesis involves four enzymes: tyrosine hydroxylase (TH), aromatic amino acid 
decarboxylase (AADC), dopamine b-hydroxylase (DbH), and 
phenylethanolamine-N-methyl transferase (PNMT) PMID:10191060 ISBN:
9780521442510     

Notes: Frequently used when describing the developing nervous system in 
zebrafish.  Cells express th or th2 genes.  
PMID: 12128258   

```

Original issue reported on code.google.com by `cer...@gmail.com` on 20 Mar 2014 at 5:54

addiehl:
Seems like a valid term.  Is this still needed @cerivs?

bvarner-ebi:
Some comments:

The proposed text def is:
"Neurons that produce dopamine, norephinephrine or adrenaline.  CA 
synthesis involves four enzymes: tyrosine hydroxylase (TH), aromatic amino acid 
decarboxylase (AADC), dopamine b-hydroxylase (DbH), and 
phenylethanolamine-N-methyl transferase (PNMT)"

- "Neurons" should be in singular form.

- The term epinephrine vs adrenaline should be used consistently:

A neuron that produces dopamine, norephinephrine or epinephrine.
OR
A neuron that produces dopamine, noradrenaline or adrenaline.

- The gloss may be too broad for the term 'catecholaminergic neuron'.

- The following GO terms already exist:

 GO:0061527 'dopamine secretion, neurotransmission'
 GO:0061533 'norepinephrine secretion, neurotransmission'
 GO:0061529 'epinephrine secretion, neurotransmission'
 
 Given that, should a new term 'catecholamine secretion, neurotransmission' be created in GO?
 
 FYI @shawntanzk

shawntanzk:
Ideally having a new term 'catecholamine secretion, neurotransmission' that has all the catecholamine secretion under it would be ideal. I'll create a ticket there. However, if there is any issue, we can also manually assert given its only 3. Give this ticket is from 2015 I don't think there is any rush. Will wait to see what GO says about the new term

anitacaron:
[...]

PR #1981, opened 2023-05-18
Addresses #114 NTR catecholaminergic neuron

aleixpuigb:
Addresses #114 NTR catecholaminergic neuron


aleixpuigb:
#gogoeditdiff

aleixpuigb:
It is strange that only `noradrenergic neuron` is inferred as a subclass of this term.

image

In GO this classes are inferred as SubclassesOf `catecholamine secretion, neurotreansmission`:

image

However, this is not reflected in CL:

image

balhoff:
@matentzn probably a justification for "base-plus".

dosumis (approved):

aleixpuigb:
> @aleixpuigb I have a guess about this. GO gets this subsumption using some behind-the-scenes GCIs for CHEBI that aren't included in our base file. If CL is building on the GO base file, some necessary axioms might be missing.

Thank you, @balhoff! That makes it clear.

dosumis:
@balhoff - my first thought was that if base is build from go-plus it should have all the necessary axioms - and if not, this is a misfeature of the GO build.  But, thinking about it, there is a ticket somewhere about derivation of base and the problems of retaining GCIs.  I'd like to look again at how we might solve that problem as I increasingly have use cases for GCIs.  CC @matentzn

balhoff:
@dosumis you are right, I would call it a misfeature of go-base. I'm not sure that we should spread these CHEBI GCIs out into other ontologies (I'm referring to the "biochebi" stuff). But we need to include the reasoning results into go-base if don't don't include those GCIs.

PR #2003, opened 2023-06-02
fixes #2001 NTR: pre-type II pneumocyte

### LINKED ISSUE #2001: pre-type 2 pneumocyte
**Preferred term label:** pre-type 2 pneumocyte

**Synonyms**
pre-type II alveolar epithelial cell PMID:12114192
alveolar pre-type II epithelial cell PMID:9624168

**Definition**
Precursor of type II pneumocyte that does not yet produce surfactant. These cells do not have lamellar bodies, which are a marker of type II pneumocyte maturity. PMID:1316350
Curator note: A central feature of the fetal respiratory distress syndrome seems to be the presence in the lung of abundant pre-type II alveolar epithelial cells. PMID:1316350

**Parent term**
pneumocyte CL:0000322

**ORCID**
https://orcid.org/0000-0002-0819-0473

paulacalipho:
fixes #2001


paulacalipho:
Dear @JoshuaFortriede, I would appreciate if you could give me a feedback on this PR. Many thanks, Paula

paulacalipho:
Thanks @JoshuaFortriede ,seems reasonable to me. I wait for @bvarner-ebi input to add these changes

bvarner-ebi:
> Thanks @JoshuaFortriede ,seems reasonable to me. I wait for @bvarner-ebi input to add these changes

Sure, these edits seems reasonable.

For reference, this is the definition of 'develops from':
"x develops from y if and only if either (a) x directly develops from y or (b) there exists some z such that x directly develops from z and z develops from y"

paulacalipho:
I can not add: pneumocyte and lacks_part some 'lamellar body' neither as SubClass of nor as Equivalent to. 'lamellar body' is not recognized

dosumis:
I'd rather avoid use lacks_part - it's not really safe for inference.  If this is important information for biologists to identify these cells in images, it can be added in the text definition. If we have cell types for which we want to add a positive assertion of the presence of a lamellar body, we can add a has_part relationship to GO [lamellar body](https://www.ebi.ac.uk/ols4/ontologies/go/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FGO_0042599) to these.  If this is not yet available in CL, you can just add a new term in Protege, pasting in the GO URL and leaqvinf everything else blank.  You can then refer to it in axioms by its ID (GO_004259).   A later round of imports will pull the term from GO before CL release.

paulacalipho:
OK, I will not add 'lacks_part' lamellar body

JoshuaFortriede:
Your comments make sense. Regarding the "lacks_part" not being safe for inference, is that because its a negative property? I ask because I believe more and more, specific cell types are described as a subclass of a cell type but lacking some key property. I understand this would cause problems with many queries that are just finding relationships between cells. However, specific negative associations can be very important. More so when more and more cells are found in different tissues with the same properties.

bvarner-ebi (approved):

Issue #2004, opened 2023-06-02
[NTR] dorsal striatum medium spiny neurons subtypes


bvarner-ebi:
**Preferred term label**
matrix D1 medium spiny neuron

**Synonyms**
matrix D1-MSN
D1M
D1-matrix
(related, abbr, DOI:10.1016/j.cub.2021.10.015)

**Definition**
A DRD1-expressing medium spiny neuron that is part of a matrix compartment of the dorsal striatum.
(DOI:10.1016/j.cub.2021.10.015)

**Parent cell type term**
CL:4023026 'direct pathway medium spiny neuron'

**What is the anatomical structure that the cell is a part of?** 
UBERON:0027368 'matrix compartment' and UBERON:0005382 'dorsal striatum'

**Additional notes or concerns**
Comment: In Rhesus macaques, both matrix D1 medium spiny neuron (MSN) and matrix D2 MSN cell types have been noted to have enriched gene expression of the following: _EPHA4_, _GDA_, _STXBP6_, and _SEMA3E_.
(DOI:10.1016/j.cub.2021.10.015)

---

**Preferred term label**
matrix D2 medium spiny neuron

**Synonyms**
matrix D2-MSN
D2M
D2-matrix
(related, abbr, DOI:10.1016/j.cub.2021.10.015)

**Definition**
A DRD2-expressing medium spiny neuron that is part of a matrix compartment of the dorsal striatum. 
(DOI:10.1016/j.cub.2021.10.015)

**Parent cell type term**
CL:4023029 'indirect pathway medium spiny neuron'

**What is the anatomical structure that the cell is a part of?** 
UBERON:0027368 'matrix compartment' and UBERON:0005382 'dorsal striatum'

**Additional notes or concerns**
Comment: In Rhesus macaques, both matrix D1 medium spiny neuron (MSN) and matrix D2 MSN cell types have been noted to have enriched gene expression of the following: _EPHA4_, _GDA_, _STXBP6_, and _SEMA3E_.
(DOI:10.1016/j.cub.2021.10.015)

---

**Preferred term label**
striosomal D1 medium spiny neuron

**Synonyms**
D1S
D1-striosome
striosome D1-MSN 
(related, abbr, DOI:10.1016/j.cub.2021.10.015)

**Definition**
A DRD1-expressing medium spiny neuron that is part of a striosome of the dorsal striatum. 
(DOI:10.1016/j.cub.2021.10.015)

**Parent cell type term**
[...]

PR #2005, opened 2023-06-02
Addresses #2004 NTR dorsal striatum medium spiny neuron subtypes

bvarner-ebi:
Addresses #2004 NTR dorsal striatum medium spiny neuron subtypes


dosumis:
#gogoeditdiff

dosumis:
image

Re use or 'part of' are the neurons entirely within the compartment(s) mentioned?
Also - I suspect that the semantics of `('dorsal striatum' and 'matrix compartment')` are not what you intend.  Is anything an instance of both 'dorsal striatum' and 'matrix compartment' ?

dosumis:
Also looks like parent classes could do with some work. e.g. - medium spiny neuron should be GABA-ergic.  SHould probably also have location recorded using has_soma_location rather than part_of given that these are projection neurons.

bvarner-ebi:
> Re use or 'part of': are the neurons entirely within the compartment(s) mentioned?

Thank you for the feedback. Will change to 'has soma location'.

> Also looks like parent classes could do with some work. e.g. - medium spiny neuron should be GABA-ergic. SHould probably also have location recorded using has_soma_location rather than part_of given that these are projection neurons.

Will open a separate ticket to address.

> Also - I suspect that the semantics of ('dorsal striatum' and 'matrix compartment') are not what you intend. Is anything an instance of both 'dorsal striatum' and 'matrix compartment' ?

The reference paper for these new cell type terms refers to MSNs found in the matrix compartments that are part of the DS. My understanding is that these matrix compartments are considered part of the striatum, so I believe the answer is yes.

> Do we need an Uberon term for "matrix compartment of dorsal striatum"?

Is it preferred to add another term to Uberon or simply assert this cell type is found in matrix compartments that are also part of the DS?

dosumis:
> Is it preferred to add another term to Uberon or simply assert this cell type is found in matrix compartments that are also part of the DS?

Strictly, we're not asserting this anywhere right now. It's a bit hard to tell from the definitions, but based on usage 'matrix compartment' seems to be a general type with instances by region, whereas the term 'dorsal striatum' is use to refer to the whole region.  e.g. 

[matrix compartment of caudate nucleus](https://www.ebi.ac.uk/ols4/ontologies/uberon/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FUBERON_0029001?lang=en)  Equivalent to
([matrix compartment](https://www.ebi.ac.uk/ols4/ontologies/uberon/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FUBERON_0027368?lang=en) and [part of](https://www.ebi.ac.uk/ols4/ontologies/uberon/properties/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FBFO_0000050?lang=en) [RO](https://www.ebi.ac.uk/ols4/ontologies/ro/properties/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FBFO_0000050) some [caudate nucleus](https://www.ebi.ac.uk/ols4/ontologies/uberon/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FUBERON_0001873?lang=en))

So the correct anonymous class expression for the region we need to refer to here would be 

"'matrix compartment' and part of some 'dorsal striatum'", not "'matrix compartment' and 'dorsal striatum'".  If we add this as a class to Uberon, the reasoner will make it a superclass of 'matrix compartment of caudate nucleus' based on the caudate nucleus being 'part of' the dorsal striatum.

I think best to add to Uberon, but happy to use an anonymous class expression in CL for now.

dosumis:
#gogoeditdiff

aleixpuigb (approved):
Everything looks fine.
My only comment is about the spelling of `extra-striosomal`. In the [dbxref](https://doi.org/10.1016/j.cub.2021.10.015) and [other articles](https://doi.org/10.1016/0304-3940(89)90048-7), it is used as `extrastriosomal`, but this might be one of the English (UK) and English (USA) differences in the spelling.

PR #2010, opened 2023-06-07
docs: Add some guidelines for term editors.

gouttegd:
This PR adds a new document to provide general guidelines for editors on how to name and define (textually and formally) terms in the ontology.

Those guidelines are mostly based on available similar guidelines in FBbt.


docs/editing_guidelines.md
```diff
@@ -0,0 +1,138 @@
+# Guidelines for CL editors
+
+## Naming terms
+
+### General rules
+1. All names should be singular nouns.
+2. Names should not be capitalised, unless they are or contain a proper name (e.g. _Johnston's organ_).
```
bvarner-ebi, line 7:
```suggestion
2. Names should not be capitalised unless they are or contain a proper name (e.g., _Johnston's organ_).
```

docs/editing_guidelines.md
```diff
@@ -0,0 +1,138 @@
...
+3. Avoid special characters: use only alpha-numeric characters, space, dash (`-`), slash (`/`), and apostrophe (`'`).
```
bvarner-ebi, line 8:
```suggestion
3. Avoid special characters: use only alphanumeric characters, space, dash (`-`), slash (`/`), and apostrophe (`'`).
```

docs/editing_guidelines.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Advice
+Bear in mind that users will often encounter terms in isolation. Long descriptive names (within reason!) are therefore preferable, especially where there is obvious potential for confusion. For example, _calyx_ (which simply means “cup”) could refer to a structure in the oviduct or in the mushroom body, depending on your field of specialisation. It is therefore better to use _mushroom body calyx_ and _oviduct calyx_ rather than simply _calyx_ alone.
```
bvarner-ebi, line 11:
```suggestion
Bear in mind that users will often encounter terms in isolation. Long, descriptive labels (within reason!) are therefore preferable, especially where there is obvious potential for confusion. For example, _calyx_ (which simply means “cup”) could refer to a structure in the oviduct or in the mushroom body, depending on your field of specialization. It is therefore better to use _mushroom body calyx_ or _oviduct calyx_ rather than simply _calyx_ alone.
```

docs/editing_guidelines.md
```diff
@@ -0,0 +1,138 @@
...
+
+When crafting a definition, editors should aim for a reasonably succinct statement about the class allowing curators and users to easily distinguish it from other, similar classes, and which captures key points of interest about that class. It should capture assertions made in the formal part of the definition (the relationships) as closely as possible without becoming stilted and difficult to read.
+
+The basic structure of a definition should be as follows:
+
+> A `genus` that `diff1` and `diff2`. It also `diff3` and `gloss`...
+
+where `genus` is a general classification and `diff1`, `diff2`, etc. are the _differentia_, which state what differentiates this class from others that share the same general classification. The gloss, when present, gives some key points of interest about the class.
+
+The first sentence of the definition should be in singular form. The remaining of the definition may then enter into plural if needed.
```
bvarner-ebi, line 27:
```suggestion
The first sentence of the definition should refer to the definiendum in singular form. The rest of the definition may then invoke the plural form.
```

docs/editing_guidelines.md
```diff
@@ -0,0 +1,138 @@
...
+It is difficult to specify, _a priori_, which assertions should be included in a textual definition. However there are some general guidelines.
+
+1. DO make sure your definition is consistent with the definition of the superclass(es).
+2. DO make sure your definition includes the information that is recorded in all the direct formal relationships to the class.
+3. AVOID assertions about structures or cell types that are not part of the cell type being described, except when they pertain to some direct relationships with the cell type being described.
+4. AVOID including details that could better be included in the definition of subtypes of the cell type being described.
+5. LIMIT information that applies to only some members of the class. This should only be used sparingly; when used, it should be made clear that it does not apply to all members of the class.
+6. AVOID using gene expression as the differentium.
+7. AVOID extensive repetition of assertions made in superclass definitions, unless these assertions are used to provide direct evidence for class membership.
+8. DO NOT add informations about what happens in mutant backgrounds or other kind of non-control conditions.
```
bvarner-ebi, line 39:
```suggestion
8. DO NOT add information about what happens in pathological states or other kind of non-control conditions.
```

docs/editing_guidelines.md
```diff
@@ -0,0 +1,138 @@
...
+9. DO NOT include reasons for believing the assertions to be true. These should be recorded in comments.
+10. DO NOT raise questions in the definition. The definition should have the sense of being definitive, which is undermined if we show doubt. Any doubt should be recorded in comments instead.
+
+
+## Comments
+Comments should be used for:
+
+1. Providing evidence. In some cases it is useful to know the type of evidence for an assertion. This should not be recorded in the definition, but _can_ be recorded in a comment.
+2. Disambiguation. Sometimes a single term is used in the literature with multiple meanings. In such cases, acomment should be added outlining these different uses and how they relate to the standard definition set in the ontology.
```
bvarner-ebi, line 48:
```suggestion
2. Disambiguation. Sometimes a single term is used in the literature with multiple meanings. In such cases, a comment should be added outlining these different uses and how they relate to the definition set in the ontology.
```

docs/editing_guidelines.md
```diff
@@ -0,0 +1,138 @@
...
+3. Reporting editorial decisions (or decisions in waiting) about the term. This includes providing a reason for obsoleting a term, or letting users and curators know that the term may be merged or split in the future, e.g. when enough evidence for the merge/split will be available.
```
bvarner-ebi, line 49:
```suggestion
3. Reporting editorial decisions (or decisions in waiting) about the term. This includes providing a reason for obsoleting a term, or letting users and curators know that the term may be merged or split in the future, e.g., when enough evidence for the merge/split will be available.
```

docs/editing_guidelines.md
```diff
@@ -0,0 +1,138 @@
...
+
+Try to be consistent in how you phrase the various types of comments. For example:
+
+* when giving a reason for obsoletion, use “Obsoleted as ...”;
+* when indicating a potential future merge, use “Possible equivalence with {other term} ...”.
+
+
+## Synonyms
+Extensive addition of synonyms help searching. Add as many synonyms as you like.
```
bvarner-ebi, line 58:
```suggestion
Extensive addition of synonyms helps "findability" of terms when searching. Accurate synonyms can and should be added liberally.
```

docs/editing_guidelines.md
```diff
@@ -0,0 +1,138 @@
...
+
+Guidelines on the type of synonyms:
+
+1. Use an _exact_ synonym when the name and the synonym can be used interchangeably, at least in a given context. This includes the case where the synonym is an abbreviation that, in a broader context, could have a different meaning. For example, _DPM_  can be an exact synonym of _dorsal pharyngeal muscle cell_, even though “DPM” could have many more meanings generally, because in the context where “dorsal pharyngeal muscle cell” is used, “DPM” is commonly used interchangeably to refer to these cells.
+2. There is no expectation that exact synonyms should be unique across the entire ontology. For example, in the _Drosophila_ Anatomy Ontology _DPM_ is also an exact synonym of _mushroom body dorsal paired medial cell_.
```
bvarner-ebi, line 63:
```suggestion
2. Exact synonyms should be unique across the ontology. In other words, if class _A_ has exact synonym "_X_", "_X_" should not be an exact synonym for any other CL term.
```

docs/editing_guidelines.md
```diff
@@ -0,0 +1,138 @@
...
+3. Be mindful of the “directionality” of the _narrow_ and _broad_ types of synonyms. They qualify the _synonym_, not the original term. For example, saying that _dmd5_ is a narrow synonym of _abdominal dorsal multidendritic neuron_ means that “dmd5“ refers to a narrower concept than _abdominal dorsal multidendritic neuron_, not the other way around.
```
bvarner-ebi, line 64:
```suggestion
3. Be mindful of the “directionality” of the _narrow_ and _broad_ types of synonyms. They qualify the _synonym_, not the original term. For example, saying that "_peripheral blood mononuclear cell_" is a narrow synonym of '_mononuclear cell_' means that "_peripheral blood mononuclear cell_" refers to a narrower concept than '_mononuclear cell_', not the other way around.
```

docs/editing_guidelines.md
```diff
@@ -0,0 +1,138 @@
...
+4. The _related_ type may be used for other cases where the overlap between the synonym and the term name may be unclear, but you want people to find the term when they are searching for the synonym.
```
bvarner-ebi, line 65:
```suggestion
4. The _related_ synonym type should be used for cases where the overlap between the synonym and the term label may be unclear, disputable or not true in all scenarios or contexts, but you want the term to be findable when searching. This includes abbreviations, which should be annotated as _related_ synonyms with synonym type 'abbreviation' (see point 1 above). 
If a synonym includes a mix of abbreviations and words, the _related_ type should still be used unless there is enough context within the synonym itself to make it clear that the synonym refers only to the concept being annotated. For example, "_lung TRM CD8-positive, CD103-positive cell_" would be an _exact_ synonym of '_lung resident memory CD8-positive, CD103-positive, alpha-beta T cell_' even though TRM is an abbreviation for "tissue resident T cell". Compare to "_IMB cell_", which should be a _related_ synonym of '_invaginating midget bipolar cell_'.
```

docs/editing_guidelines.md
```diff
@@ -0,0 +1,138 @@
...
+
+
+## Considerations on style
+The following considerations apply both to all human-readable fields (names, textual definitions, comments, synonyms).
+
+* **US english or British english?** Where there are differences in the accepted spelling between english and US usage, use the US form. British english variants of the labels may be added as synonyms (e.g., a term labelled “epithelial cell of esophagus” may have “epithelial cell of oesophagus” as an exact synonym).
+* **Use of jargon.** The aim of the ontology is to provide useable descriptions and links to the reader. Consequently, try to avoid obscure jargon or pretentious latin/greek, especially where widely understood, plain alternatives exist. Where it will aid searching, such terms may be added to synonyms and/or as asides in the “gloss” part of definitions.
```
bvarner-ebi, line 72:
```suggestion
* **Use of jargon.** The aim of the ontology is to provide useable descriptions and links to the reader. Consequently, try to avoid obscure jargon or pretentious Latin/Greek, especially where widely understood, plain alternatives exist. Where it will aid searching, such terms may be added to synonyms and/or as asides in the “gloss” part of definitions.
```

docs/editing_guidelines.md
```diff
@@ -0,0 +1,138 @@
...
+* **That or which?** These are so interchangead that there isn’t really a rule anymore. But as a guideline, use “which” after a comma (e.g., “this study, which cost $10,000, was a success”), and “that” when no comma is used (e.g., “the study that cost $10,000 was a success”).
```
bvarner-ebi, line 73:
```suggestion
* **That or which?** These are so interchangeable that there isn’t really a rule anymore. But as a guideline, use “which” after a comma (e.g., “this study, which cost $10,000, was a success”), and “that” when no comma is used (e.g., “the study that cost $10,000 was a success”).
```

docs/editing_guidelines.md
```diff
@@ -0,0 +1,138 @@
...
+* **Use of hyphens.** Yes to those that help clarity (e.g., “posterior-most”) and those that are regularly used/accepted in the literature. Generally, no hyphens after prefixes such as “sub”, “mid”, “semi”, “hemi”, etc. (e.g., “hemidesmosome” instead of “hemi-desmosome”), unless it helps with clarity (e.g. “multi-innervated”). No hyphens for composed location adjectives (e.g., “posteroanterior”), unless there are more than two compounds (e.g., “ventro-posterolateral”).
```
bvarner-ebi, line 74:
```suggestion
* **Use of hyphens.** Yes to those that help clarity (e.g., “posterior-most”) and those that are regularly used/accepted in the literature. Generally, no hyphens after prefixes such as “sub”, “mid”, “semi”, “hemi”, etc. (e.g., “hemidesmosome” instead of “hemi-desmosome”), unless it helps with clarity (e.g., “multi-innervated”). No hyphens for composed location adjectives (e.g., “posteroanterior”), unless there are more than two compounds (e.g., “ventro-posterolateral”).
```

docs/editing_guidelines.md
```diff
@@ -0,0 +1,138 @@
...
+> Class: 'crystal cell'
+>   SubClassOf: 'hemocyte'
+>   SubClassOf: 'develops from' some 'procrystal cell'
+> ```
+>
+> The first `SubClassOf` axiom denotes an actual classification, whereas the second denotes a relationship.
+
+
+### Asserting classification
+In order to keep the ontology maintainable, asserted classifications should be limited where possible. Ideally, all terms would have only a single asserted parentt. Given the presence of suitable logically defined classes (see below) and sufficient relationships for the term you are making, additional classification can be inferred automatically by reasoning.
```
bvarner-ebi, line 109:
```suggestion
In order to keep the ontology maintainable, asserted classifications should be limited where possible. Ideally, all terms would have only a single asserted parent (also known as "superclass"). Given the presence of suitable logically defined classes (see below) and sufficient relationships for the term you are making, additional classification can be inferred automatically by reasoning.
```

docs/editing_guidelines.md
gouttegd: This requires further discussion.

For what it’s worth, I disagree.

1. Such a policy, if we indeed want to have it, should be enforced in CI checks, not merely stated in guidelines.
2. This would mean that synonyms that may have several meanings in a broader context would all end up as being _related_ synonyms. That category is already the “catch-all” category for all synonyms that are neither exact, narrow, or related, making it the least meaningful category. Such a policy would dilute its meaning even more.
3. This makes even less sense when you consider that ultimately, CL is not a standalone ontology but a part of a much larger set. Enforcing uniqueness of exact synonyms within CL is pointless if we cannot enforce uniqueness OBO-wide.

gouttegd:
@bvarner-ebi Thank you for your extensive review. However, I would prefer if for now the reviews were focused on the _contents_ of the guidelines.

I appreciate your commitment to the rules of typography, but fixing missing commas after `.e.g.` can wait until everybody agree on what the guidelines should be. :)

docs/editing_guidelines.md
```diff
@@ -0,0 +1,138 @@
...
+
+* what the cell is part of;
+* what it innervates (or, if more precisely known, what its axon(s) or dendrite(s) innervate);
+* its function.
+
+Care must be taken not to use logically defined classes too liberally. You should be satisfied that all automatic classifications that result would make sense (or at least be justifiable) to a biologist and would cover most cases of usages with as few edge cases as possible.
+
+If you are unsure whether a characteristic of a cell should be expressed as a relationship or as a differentium in a logically defined class: use a relationship.
+
+The textual definition of a logically defined class should be a plain english equivalent of the logical definition. However, in some cases it can be useful to add more details in the “gloss” part, when relevant.
```
bvarner-ebi, line 133:
```suggestion
The textual definition of a logically defined class should be a plain English equivalent of the logical definition. However, in some cases it can be useful to add more details in the “gloss” part when relevant.
```

bvarner-ebi (changes requested):
Thank you for preparing this, @gouttegd.
Kindly see inline suggestions.

docs/editing_guidelines.md
bvarner-ebi: This was discussed and reflected [here](https://github.com/obophenotype/uberon/issues/2363#issuecomment-1446928901). 

>This would mean that synonyms that may have several meanings in a broader context would all end up as being related synonyms.

Yes, annotating terms as exact synonyms should be used conservatively and only when they refer exactly to the concept captured in the ontology.

gouttegd:
Thank you. I’ll wait until other reviewers weigh in on the contents of the guidelines (especially regarding the use of exact synonyms, since we disagree on that) before fixing the typos and other minor stuff.

docs/editing_guidelines.md
gouttegd: My bad, I must have missed that particular Uberon call.

I still think it’s a bad idea, but if there is already an agreement on the issue then it’s fine by me.

gouttegd: But point 1 above still stands: if it is the policy of Uberon/CL that exact synonyms should be unique across the ontology, then the policy should be enforced in CI checks. You cannot rely on editors 1) being aware of the policy (hell, I am an editor and I was not aware of it!) and 2) taking the time to manually check for uniqueness at edit time.

docs/editing_guidelines.md
dosumis: This is explicitly a rule about naming.  The change makes is a rule about words.

dosumis: I guess the context helps - but the main aim is to ensure editors avoid the habit of capitalising the first letter of a full name.

docs/editing_guidelines.md
```diff
@@ -0,0 +1,138 @@
+# Guidelines for CL editors
+
+## Naming terms
+
+### General rules
+1. All names should be singular nouns.
```
bvarner-ebi, line 6:
```suggestion
1. All labels should be singular nouns.
```

docs/editing_guidelines.md
gouttegd: @dosumis Well, FBbt and CL already _do_ diverge on that issue. We have a lot of duplicated exact synonyms across the ontology, because we allow synonyms to be “exact in a given context”. And for the record I do not plan to change that in FBbt, whatever CL is doing.

dosumis: For the record, I prefer FBbt's approach but am consistently outvoted in general discussion.  In the end we should judge by some clear statements of use cases.  VFB approach is not great for NLP, but I think is fine for search - users can and should be trusted to do the final disambigutaion.

docs/editing_guidelines.md
addiehl: I agree that "label" or "labels" should be used throughout rather than "name" or "names", particularly as the annotation property used is 'label'.

docs/editing_guidelines.md
```diff
@@ -0,0 +1,140 @@
...
+3. Reporting editorial decisions (or decisions in waiting) about the term. This includes providing a reason for obsoleting a term, or letting users and curators know that the term may be merged or split in the future, e.g., when enough evidence for the merge/split will be available.
+
+Try to be consistent in how you phrase the various types of comments. For example:
+
+* when giving a reason for obsoletion, use “Obsoleted as ...”;
+* when indicating a potential future merge, use “Possible equivalence with {other term} ...”.
+
+
+## Synonyms
+Extensive addition of synonyms help “findability” of terms when search. Synonyms can and should be added liberally.
```
aleixpuigb, line 59:
```suggestion
Extensive addition of synonyms helps “findability” of terms when search. Synonyms can and should be added liberally.
```

gouttegd:
@bvarner-ebi : Since you “requested changes” in your initial review, you need to explicitly approve the PR (to signify that you agree with how your comments had been addressed) before the PR can be merged.

@zoependlington : @dosumis was keen on having your opinion on those guidelines, so could you please have a look before we decide to merge?

docs/editing_guidelines.md
```diff
@@ -0,0 +1,140 @@
...
+4. The _related_ synonym type should be used for cases where the overlap between the synonym and the term label may be uncler, disputable or not true in all scenarios or contexts, but you want the term to be findable when searching. This includes abbreviations, which should be annotated as _related_ synonyms with synonym type “abbreviation” (see point 1 above).
+5. If a synonym includes a mix of abbreviations and words, the _related_ type should still be used unless there is enough context within the synonym itself to make it clear that the synonyms refers only to the concept being annotated. For example, “lung TRM CD8-positive, CD103-positive cell” would be an exact synonym of “lung resident memory CD8-positive, CD103-positive, alpha-beta T cell”, even though “TRM” (in this case) is an abbreviation for “tissue resident T cell”. Note that without this context “TRM” should not be considered an exact synonym for “tissue resident T cell” as “TRM” could also mean “treatment-related mortality”, another OBO ontology concept. Compare the previous example to “IMB cell”, which should be a _related_ synonym of “invaginating midget bipolar cell”.
```
bvarner-ebi, line 67:
```suggestion
5. If a synonym includes a mix of abbreviations and words, the _related_ type should still be used unless there is enough context within the synonym itself to make it clear that the synonym refers only to the concept being annotated. For example, “lung TRM CD8-positive, CD103-positive cell” would be an exact synonym of “lung resident memory CD8-positive, CD103-positive, alpha-beta T cell”, even though “TRM” (in this case) is an abbreviation for “tissue resident T cell”. Note that without this context “TRM” should not be considered an exact synonym for “tissue resident T cell” as “TRM” could also mean “treatment-related mortality”, another OBO ontology concept. Compare the previous example to “IMB cell”, which should be a _related_ synonym of “invaginating midget bipolar cell”.
```

Minor grammar edit.

zoependlington (approved):
This all reads well to me. The only thing I could think is perhaps a note about roman numerals vs using numbers if it applies at all to cell terms (e.g. in disease ontologies type 2 diabetes mellitus vs type II diabetes mellitus), but not essential!

dosumis:
@gouttegd  - I think we're ready to merge.  While I'm with you on the synonym issue, I am consistently outvoted on it for CL, so happy to stick with this policy.  

Do you see this as integrated doc for FBbt and CL in future?  If so, how should we co-ordinate? Would you like to add a comment to indicate that FBbt differs, linking to the FBBt doc?

gouttegd:
@dosumis 

> I think we're ready to merge

Looks like it indeed. I’ll merge now.

> Do you see this as integrated doc for FBbt and CL in future?

I am not sure yet. On one hand it would be nice to have a single source of truth for the guidelines, rather than a duplicated document, but on the other hand I’d like to avoid having the FBbt documentation rely on an external (from FBbt’s point of view) source, especially since the CL policy could be modified at any time by CL editors in a direction that FBbt does not necessarily want to follow (as is already the case for the synonym usage policy).

I think that for now at least, FBbt will keep its own document separately, and I’ll try to make sure the FBbt and CL docs stay “in sync”.

Issue #1975, opened 2023-05-11
[NTR] intermediate cell of urothelium, umbrella cell of urothelium


jenjou:
**Preferred term label**
intermediate cell of urothelium

**Synonyms** (add reference(s), please)
N/A

**Definition** (free text, with reference(s), please. PubMed ID format is PMID:XXXXXX)
A regenerative layer of cells directly above basal cells in the urothelium.

See PMID:36180582
Corresponds to FMA:84155 Intermediate cell of urothelium

**Parent cell type term** (check the hierarchy here https://www.ebi.ac.uk/ols/ontologies/cl)
CL:0000731  urothelial cell

**What is the anatomical structure that the cell is a part of?** (check Uberon: https://www.ebi.ac.uk/ols/ontologies/uberon)
UBERON:0000365 Urothelium

----
**Preferred term label**
umbrella cell of urothelium

**Synonyms** (add reference(s), please)
superficial cell of urothelium, facet cell of urothelium

**Definition** (free text, with reference(s), please. PubMed ID format is PMID:XXXXXX)
Fully differentiated cells on the apical surface of the urothelium that function as a barrier.

See PMID:36180582
Corresponds to FMA:84149 Umbrella cell of urothelium

**Parent cell type term** (check the hierarchy here https://www.ebi.ac.uk/ols/ontologies/cl)
CL:0000731  urothelial cell

**What is the anatomical structure that the cell is a part of?** (check Uberon: https://www.ebi.ac.uk/ols/ontologies/uberon)
UBERON:0000365 Urothelium

----

**Your ORCID**
https://orcid.org/0000-0003-3065-9183

**Additional notes or concerns**
These two terms form a "set" with the 3rd cell type of the urothelium, CL:1000486 basal cell of urothelium.

bvarner-ebi:
@jenjou, thank you for these new terms.
Below are some proposed updates.
Can you advise if you agree or what should be added/removed?

----

**Preferred term label** 
intermediate cell of urothelium

**Synonyms**
 urothelial intermediate cell
(exact, doi:10.1007/978-3-030-14366-4_8)

**Definition**
A urothelial cell that is part of the regenerative layer(s) of cells directly superficial to basal cells in urothelium. The layer of intermediate cells in the urothelium ranges from one to several layers thick depending on the species with intermediate cells attached to adjacent cell layers and one another via desmosomes.
doi:10.1038/s41385-022-00565-0

**dbxref** FMA:84155

[...]

PR #2024, opened 2023-06-16
Addresses #1975 [NTR] urothelial cell types

bvarner-ebi:
Addresses #1975 [NTR] urothelial cell types


aleixpuigb (approved):
No breaking changes noted.

Issue #2054, opened 2023-07-06
Replace synapsed_by with 'synapsed by'


bvarner-ebi:
The following two object properties appear in CL:

**synapsed_by** (http://purl.obolibrary.org/obo/uberon/core#synapsed_by)
**'synapsed by'** (http://purl.obolibrary.org/obo/RO_002103)

The former has no annotations, characteristics or descriptions and appears in one CL class:
`'cartwheel cell' SubClassOf synapsed_by some 'granule cell'`

However, 'synapsed by' is used in several CL classes and appears to be a fully annotated RO term.

- [x] This issue tracks replacing the one instance of synapsed_by with 'synapsed by'. 

Pending issues to address:
- Remove synapsed_by from CL. The IRI has "/uberon/core#synapsed_by", but I do not see a native synapsed_by in Uberon. Is it coming from somewhere else? Can an editor simply delete it from CL?

anitacaron:
Recently, we replaced the synapsed_by in Uberon with the synapses in RO https://github.com/obophenotype/uberon/issues/2811

bvarner-ebi:
> Recently, we replaced the synapsed_by in Uberon with the synapses in RO [obophenotype/uberon#2811](https://github.com/obophenotype/uberon/issues/2811)

Thank you. In that case, can I simply delete synapsed_by from CL?

anitacaron:
Yes, that's fine.

PR #2055, opened 2023-07-06
Addresses #2054 Replaced synapsed_by with 'synapsed by' in 'cartwheel…

bvarner-ebi:
Addresses #2054 Replaced synapsed_by with 'synapsed by' in 'cartwheel cell'


bvarner-ebi:
#gogoeditdiff

aleixpuigb (approved):

Issue #2013, opened 2023-06-08
Establish standard for CL database_cross_reference and contributor annotations


bvarner-ebi:
In the process of reviewing #1997, inconsistencies were noted in how annotations are recorded.

**Annotation properties to be clarified:**

**database_cross_reference:**
- [x] Should the value of database_cross_references always be recorded as **literal**, regardless if it is a CURIE, URI, URL, ORCID or any other type of string?

For example, 
the DOI database_cross_reference for the text definition of CL:4023072 'brain vascular cell' where it is literal

<img width="554" alt="Screenshot 2023-06-08 at 11 50 53" src="https://github.com/obophenotype/cell-ontology/assets/95643797/c2f8dd92-706f-417f-9505-b4adae978e11">

vs 

the DOI database_cross_reference for the exact synonym "hepatic progenitor cell" in CL:0002196 'hepatic oval stem cell' where it is an IRI

<img width="552" alt="Screenshot 2023-06-08 at 11 50 21" src="https://github.com/obophenotype/cell-ontology/assets/95643797/6fe3c7c3-782c-4c5b-a264-197899a06706">

- [x] Does the decision above apply to BOTH database_cross_reference annotations applied generally to the term and to annotations of annotations (e.g., when a text definition is annotated with a dbxref)?
- [x] Does the decision above apply to BOTH CURIE and URI formats?
- [x] When a reference has both a PMID and doi reference, is either one preferred?

**contributor:**
- [x] Should contributor ORCIDs be recorded as literals or IRIs?

For all of the above:

Is there a standard for casing?
- [x] doi: vs DOI:, pmid: vs PMID:, etc.
 -- [doi should be lowercase](https://www.doi.org/the-identifier/resources/handbook/2_numbering#2.6.1), [PMID should be uppercase](https://pubmed.ncbi.nlm.nih.gov/help/)
- [x] Are there any circumstances where a whole URI is preferred over CURIE or vice versa? For example, https://www.orcid.org/ vs ORCID:

bvarner-ebi:
FYI @matentzn

gouttegd:
Regarding DOIs, my take is that they should be recorded as `doi:xx.yyyy/abc`, either as a literal string (which is what most people do) or as an IRI (which strictly speaking is more correct).

[...]

PR #2058, opened 2023-07-07
Update editor guidelines re: standardising database_cross_reference and contributor annotations

bvarner-ebi:
Context: #2013

### DIFF
--- a/docs/editing_guidelines.md
+++ b/docs/editing_guidelines.md
@@ -103,7 +103,7 @@ Citations are made by cross-references, that is by adding `http://www.geneontolo
 
 If the main source for an assertion is a term in another ontology, the short identifier for that term may be used as a cross-reference. For example, `WBbt:0006799` to cross-reference a term in the _C. elegans_ Gross Anatomy Ontology.
 
-DO NOT use URLs as database cross-references. They should be entered using `rdfs:seeAlso` (see below).
+DO NOT enter URLs as database cross-references. They should be entered using `rdfs:seeAlso` (see below).
 
 ORCID identifiers may also be used when the only available source for an assertion is an individual researcher. However, this should be AVOIDED.
 
--- a/docs/editing_guidelines.md
+++ b/docs/editing_guidelines.md
@@ -98,7 +98,7 @@ If the cross-reference can be represented as a CURIE and is not an ORCID: Enter 
 
 Non-CURIE values (e.g., URLs) are discouraged, but when used are also entered the same way, i.e., as a literal string with no Datatype selected.
 
-In cases where more than one CURIE is available for a resource, e.g., both a PMID and a doi are available, there is no preference for one over the other.
+In cases where more than one CURIE is available for a resource, either is acceptable, but using the more semantically specific identifier is recommended. For example, when both a PMID and a doi are available for a resource, using the PMID is recommended since it indicates the cross-reference points to a paper, as opposed to a doi which could point to any digital object.
 
 If the cross-reference is an ORCID, enter the ORCID as an IRI in the IRI field on the “IRI Editor” tab.
 
--- a/docs/editing_guidelines.md
+++ b/docs/editing_guidelines.md
@@ -102,7 +102,7 @@ In cases where more than one CURIE is available for a resource, either is accept
 
 If the cross-reference is an ORCID, enter the ORCID as an IRI in the IRI field on the "IRI Editor" tab, for example `https://orcid.org/0000-0002-7356-1779`.
 
-To restate, in all cases except ORCIDs, the CURIE is preferred and entered as a literal string. For ORCIDs, the IRI is preferred and is entered as an IRI.
+To restate, in all cases except ORCIDs, the CURIE is preferred and entered as a literal string. ORCIDs MUST be entered as an IRI.
 
 The above instructions apply whether the cross-reference is added to another annotation (e.g., annotating a text defintion or comment) or adding to the overall class (i.e., not an annotation of another annotation). 
 
--- a/docs/editing_guidelines.md
+++ b/docs/editing_guidelines.md
@@ -111,7 +111,7 @@ ORCID identifiers may also be used when the only available source for an asserti
 
 For CURIEs and ORCIDs: In the "Create Annotation" window, select the annotation property **database_cross_reference**.
 
-For CURIEs: Enter the CURIE, using the [bioregistry OBO context](https://bioregistry.io/context/obo) prefix ([link to prefixmap](https://github.com/biopragmatics/bioregistry/blob/main/exports/contexts/obo.context.jsonld), as a Value on the "Literal" tab. Leave Datatype empty.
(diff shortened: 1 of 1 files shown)


docs/editing_guidelines.md
```diff
@@ -90,6 +90,25 @@ If the main source for an assertion is a term in another ontology, the short ide
...
+
+If the cross-reference is an ORCID, enter the ORCID as an IRI in the IRI field on the “IRI Editor” tab.
+
+To restate, in all cases except ORCIDs, the CURIE is preferred and entered as a literal string. For ORCIDs, the IRI is preferred and is entered as an IRI.
+
+The above instructions apply whether the cross-reference is added to another annotation (e.g., annotating a text defintion or comment) or adding to the overall class (i.e., not an annotation of another annotation). 
+
+## Term contributors
+
+ When adding an ORCID to identify a term contributor, in the "Create Annotation" window, select the annotation property **dcterms:contributor**. The ORCID is still entered as an IRI on the “IRI Editor” tab.
```
aleixpuigb, line 111:
```suggestion
When adding an ORCID to identify a term contributor, in the "Create Annotation" window, select the annotation property **dcterms:contributor**. The ORCID is still entered as an IRI on the “IRI Editor” tab.
```

docs/editing_guidelines.md
```diff
@@ -90,6 +90,25 @@ If the main source for an assertion is a term in another ontology, the short ide
 
 ORCID identifiers may also be used when the only available source for an assertion is an individual researcher. This should be done sparingly.
 
+**Technical details of adding a cross-reference using Protégé**:
+
+For all: In the "Create Annotation" window, select the annotation property **database_cross_reference**.
+
+If the cross-reference can be represented as a CURIE and is not an ORCID: Enter the CURIE, using the bioregistry standard prefix, as a Value on the "Literal" tab. Leave Datatype empty.
```
matentzn, line 97:
```suggestion
If the cross-reference can be represented as a CURIE and is not an ORCID: Enter the CURIE, using the [bioregistry OBO context](https://bioregistry.io/context/obo) prefix ([link to prefixmap](https://github.com/biopragmatics/bioregistry/blob/main/exports/contexts/obo.context.jsonld), as a Value on the "Literal" tab. Leave Datatype empty.
```

docs/editing_guidelines.md
```diff
@@ -90,6 +90,25 @@ If the main source for an assertion is a term in another ontology, the short ide
...
+
+Non-CURIE values (e.g., URLs) are discouraged, but when used are also entered the same way, i.e., as a literal string with no Datatype selected.
```
matentzn, line 99: I would recommend to forbid them, and use rdfs:seeAlso for URLs. A URL is not a "database cross reference" even when the meaning is stretched.

docs/editing_guidelines.md
```diff
@@ -90,6 +90,25 @@ If the main source for an assertion is a term in another ontology, the short ide
...
+
+In cases where more than one CURIE is available for a resource, e.g., both a PMID and a doi are available, there is no preference for one over the other.
```
matentzn, line 101:
```suggestion
In cases where more than one CURIE is available for a resource, e.g., both a PMID and a doi are available, we recommend using PMID by default, as it is a more _semantically meaningful_ identifier: you can immediately see that the cross-reference is to a paper, while a doi could be essentially anything. However, both doi and PMID are acceptable.
```

@bvarner-ebi this is my personal preference, it is a bit more "guiding" them simply saying "it does not matter". What do you think?

docs/editing_guidelines.md
```diff
@@ -90,6 +90,25 @@ If the main source for an assertion is a term in another ontology, the short ide
...
+
+If the cross-reference is an ORCID, enter the ORCID as an IRI in the IRI field on the “IRI Editor” tab.
```
matentzn, line 103:
```suggestion
If the cross-reference is an ORCID, enter the ORCID as an IRI in the IRI field on the "IRI Editor" tab, for example `https://orcid.org/0000-0002-7356-1779`.
```

docs/editing_guidelines.md
```diff
@@ -90,6 +90,25 @@ If the main source for an assertion is a term in another ontology, the short ide
...
+
+To restate, in all cases except ORCIDs, the CURIE is preferred and entered as a literal string. For ORCIDs, the IRI is preferred and is entered as an IRI.
```
matentzn, line 105:
```suggestion
To restate, in all cases except ORCIDs, the CURIE is preferred and entered as a literal string. ORCIDs MUST be entered as an IRI.
```

matentzn (commented):
I recommend to study this: https://www.ietf.org/rfc/rfc2119.txt

Makes recommendations on the web much clearer!

docs/editing_guidelines.md
gouttegd: Of note, the “bioregistry OBO context” seems to recommend using `DOI` instead of `doi` as the “prefix” for DOI identifiers, which is against both what is recommended by the DOI specification and what we recommend ourselves slightly earlier in this very document.

gouttegd (approved):

docs/editing_guidelines.md
matentzn: Additionally, this is against bioregistry itself - it is a manual override done by the OBO context because most ontologies refer to dois using capital prefix. I am not opposed to trying to make a change here, but I would say we should stick to the context and debate capitalisation in that context rather than here

bvarner-ebi: For the purpose of this documentation, should the advise be to follow the [bioregistry standard](https://bioregistry.io/registry/doi) (doi) or the [OBO context](https://bioregistry.io/context/obo) (DOI)?

matentzn: I cant make that call alone, but I would prefer OBO context.

docs/editing_guidelines.md
bvarner-ebi: Thanks for the suggestion, @matentzn. How about:

In cases where more than one CURIE is available for a resource, either is acceptable, but using the more semantically specific identifier is recommended. For example, when both a PMID and a doi are available for a resource, using the PMID is recommended since it indicates the cross-reference points to a paper, as opposed to a doi which could point to any digital object.

docs/editing_guidelines.md
bvarner-ebi: Added to CL meeting agenda

matentzn: I think the people at your meeting wont be able to answer that question; its better to ask @anitacaron to raise it at the OBO ontology tech call (which as it so happens is also today), because this should not be decided on an ontology by ontology level.

bvarner-ebi: > I think the people at your meeting wont be able to answer that question; its better to ask @anitacaron to raise it at the OBO ontology tech call (which as it so happens is also today), because this should not be decided on an ontology by ontology level.

@anitacaron, can you raise this issue and the [one on URL links below](https://github.com/obophenotype/cell-ontology/pull/2058#discussion_r1256022366) at the OBO onto tech call?

anitacaron: Sure

bvarner-ebi: Per @anitacaron, it was decided that `doi:` should be the standard.

PR #2073, opened 2023-07-19
Addresses Refresh imports 07-2023

bvarner-ebi:
Imports refreshed in preparation of 07-2023 CL release.

Cursory check reveals expected updates.

### DIFF
--- a/src/ontology/mappings/fbbt-mappings.sssom.tsv
+++ b/src/ontology/mappings/fbbt-mappings.sssom.tsv
@@ -2,6 +2,7 @@
 #  FBbt: "http://purl.obolibrary.org/obo/FBbt_"
 #  UBERON: "http://purl.obolibrary.org/obo/UBERON_"
 #  CL: "http://purl.obolibrary.org/obo/CL_"
+#  semapv: "https://w3id.org/semapv/vocab/"
 #mapping_provider: "http://purl.obolibrary.org/obo/FBbt.owl"
 subject_id	subject_label	predicate_id	object_id	mapping_justification
 FBbt:00000001	organism	semapv:crossSpeciesExactMatch	UBERON:0000468	semapv:ManualMappingCuration
--- a/src/patterns/definitions.owl
+++ b/src/patterns/definitions.owl
@@ -7,8 +7,8 @@ Prefix(rdfs:=<http://www.w3.org/2000/01/rdf-schema#>)
 
 
 Ontology(<http://purl.obolibrary.org/obo/cl/patterns/definitions.owl>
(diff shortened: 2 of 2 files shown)


aleixpuigb (approved):

src/ontology/imports/merged_import.owl
```diff
@@ -116500,6 +116537,7 @@ AnnotationAssertion(Annotation(<http://www.geneontology.org/formats/oboInOwl#has
 AnnotationAssertion(<http://www.geneontology.org/formats/oboInOwl#id> <http://purl.obolibrary.org/obo/PR_000001006> "PR:000001006")
 AnnotationAssertion(rdfs:comment <http://purl.obolibrary.org/obo/PR_000001006> "Category=gene. Requested by=CL.")
 AnnotationAssertion(rdfs:label <http://purl.obolibrary.org/obo/PR_000001006> "receptor-type tyrosine-protein phosphatase C")
+AnnotationAssertion(<ttps://w3id.org/semapv/vocab/crossSpeciesExactMatch> <http://purl.obolibrary.org/obo/PR_000001006> <http://purl.obolibrary.org/obo/PR_P08575>)
```
gouttegd, line 116540: Something wrong happened here: the IRI for the annotation property should be `https://w3id.org/semapv/vocab/crossSpeciesExactMatch`, not `ttps://w3id.org/semapv/vocab/crossSpeciesExactMatch`.

This is probably an error in the upstream import, so I won’t block the PR for that, but this error must be noted. I’ll try tracking the provenance of the error.

gouttegd (approved):

src/ontology/imports/merged_import.owl
bvarner-ebi: Thanks for investigating, @gouttegd.

gouttegd: FYI: the error came from the PR slim (https://github.com/obophenotype/pro_obo_slim/issues/32). It’s fixed now, so the next time the imports are refreshed those incorrect IRIs will go away automatically.

bvarner-ebi: Thank you, @gouttegd. Unless advised to do sooner, the next import refresh will happen after the July release and before the August release.

anitacaron: If the issue is already fixed, I suggest refreshing the imports again today not to propagate the issue.

gouttegd: I agree. I don’t like the idea of making a release with known errors in it.

PR #2087, opened 2023-08-03
Change telocyte and lymphoblast subclass to seperate subclass

### LINKED ISSUE #2086: Telocyte and Lymphoblast do not have explicit parents
Current telocyte (CL:0017004) and lymphoblast (CL:0017005) do not have explicit parentage.
The only subclass they have is 
lymphoblast:
image
telocyte:
image
I am guessing this is why they seem orphaned in the .obo files
Can we split these up into separate subClassOf restrictions? Is there any implications on this?

shawntanzk:
Fixes #2086

Previously was intersection, making them orphany 
note: not 100% sure about implication, but I think this is correct?
Either way, should probably have explicit parentage 
thoughts @aleixpuigb and/or @bvarner-ebi?

Thanks


lubianat (approved):
Looks good for me as far as I can tell.

PR #2089, opened 2023-08-04
Replace invalid DOI IRIs in xrefs with doi: CURIE strings.

balhoff:
These values were stored as IRIs with the invalid protocol `DOI`.


anitacaron (approved):

anitacaron:
I am tagging @bvarner-ebi and @aleixpuigb as the core maintainers to check this too.

Issue #1263, opened 2021-09-17
As depicted_by annotations pointing to SwissBioPics


cmungall:
These are fantastic and used in Uniprot
https://www.swissbiopics.org/

CC-BY license:
https://www.swissbiopics.org/help

dosumis:
Cool.  I think they should be linked via depicted_by.  @paolaroncaglia  - can you have a go at doing this?

@cmungall  - wondering id possible to link down to the annotation level, and if so whether GO CC should be adding these links.

dosumis:
@paolaroncaglia  - Uberon example here for pattern: https://docs.google.com/presentation/d/11WeCHCeGYSPEO7hUYFTdPivptxX4ajj5pVHDm24j4JA/edit#slide=id.g9a5e4d7c09_0_163

paolaroncaglia:
Notes for self: if we do implement at the CL level:
https://www.swissbiopics.org/name/Animal_cell => CL:0000548 'animal cell'
https://www.swissbiopics.org/name/Animal_epithelial_cell => CL:0000066 'epithelial cell' (Equivalent To 'animal cell' and ('part of' some epithelium))
TBC

lubianat:
By the way, here is a SPARQL query for Wikidata items that have (1) a crossref to a CL ID (2) a picture in Wikimedia Commons: 

https://w.wiki/4F6T


All images in Commons have URLs and could be linked to CL elsewhere, though the design is not standardized: 

![image](https://user-images.githubusercontent.com/7917951/137942026-7645f368-d62a-4434-be05-083555cf0757.png)

paolaroncaglia:
This ticket might be better addressed by the Tech WG. It's not a priority, but it's quite consistent, and may have sub-tasks (see https://github.com/obophenotype/cell-ontology/issues/1263#issuecomment-922487106). I'll remove my assignment and add to the Tech board so that a strategy can be drafted. Feel free to assign any appropriate task to me once that's done. Thank you!

matentzn:
Good ticket for @ubyndr to start working on whenever there is nothing else urgent to do!

lubianat:
Adding tasks from @paolaroncaglia's notes. I'll try and add these depicted by statements. 

- [ ]  [swissbiopics.org/name/Animal_cell](https://www.swissbiopics.org/name/Animal_cell) => CL:0000548 'animal cell'
[...]

PR #2092, opened 2023-08-08
Add 'depicted by' statements to SwissBioPics

lubianat:
Related to issue #1263


src/ontology/cl-edit.owl
```diff
@@ -3604,6 +3604,7 @@ AnnotationAssertion(oboInOwl:hasDbXref obo:CL_0000025 "MESH:D010063")
 AnnotationAssertion(oboInOwl:hasExactSynonym obo:CL_0000025 "mature oocyte")
 AnnotationAssertion(oboInOwl:hasExactSynonym obo:CL_0000025 "ovum")
 AnnotationAssertion(rdfs:label obo:CL_0000025 "egg cell")
+AnnotationAssertion(<http://xmlns.com/foaf/0.1/depicted_by> obo:CL_0000025 "https://www.swissbiopics.org/name/Animal_egg_cell")
```
anitacaron, line 3607: It's expected to use a link with an image extension and the datatype `xsd:anyURI`. You could embed a `seeAlso` annotation to add the link you used here.

I found the link to the static image by "copying link address" (right click) in the arrow icon next to the title "Animal Egg cell" https://www.swissbiopics.org/api/image/Egg_cell.svg

anitacaron (changes requested):

src/ontology/cl-edit.owl
anitacaron: Please take a look at the same comment for each of your changes in the PR.

lubianat: @anitacaron thanks! I saw the SVG link too; the original URL seems to be a bit more informative, because the image becomes interactive, and one can see the different parts of the cell type. 

I'll make the changes, but do you know if these expectations are documented anywhere? 

http://xmlns.com/foaf/0.1/depicted_by seems to go 404, while http://xmlns.com/foaf/0.1/depiction exists with some documentation: 

https://xmlns.com/foaf/0.1/#term_depiction

lubianat: I've changed the datatype, but I am a bit unsure with regards to the best approach for the URL. 

I imagine that people accessing the ontology on OLS, for example, would have a better user experience by being directed to the interactive image.

Maybe we can discuss that briefly in the next CL call?

anitacaron: You're right. I can't find the property `foaf:depicted_by`, only `foaf:depiction`. We should change the property in another issue. It's used in many terms in Uberon, but only one here in CL.

If you checked the docs for `foaf:depiction`, the range is [foaf:Image](http://xmlns.com/foaf/0.1/#term_Image).

>Digital images (such as JPEG, PNG, GIF bitmaps, SVG diagrams etc.) are examples of [foaf:Image](http://xmlns.com/foaf/0.1/#term_Image).

OLS uses the property to render the image and only works with a digital image. [Example ](https://www.ebi.ac.uk/ols4/ontologies/uberon/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FUBERON_0000401). It won't work if you keep the link to a web page.

anitacaron: Feel free to add the link to the interactive image using the annotation `rdfs:seelAlso`

anitacaron (approved):

Issue #2103, opened 2023-08-16
nociceptor appears in both uberon and CL as different terms (CL edits)


bvarner-ebi:
**Context:** https://github.com/obophenotype/uberon/issues/3020

**[CL:0000198 ‘pain receptor cell’](http://purl.obolibrary.org/obo/CL_0000198)**

**Edit synonym annotations:**

- [x] add annotation ‘see also’ https://meshb.nlm.nih.gov/record/ui?ui=D009619 to exact synonym “nociceptor”

- [x] add annotation ‘see also’ https://www.thefreedictionary.com/nocireceptor to exact synonym “nocireceptor”

- [x] add term annotations:
'see also' MESH:A08.675.650.915.875; ‘see also’ MESH:A08.800.950.875; ‘see also’ MESH:A11.671.650.915.875

PR #2104, opened 2023-08-16
Addresses #2103 Edit 'pain receptor cell'

bvarner-ebi:
Addresses #2103 Edit 'pain receptor cell'


shawntanzk (approved):
not 100% sure about the use of seeAlso, but will trust you know what you are doing. 
Dont particularly like the use of freedictionary.com (not sure about archival policy and all) but I get that use of nocireceptor is rare in literature. Generally looks alright.

bvarner-ebi:
> not 100% sure about the use of seeAlso, but will trust you know what you are doing.
> Dont particularly like the use of freedictionary.com (not sure about archival policy and all) but I get that use of nocireceptor is rare in literature. Generally looks alright.

Updated reply.

Thanks for the prompt review.
The idea is that since the MeSH references don't appear to be in the OBO registry format (the references are TreeNumbers and not Unique IDs), 'see Also' was safer than database_cross_reference. The prefix does not create a hyperlink in Protégé like the Unique numbers do. Can adjust if advised.

If you have a preferred dbxref for nocireceptor, kindly advise.

shawntanzk:
> the references are TreeNumbers and not Unique IDs

ah makes sense. one possibility is to use IDs in mesh (eg mesh:D009619 - https://meshb.nlm.nih.gov/record/ui?ui=D009619). This prefix is registered in bioregistry so thats an extra nice thing

> The prefix does not create a hyperlink in Protégé

Prefix in protege only creates hyperlink if we go in and add it, see https://github.com/protegeproject/protege/pull/1049 for example

> If you have a preferred dbxref for nocireceptor, kindly advise.

The only one I got is not great too. ScienceDirect shows a bit of a dictionary that has it (https://www.sciencedirect.com/topics/immunology-and-microbiology/pain-receptor#:~:text=Nociceptors%20(nocireceptors%2C%20pain%20receptors),in%20response%20to%20certain%20stimuli.) 

This is the book https://www.sciencedirect.com/science/article/abs/pii/B9780123849311000143 but I cant download it to verify so dont like that too.

dosumis:
@bvarner-ebi  commented on Slack

A related question: are there considerations when adding classes (like excitatory/inhibitory (E/I) neuron) where the classes can never be defined in isolation?


In this example, there are no intrinsic properties of a cell type that alone could define it as an excitatory neuron. While glutamate is mainly an excitatory neurotransmitter in the CNS, this is dependent on the receptor of the post-synaptic cell. There are metabotropic glutamate receptors which are primarily inhibitory, so a neuron that releases glutamate is not de facto excitatory. Whether or not the cell is E/I is determined by external factors. If a glutamatergic neuron is adjacent to a cell with ionotropic glutamate receptors and another cell with metabotropic mGluR2/3 receptors, is it E or I?
With these considerations, it seems using a class like this would not be appropriate for annotating single cell data.

bvarner-ebi:
Confirmed with @anitacaron and @cmungall that there is an effort to replace MeSH Tree Numbers with MeSH Unique IDs.

#2105 removes the Tree Numbers, retains the Unique ID, and adds datatype anyURI to URL references, which was missed in this PR.

A separate issue will be made to replace MeSH Tree Numbers with Unique IDs across the ontology.

Issue #2113, opened 2023-08-18
Resolve remaining MeSH Tree Numbers on native CL classes


bvarner-ebi:
Following up on the work done in https://github.com/obophenotype/cell-ontology/pull/1558, this issue resolves [the remaining MeSH Tree Numbers that were unmapped](https://github.com/obophenotype/cell-ontology/pull/1558#issuecomment-1107905215) in native CL classes.

See Action items below:

ID | Label | Current TreeNumber | Proposed UniqueID mapping | Action
-- | -- | -- | -- | --
CL:0000568 | amine precursor uptake and decarboxylation cell | MESH:A11.070 | D001078 "APUD Cells" | Replace Tree Number with Unique ID. Move reference annotation to whole class, not defintion.
CL:0000132 | corneal endothelial cell | MESH:A09.371.060.067.31 | N/A- Did not find. Did find D004728 "Endothelium, Corneal", (A09.371.060.067.318), which appropriately appears as an annotation on UBERON:0001985 'corneal endothelium'. | Remove Tree Number. If a substitute is demanded, consider adding MESH:D004728, but may be redundant (already on UBERON term). Consider adding MESH:D055954 "Corneal Endothelial Cell Loss", but may not be appropriate scope (pathology) although it also references cell type.
CL:0000333 | migratory neural crest cell | MESH:A16.254.600 | N/A- Did not find. | Remove Tree Number. No suitable MeSH substitute identified. Replace with dbxref doi:10.1016/j.stem.2015.02.017
CL:0000577 | type EC enteroendocrine cell | MESH:A03.492.766.440.250 | D004759 "Enterochromaffin Cells" | Replace Tree Number with Unique ID. Move reference annotation to whole class, not defintion.
CL:0000588 | odontoclast | MESH:A11.329.679 | N/A- Did not find. | Remove Tree Number. No suitable substitute identified.
CL:0000547 | proerythroblast | MESH:A11.118.290.350.200 | N/A- Did not find. Closest appears to be D004900 "Erythrolasts" with related concept "Proerythroblasts". | Remove Tree Number. No suitable substitute identified.

Additionally, update editor doc advising to use MeSH Unique ID numbers, not Tree Numbers, in annotations.

PR #2114, opened 2023-08-18
Addresses #2113 Update MESH Tree Number cross-references

bvarner-ebi:
Addresses #2113 Update MESH Tree Number cross-references

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -4688,7 +4688,7 @@ SubClassOf(obo:CL_0000131 ObjectSomeValuesFrom(obo:RO_0002202 obo:CL_0000223))
 
 # Class: obo:CL_0000132 (corneal endothelial cell)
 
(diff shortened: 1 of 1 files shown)


src/ontology/cl-edit.owl
```diff
@@ -8269,9 +8269,10 @@ SubClassOf(obo:CL_0000567 obo:CL_0000198)
...
 AnnotationAssertion(oboInOwl:hasDbXref obo:CL_0000568 "BTO:0003866")
 AnnotationAssertion(oboInOwl:hasDbXref obo:CL_0000568 "FMA:83114")
+AnnotationAssertion(oboInOwl:hasDbXref obo:CL_0000568 "MESH:D001078")
```
aleixpuigb, line 8275: Is this DbXref supposed to be annotated to the defintion?

aleixpuigb (changes requested):
Consider changing the inline comment

src/ontology/cl-edit.owl
bvarner-ebi: My interpretation of the [conversation thread here](https://github.com/obophenotype/cell-ontology/issues/1557) is that there should be an effort to move these references to the class/node level.
Also, the MESH:[UniqueID] links do not direct to a page with a text definition: https://id.nlm.nih.gov/mesh/[UniqueID].html.
However, links with the format https://meshb.nlm.nih.gov/record/ui?ui=[UniqueID] (not the same as the MESH prefix) do include information for text definitions.

aleixpuigb (approved):

PR #2116, opened 2023-08-21
Update guidance on adding URL dbxrefs (editing_guidelines.md)

bvarner-ebi:
Update guidance on adding URL dbxrefs (editing_guidelines.md).

### DIFF
--- a/docs/editing_guidelines.md
+++ b/docs/editing_guidelines.md
@@ -113,7 +113,9 @@ ORCID identifiers may also be used when the only available source for an asserti
 
 **Technical details of adding a cross-reference using Protégé**:
 
-For CURIEs and ORCIDs: In the "Create Annotation" window, select the annotation property **database_cross_reference**.
+For CURIEs, ORCIDs: In the "Create Annotation" window, select the annotation property [**database_cross_reference**](http://www.geneontology.org/formats/oboInOwl#hasDbXref).
+For adding URLs to text definitions or synonyms:  In the "Create Annotation" window, select the annotation property [**database_cross_reference**](http://www.geneontology.org/formats/oboInOwl#hasDbXref).
+For adding URLs to axioms that are NOT text definitions or synonyms:  In the "Create Annotation" window, select the annotation property [**source**](http://www.geneontology.org/formats/oboInOwl#source).
 
 For CURIEs: Enter the CURIE, using the [bioregistry OBO context](https://bioregistry.io/context/obo) prefix ([link to prefixmap](https://github.com/biopragmatics/bioregistry/blob/main/exports/contexts/obo.context.jsonld)), as a Value on the "Literal" tab. Leave Datatype empty.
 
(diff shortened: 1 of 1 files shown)


docs/editing_guidelines.md
```diff
@@ -113,21 +113,20 @@ ORCID identifiers may also be used when the only available source for an asserti
...
 
 For CURIEs: Enter the CURIE, using the [bioregistry OBO context](https://bioregistry.io/context/obo) prefix ([link to prefixmap](https://github.com/biopragmatics/bioregistry/blob/main/exports/contexts/obo.context.jsonld)), as a Value on the "Literal" tab. Leave Datatype empty.
 
 In cases where more than one CURIE is available for a resource, either is acceptable, but using the more semantically specific identifier is recommended. For example, when both a PMID and a doi are available for a resource, using the PMID is recommended since it indicates the cross-reference points to a paper, as opposed to a doi which could point to any digital object.
 
 For ORCIDs: Enter the ORCID as an IRI in the IRI field on the "IRI Editor" tab, for example `https://orcid.org/0000-0002-7356-1779`.
 
-For URLs: In the "Create Annotation" window, select the annotation property **'see also'** (`rdfs:seeAlso`). Enter the URL as a literal string with Datatype `xsd:anyURI` selected.
-DO NOT use database_cross_reference with a value that is a URL. 
+For URLs: Enter the URL as a literal string with Datatype `xsd:anyURI` selected.
```
matentzn, line 126: You need to test this, I dont remember. Do it both ways save it check diff. Than compare to rest of Uberon

bvarner-ebi: @matentzn, to clarify, this is CL documentation, and this section differs from Uberon edits ([which you reviewed here](https://github.com/obophenotype/uberon/pull/3030/files)). Uberon currently does NOT save the xsd:URI datatype, reportedly due to it being obo format. 
cl-edit.owl will retain this datatype, Uberon does not.

matentzn: Yes, but you still want to retain some kind of "sane" obo serialisation? (cl.obo)..

gouttegd: @matentzn Don’t think there is any problem here. When converting to OBO it would be serialised either as `[http://example.org/my/cross/reference]` (if it’s a cross-reference annotation) or as `{source=http://example.org/my/source}` (if it’s a source annotation).

In both cases the `anyURI` datatype information would be lost in the OBO product (when reading such an OBO file again the annotation would be typed as a string), but since for CL OBO is only an export format and not the edit format this shouldn’t be a problem.

matentzn: Ok!

gouttegd (approved):

docs/editing_guidelines.md
bvarner-ebi: Thanks for the review/feedback!

PR #2134, opened 2023-09-01
Add OMO as import

anitacaron:
This is importing two annotations:
- OMO:0002000 (defined by construct) - already used by `never_in_taxon`
- OMO:0003004 (plural form) - to be used instead of the annotations `http://purl.obolibrary.org/obo/uberon#PLURAL`, 
`http://purl.obolibrary.org/obo/uberon/core#PLURAL` and `http://purl.obolibrary.org/obo/cl#PLURAL`


bvarner-ebi (changes requested):
Branch has conflicts

dosumis (approved):

bvarner-ebi (approved):

Issue #2148, opened 2023-09-12
Remove tissue region specificity from term label or else NTR


jychien:
The children of [L2/3-6 intratelencephalic projecting glutamatergic cortical neuron](https://www.ebi.ac.uk/ols/ontologies/cl/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FCL_4023040) have the text " of primary motor cortex" at the end of the term label:

<img width="824" alt="Screenshot 2023-09-12 at 11 32 00 AM" src="https://github.com/obophenotype/cell-ontology/assets/52470876/acdaa0f2-1297-4a2c-9019-048718f9b59f">\

Is there a particular reason the region of the brain was added to the term names? The primary motor cortex is not referenced in the definition. And there isn't a term for just "L2/3 intratelencephalic projecting glutamatergic neuron". Can the children terms have either have " of primary motor cortex" removed from label, or can a new term be added without the specific brain tissue location? That way, cells annotated as, for example, **L2/3 IT** can be mapped to "L2/3 intratelencephalic projecting glutamatergic neuron" rather than have to default to the parent "L2/3-6 intratelencephalic projecting glutamatergic cortical neuron" since it is not definitively part of the primary motor cortex.

Example journals: 

- https://www.biorxiv.org/content/10.1101/2023.05.08.539485v2.full.pdf (https://cellxgene.cziscience.com/e/c2876b1b-06d8-4d96-a56b-5304f815b99a.cxg/)
- https://www.nature.com/articles/s41586-021-03500-8 (https://cellxgene.cziscience.com/e/35081d47-99bf-4507-9541-735428df9a9f.cxg/)

bvarner-ebi:
Thank you for the questions, @jychien.

>Is there a particular reason the region of the brain was added to the term names?

The cross-references for classes L2/3 - L5 are specific to motor cortex. The cross-reference for the L6 class appears to be focussed on somatosensory cortex, so I cannot explain that one.

> The primary motor cortex is not referenced in the definition.

It is referenced in all logical definitions for these classes and in all text definitions except for 'L2/3 intratelencephalic projecting glutamatergic neuron of the primary motor cortex'. The text definition for this class can be updated to be aligned with the rest.

Thank you for the references. The second reference appears to be focussed on primary motor cortex, so those terms could be mapped to the already existing CL classes.
[...]

PR #2149, opened 2023-09-14
Addresses #2148 Edit text definition of 'L2/3 intratelencephalic projecting glutamatergic neuron of the primary motor cortex'

bvarner-ebi:
Addresses #2148 Edit text definition of 'L2/3 intratelencephalic projecting glutamatergic neuron of the primary motor cortex' so it is aligned with logical definition.


bvarner-ebi:
FYI @jychien

aleixpuigb (approved):

PR #2168, opened 2023-10-03
Fixes #2167 NTR onychocyte

### LINKED ISSUE #2167: [NTR] onychocyte

**Preferred term label**
onychocyte

A differentiated flat keratinocyte that is part of a nail. An onychocyte is firmly adherent and do not desquamate.
PMID:24079581

keratinocyte

`'part of' some nail`

**Your ORCID**
https://orcid.org/0000-0001-6677-8489

aleixpuigb:
Fixes #2167 NTR onychocyte


bvarner-ebi (commented):
- There are several inherited SubClass Of statements that may not apply to this cell type, e.g., 'capable of' some 'positive regulation of neutrophil chemotaxis'; 'capable of' some 'antimicrobial peptide secretion'; 'part of' some 'skin epidermis'; etc
- UBERON considers the 'nail plate' to be part of 'skin epidermis' and this cell type is also considered to be part of 'skin epidermis'. Consider confirming this with a SME.
- 'nail plate' has SubClass Of 'has part' some keratinocyte. Consider if adding 'nail plate' SubClass Of 'has part' some onychocyte is appropriate.

aleixpuigb:
Thank you for your comments @bvarner-ebi 

> * There are several inherited SubClass Of statements that may not apply to this cell type, e.g., 'capable of' some 'positive regulation of neutrophil chemotaxis'; 'capable of' some 'antimicrobial peptide secretion'; 'part of' some 'skin epidermis'; etc

I agree that some of them might not be applicable, thank you for noticing. However, that should be addressed in a different ticket as that is a problem of keratinocytes, since the axiom would not be applicable to all cells. I will do some curation to which axioms should be removed to the general class and be added to the more specific terms.

> * UBERON considers the 'nail plate' to be part of 'skin epidermis' and this cell type is also considered to be part of 'skin epidermis'. Consider confirming this with a SME.

The SME (author of the ASCT+B skin table) considers the nail plate part of the skin epidermis.

> * 'nail plate' has SubClass Of 'has part' some keratinocyte. Consider if adding 'nail plate' SubClass Of 'has part' some onychocyte is appropriate.

I could not confirm that onychocyte is the only keratinocyte present in the nail plate, hence I have not used a logical definition. Since, onychocyte is 'part of' some 'nail plate', I don't think it is necessary to have the relation in the opposite direction with 'has part' and the existing axiom might be still useful.

bvarner-ebi (approved):

PR #2174, opened 2023-10-06
Added ‘has soma location’ RO:0002100 to get_scope_query method

### LINKED ISSUE #1955: [slim coverage report] Add 'has soma location'
There are some neurons that are in an organ but the dendrites are not, and they use the relationship 'has soma location' RO:0002100. Currently, these classes (eg retinal ganglion cell in the eye) are not covered by the report. Would it be possible to include it?

ubyndr:
Resolves #1955 

Tested it with eye_upper_slim.csv and  UBERON:0000970. You can see the output of coverage calculation below.

[eye_slim_test_no_soma.csv](https://github.com/obophenotype/cell-ontology/files/12832769/eye_slim_test_no_soma.csv)
[eye_slim_test.csv](https://github.com/obophenotype/cell-ontology/files/12832770/eye_slim_test.csv)

### DIFF
--- a/src/scripts/generic_coverage.py
+++ b/src/scripts/generic_coverage.py
@@ -92,10 +92,12 @@ def get_scope_query(scope_term: str) -> str:
         PREFIX owl: <http://www.w3.org/2002/07/owl#>
         PREFIX UBERON: <http://purl.obolibrary.org/obo/UBERON_>
         PREFIX CL: <http://purl.obolibrary.org/obo/CL_>
+        PREFIX BFO: <http://purl.obolibrary.org/obo/BFO_>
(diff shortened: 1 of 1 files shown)


aleixpuigb (approved):
I base the review on the output, but not on the code.
Adding the object property 'has soma location' to the query increases the number of cell types that are located in a certain organ (in the eye in this case).
Therefore, as 'retinal ganglion cell' is not in the slim list, all the terms under this class are only captured when 'has soma location' has been added, dropping the coverage percentage.
It might still be good that another person reviews the code, but I will leave it to the assignee's consideration.

ubyndr:
We can merge it since nothing major has changed in the code

dosumis (approved):

PR #2182, opened 2023-10-12
Add slim coverage check running on local (branch) copy CL

### LINKED ISSUE #2177: Add slim coverage check as GitHub Action running on local (branch) copy CL
Running against UberGraph means that changes to the ontology are not reflected in the report.  Could we acheieve this by running RelationGraph for the relevant relation types in a branch and then running SPARQL queries as we currently do against UberGraph.

CC @bvarner-ebi

ubyndr:
Resolves #2177 

We decided not to follow the GitHub action route in this issue. Editors should use the following commands to run coverage reports via the Makefile in the ontology directory.;

- for specific upper_slim report:
```sh
sh run.sh make reports/general_cell_types_upper_slim_report.csv
```
```sh
sh run.sh make reports/blood_and_immune_upper_slim_report.csv
```
```sh
sh run.sh make reports/eye_upper_slim_report.csv
```
- for all upper_slim reports:
```sh
sh run.sh make slim_coverage
```

cc @aleixpuigb @bvarner-ebi


requirements.txt
```diff
@@ -1 +1,4 @@
 mkdocs
+
+rdflib~=7.0.0
+SPARQLWrapper~=2.0.0
```
anitacaron, line 4: This is part of the ODK, you don't need to install :)

src/scripts/requirements.txt
anitacaron, line 1: empty requirements.txt file?

requirements.txt
ubyndr: Bradley wanted to use script itself without running it through make command, so it is kinda needed, isn't it?

anitacaron: is there a reason?

anitacaron: He can run the command without make but still use ODK, which is recommended.

src/scripts/requirements.txt
anitacaron: in any case, I don't see this is needed.

requirements.txt
ubyndr: cc @bvarner-ebi

aleixpuigb (dismissed):
The tests performed have been successful, with a running time of 20 minutes.

src/scripts/requirements.txt
```diff
@@ -0,0 +1 @@
+rdflib~=7.0.0
```
anitacaron, line 1: really need this?

ubyndr: I had deleted it from my local branch, dunno why it still exists in the PR...

src/scripts/generic_coverage.py
```diff
@@ -98,7 +127,6 @@ def get_scope_query(scope_term: str) -> str:
         WHERE
         {{
           ?scope_member rdfs:subClassOf|BFO:0000050|RO:0002100 {_scope} .
```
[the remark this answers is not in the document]
ubyndr: relation-graph command should help with these, shouldn't it? Did I misinterpreted the command's function?

aleixpuigb (changes requested):
I have test the eye slim with `eye` UBERON:0000970 this time and the coverage includes UBERON terms which should not. Moreover, it doesn't cover many cell types that it used to. Are the object properties overlaps or 'has soma location' (subproperty of overlaps) included?
The running time was 13:30 minutes.

src/scripts/generic_coverage.py
```diff
@@ -124,7 +152,6 @@ def get_superclass_value_query(term_iri_list: List[str], _scope: str) -> str:
             {{
               ?term rdfs:subClassOf ?super. ?super rdfs:label ?label. 
               ?super rdfs:subClassOf|BFO:0000050 {_scope}.
-              ?super rdfs:isDefinedBy <http://purl.obolibrary.org/obo/cl.owl> .
```
anitacaron, line 127: Should change it to filter:

```
FILTER (STRSTARTS(STR(?super), "http://purl.obolibrary.org/obo/CL_"))
```

src/scripts/generic_coverage.py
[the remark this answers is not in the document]
anitacaron: Need to double-check.

bvarner-ebi (approved):
Report successfully generated

aleixpuigb (approved):
The report has been generated with the expected coverage and uncovered terms.

Issue #1281, opened 2021-10-07
[NTR] Ureteric bud cell


zimmwe:
Note - please check that the term does not already exist (check OLS: https://www.ebi.ac.uk/ols/ontologies/cl)

**Preferred term label:**
Ureteric bud cell

**Synonyms**

**Definition (free text, please give PubMed ID in format PMID:XXXXXX)**
Epithelial cells of the uretic bud.

PMID: 9374839
PMID: 24183650
PMID: 17133361

**Parent cell type term (check the hierarchy here https://www.ebi.ac.uk/ols/ontologies/cl):**
kidney epithelial cell  CL:0002518

**What is the anatomical structure that the cell is a part of? Please check Uberon: https://www.ebi.ac.uk/ols/ontologies/uberon**
ureteric bud  UBERON:0000084

**Your nano-attribution (ORCID)**
0000-0002-1425-877X

**Any additional notes or concerns**

paolaroncaglia:
Dear @zimmwe ,

Thank you for your request.

Similarly to your other ticket, may I please ask what project or group is this for? So we can assign the task appropriately.

Also, could you please confirm if the proposed definition 
"Epithelial cells of the uretic bud."
should read
"Epithelial cells of the _ureteric_ bud."
instead?

Thanks.

bvarner-ebi:
@emquardokus, checking in on this ticket.
Is this a cell type you still plan to add? If so, I suggest to remove the "stale" label. The original request is from 2021.

emquardokus:
@bvarner-ebi removed "stale" label; Paola had been working on this; I think we figured it had KPMP potential as an expert group. Created and added NIBR. This is not part of a HuBMAP ASCT+B table as it's kidney development.

bvarner-ebi:
@zimmwe, @emquardokus,
Apologies for the extended delay on this ticket. 
After reviewing the references above, below are the proposed edits.
It would be appreciated if a subject matter expert will confirm that what is noted in the definitions are broadly applicable and not overly specific to particular taxa.

--

**Preferred term label:**
ureteric bud cell

**Synonym**
UB cell (related, abbreviation, PMID:24183650)

**Text Definition**
[...]

PR #2185, opened 2023-10-13
Addresses #1281 NTR 'ureteric bud cell'

bvarner-ebi:
Addresses #1281 NTR 'ureteric bud cell'

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -2937,6 +2937,7 @@ Declaration(Class(obo:CL_4030055))
 Declaration(Class(obo:CL_4030056))
 Declaration(Class(obo:CL_4030057))
 Declaration(Class(obo:CL_4030058))
+Declaration(Class(obo:CL_4030066))
 Declaration(Class(obo:CL_4032000))
 Declaration(Class(obo:CL_4032001))
 Declaration(Class(obo:CL_4033000))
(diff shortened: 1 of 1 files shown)


aleixpuigb (commented):
Consider adding UB cell as a related synonym (abbreviation)
https://github.com/obophenotype/cell-ontology/issues/1281#issuecomment-1750806327

src/ontology/cl-edit.owl
```diff
@@ -31105,6 +31106,16 @@ AnnotationAssertion(Annotation(oboInOwl:hasDbXref "PMID:37292694") Annotation(ob
 AnnotationAssertion(rdfs:label obo:CL_4030065 "L6 intratelencephalic projecting glutamatergic neuron")
 EquivalentClasses(obo:CL_4030065 ObjectIntersectionOf(obo:CL_4023040 ObjectSomeValuesFrom(obo:RO_0002100 obo:UBERON_0005395)))
 
+# Class: obo:CL_4030066 (ureteric bud cell)
+
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref "PMID:17133361") Annotation(oboInOwl:hasDbXref "PMID:24183650") Annotation(oboInOwl:hasDbXref "PMID:9374839") obo:IAO_0000115 obo:CL_4030066 "An epithelial cell that is part of a ureteric bud. A ureteric bud cell has the potential to induce metanephric mesenchymal cells to proliferate and convert to epithelia that form renal tubules via: (1) the secretion of multiple diffusible growth factors that rescue renal progenitors from apoptosis and (2) contact-dependent mechanisms that induce mesenchymal-epithelial conversion.")
```
aleixpuigb, line 31111: > A ureteric bud cell has the potential to induce metanephric mesenchymal cells to proliferate and convert to epithelia that form renal tubules via

These are two separate processes, and from the reference, during proliferation there is not epithelialization:
image
Consider clarifying that both processes are independent.

> the secretion of multiple diffusible growth factors that rescue renal progenitors from apoptosis

As it is written, I understand that this is the mechanism of how UB cells induce proliferation. However, while growth factors can induce this proliferation, there is only reference to the apoptosis rescue. Consider either removing the "via" fro proliferation, or adding the mechanism of proliferation.

aleixpuigb (changes requested):
Consider the suggestions on the textual definition.

src/ontology/cl-edit.owl
aleixpuigb: I think that is a good solution for the second point, thank you. However, in the previous sentence, consider making clear that proliferation/apoptosis inhibition and epithelialization are not occurring simultaneously.

bvarner-ebi: > consider making clear that proliferation/apoptosis inhibition and epithelialization are not occurring simultaneously.

Is that not clear by breaking it into 2 parts in the definition? Two separate mechanisms: (1) proliferation via signalling from diffusible growth factors and (2) epithelialisation via contact-dependent mechanisms.

aleixpuigb: > How about the following:
"... (1) the secretion of multiple diffusible growth factors that rescue renal progenitors from apoptosis and stimulate them to proliferate and (2) ..."

Once these changes are implemented, I will approve the PR.

aleixpuigb (approved):

PR #2204, opened 2023-10-19
Remove `oort` step from the release pipeline

anitacaron:
With the new OWLAPI 4.5.26, which allows arbitrary annotation properties in the OBO parser and oort using a previous OWLAPI version, we are having conflicts when converting it. Also, the oort step is not used in any release artefact.


gouttegd (approved):
No objection at all. The `oort` target serves no useful purpose and relies on obsolete tooling that cannot cope with recently introduced changes in OWL API/ROBOT.

In the future, we should go further and actually remove the now unused OORT-dependent rules, but making sure those rules are not invoked anymore is a good start.

PR #2211, opened 2023-10-31
Adding `ontobot-change-agent` to create PRs based on appropriate Issues.

hrshdhgd:
This PR adds a workflow `ontobot.yml` which monitors issues created and on finding issues following a particular pattern (with issue starting with `Hey ontobot, apply:` followed by KGCL commands as bullets)  will trigger the workflow that performs suitable actions on the resource file and create a PR reflecting the changes reflected.

Example:
- Issue: Obsoleting a term: https://github.com/hrshdhgd/cell-ontology/issues/9
- Corresponding PR: https://github.com/hrshdhgd/cell-ontology/pull/10

TODO: Have a `GH_TOKEN` in the secrets section of this repo for all of this to work.


cmungall:
I added GH_TOKEN

.github/workflows/ontobot.yml
```diff
@@ -0,0 +1,82 @@
...
+            return repo
+
+      - name: Set branch name
+        id: vars
+        run: |
+          echo "resource=src/ontology/cl-edit.owl" >> $GITHUB_ENV
+          echo "branch-name=kgcl_automation_"${{ steps.gh-script-issue.outputs.result }} >> $GITHUB_ENV
+      
+      - name: Get jar & create alias command.
+        run: wget https://github.com/gouttegd/kgcl-java/releases/download/kgcl-0.2.0/kgcl-robot-standalone-0.2.0.jar -O kgcl-robot.jar
```
gouttegd, line 52: This is downloading a full-fledged ROBOT archive (nearly ~90 MB). It’s a shame given that the plugin version is only ~2 MB and ROBOT now has support for plugins.

.github/workflows/ontobot.yml
```diff
@@ -0,0 +1,82 @@
...
+        run: |
+          pip install ontobot-change-agent
+
+      - name: Run ochange
+        id: ochange
+        run: |
+          ochange process-issue ${{ env.resource }} \
+          -r ${{ steps.gh-script-repo.outputs.result }} \
+          -n ${{ steps.gh-script-issue.outputs.result }} \
+          -g ${{ secrets.GH_TOKEN }} \
```
gouttegd, line 64: Is there a particular reason for not using the default `secrets.GITHUB_TOKEN`, which is automatically set up by GitHub with each workflow?

.github/workflows/ontobot.yml
gouttegd: I’ll open a ticket on [ontobot-change-agent](https://github.com/hrshdhgd/ontobot-change-agent) to discuss how the agent could be updated to use a ROBOT plugin when available.

.github/workflows/ontobot.yml
hrshdhgd: Honestly, we did not notice another existing token. Must have been a foresight, sorry.

gouttegd: No worries. :)

.github/workflows/ontobot.yml
gouttegd: The ticket is here: https://github.com/hrshdhgd/ontobot-change-agent/issues/67

gouttegd: In the meantime, I have no objection to start using ontobot-change-agent without waiting for that change to be implemented. It can always be updated later.

gouttegd (dismissed):
As discussed above, I think we should avoid downloading a full version of ROBOT just for the `kgcl` command, but this is something that can always be improved later.

It looks good to me and I’m eager to see it in action.

hrshdhgd:
@gouttegd : I have updated the workflow to use `kgcl-java` as a robot plugin (for robot within ODK) rather than downloading the standalone resource (although that feature is also available) - exactly as you had envisioned.

gouttegd (approved):

PR #2223, opened 2023-11-17
File clean up fix for generic_coverage.py

ubyndr:
Files generated during coverage calculation are causing issues in certain commits. @bvarner-ebi has requested that we address these files within the script.

### DIFF
--- a/src/scripts/generic_coverage.py
+++ b/src/scripts/generic_coverage.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+
 import argparse
 import csv
 from collections import defaultdict
@@ -256,6 +257,10 @@ def modify_docker_script(input_file, _output_file):
         f.writelines(lines)
 
 
+def clean_up():
+    os.remove("test.ttl")
+
+
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     parser.add_argument(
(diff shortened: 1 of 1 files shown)


src/scripts/generic_coverage.py
```diff
@@ -329,8 +334,6 @@ def modify_docker_script(input_file, _output_file):
     cl_rel = Graph().parse("../ontology/test.ttl", format="ttl")
     cl = cl_base + cl_rel
 
-    cl.serialize("ontology.owl", format="turtle")
-
```
anitacaron, line 333: Instead, you could save in the `../ontology/tmp/` folder

anitacaron (changes requested):

src/scripts/generic_coverage.py
ubyndr: It wasn't needed anyway. I saved it to check something for debug purposes.

bvarner-ebi (dismissed):
_test.ttl_ and _ontology.owl_ no longer persist after running `sh run.sh make reports/kidney_upper_slim_report.csv`. 
_definitions.owl_ still persists.

Report seems to be generated the same as before.

src/scripts/generic_coverage.py
anitacaron: Better to add the test.ttl to the tmp folder, instead

ubyndr: done

anitacaron (approved):

PR #2235, opened 2023-11-24
Fixes #1889 neuroendocrine cell

### LINKED ISSUE #1889: neuroendocrine cell
Current def 

[neuroendocrine cell](http://purl.obolibrary.org/obo/CL_0000165)
A neuron that is capable of some hormone secretion in response to neuronal signals.
Equivalent to: ([neuron](https://www.ebi.ac.uk/ols4/ontologies/cl/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FCL_0000540?lang=en) and [capable of](https://www.ebi.ac.uk/ols4/ontologies/cl/properties/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FRO_0002215?lang=en) [(RO)](https://www.ebi.ac.uk/ols4/ontologies/ro/properties/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FRO_0002215) some [hormone secretion](https://www.ebi.ac.uk/ols4/ontologies/cl/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FGO_0046879?lang=en) [(GO)](https://www.ebi.ac.uk/ols4/ontologies/go/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FGO_0046879))

image

Potential issues:

1. Based on  the name, we might expect something stricter (corresponding to the endocrine part of the name):
Neuroendocrine cells are cells that receive neuronal input (through [neurotransmitters](https://en.wikipedia.org/wiki/Neurotransmitter) released by nerve cells or neurosecretory cells) and, as a consequence of this input, release messenger molecules ([hormones](https://en.wikipedia.org/wiki/Hormones)) into the blood.
"https://en.wikipedia.org/wiki/Neuroendocrine_cell"

But not clear that all cells with this name actually secrete into the circulatory system.

- [ ] Review all neuroendocrine cell subclasses for cases where secretion is not to circulatory system.

With examples in place, we need to decide whether to classify these under neuroendocrine or not (and if not to add comments explaining why) OR keep the current definition of neuroendocrine cell and add a comment to that to explain why it is not classified under endocrine cell.  I favour the latter. Comment should include examples.

2. It is not clear that all cells of this type should be considered neurons (current name is neutral but classification remains neuron.

Not sure whether anything more needed here.  There is already a comment that kind-of explains.

aleixpuigb:
Fixes #1889 neuroendocrine cell

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -5021,11 +5021,9 @@ AnnotationAssertion(Annotation(oboInOwl:hasDbXref "PMID:5342440") Annotation(obo
 
 The modern definition of neurosecretion has evolved to include the release of any neuronal secretory product from a neuron.")
 AnnotationAssertion(rdfs:label obo:CL_0000165 "neuroendocrine cell")
-EquivalentClasses(obo:CL_0000165 ObjectIntersectionOf(obo:CL_0000540 ObjectSomeValuesFrom(obo:RO_0002215 obo:GO_0046879)))
+EquivalentClasses(obo:CL_0000165 ObjectIntersectionOf(obo:CL_0000163 ObjectSomeValuesFrom(obo:BFO_0000050 obo:UBERON_8600018)))
 SubClassOf(obo:CL_0000165 obo:CL_0000163)
 SubClassOf(obo:CL_0000165 obo:CL_0000393)
-SubClassOf(obo:CL_0000165 obo:CL_0000527)
-SubClassOf(obo:CL_0000165 obo:CL_0000710)
 
 # Class: obo:CL_0000166 (chromaffin cell)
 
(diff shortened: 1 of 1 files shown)


aleixpuigb:
#gogoeditdiff

aleixpuigb:
Hi @dosumis, I think all discussed edits have been implemented. Cheers!

aleixpuigb:
#gogoeditdiff

gouttegd:
@dosumis: Given that it has been a long time since the last discussion on #1889, is the fix proposed here (which is basically to define `neuroendocrine cell` as `'endocrine cell that is part_of some 'neuroendocrine system'`, as suggested in [this comment](https://github.com/obophenotype/cell-ontology/issues/1889#issuecomment-1694675846)) still your preferred approach? Or should we rethink about the entire issue?

dosumis:
#gogoeditdiff

Note - we will ask for feedback on this issue from the Gut Cell Atlas team - who will be reviewing EC cells, a number of which are sometimes classified as neuro-endocrine.

scheuerm:
This is also a cell type defined in the lung atlas and so we should make sure that the definition is consistent with this location as well.

Issue #2245, opened 2023-12-18
Multiple issues with retinal photopic cells


cmungall:
There are many issues with the classification of terms highlighted in the image below:

<img width="1551" alt="image" src="https://github.com/obophenotype/cell-ontology/assets/50745/923ef8e4-45a5-4107-9606-906570077d4d">

(drawn with OAK and `cl viz -p i --down CL:0000490 CL:0000749 "OFF-bipolar cell" CL:0000573`)

- why is retinal cone cell not classified under retinal cell? 
- why is photopic receptor cell not classified under visual system neuron?
- why isn't ON-blue cone bipolar cell not classified under retinal cone cell?
- the classification of photopic receptor cell seems massively incomplete?
- etc

Overall it looks like there are just too many axes of classification going on with insufficient axiomatization to render them complete

In some cases it seems like CL has grouping classes that are just inherently hard to make complete. E.g. my understanding is that role in photopic vision is not discrete. cones are typically photopic, rods typically scotopic, but this is not all or nothing (if it were, then there would be no need for a distinct "photopic receptor cell" classification)

Some of these groupings may be historic - e.g it looks like both photopic and scotopic receptor cells have been there from the early days of CL, and they are undefined. These could probably be obsoleted. 

The classification here appears wrong:

* [] CL:0000210 ! photoreceptor cell
    * [i] **CL:0000489 ! scotopic photoreceptor cell**
    * [i] CL:0000488 ! visible light photoreceptor cell
        * [i] **CL:0000490 ! photopic photoreceptor cell**

AFAIK, both scotopic and photopic are visible, the distinction is to do with light levels. `visible light photoreceptor cell` is also massively incomplete - other than photopic, it only has fly neurons.

I can see a number of challenges with making axiomatization complete here related to species specificity. There is a lot of tacit information to navigate around; we have the terrible GO-ish "camera-type eye" which needs to be fixed in GO and Uberon, there is also the fact that "retina" is used for the vertebrate structure, and the more generic photoreceptor array used for adult Drosophila retinal cells. And there are the usual annoying exceptions to photoreceptors being found in retinas. In fact CL:0000210 photoreceptor cell could technically be used by PO. 

This is all compounded by the different uses of has-soma-location vs part of in the ontology...

dosumis:
I'm dubious about many grouping classes here:

Suggested obsoletions:

- retinal cell (to general)
- scotopic photoreceptor cell (most photoreceptor cell mix both roles?)
[...]

PR #2248, opened 2023-12-18
Making camera eye photoreceptor cell part of retina.

cmungall:
See #2245

Note that this is a hidden GCI, which acts as a rule - I
don't there are photoreceptors in a vertebrate eye that are
not in the retina


dosumis:
#gogoeditdiff

dosumis (approved):

PR #2314, opened 2024-03-25
Use a dedicated token for the Ontobot workflow.

gouttegd:
This commit updates the ontobot workflow to make it use a dedicated token (`ONTOBOT_TOKEN`), rather than the default `GITHUB_TOKEN`.

This should ensures that PRs created by the bot can trigger other workflows, notably the CI workflow.

(This is merely the same thing as what was [already done on Uberon](https://github.com/obophenotype/uberon/pull/3225), for the same reasons.)

### DIFF
--- a/.github/workflows/ontobot.yml
+++ b/.github/workflows/ontobot.yml
@@ -50,7 +50,7 @@ jobs:
         id: gh-script-issue
         uses: actions/github-script@v6
         with:
-          # github-token: ${{ secrets.GH_TOKEN }}
+          github-token: ${{ secrets.ONTOBOT_TOKEN }}
           script: |
             const issue_number = context.issue.number
             const repo = context.repo.owner+"/"+context.repo.repo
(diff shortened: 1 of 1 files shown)


.github/workflows/ontobot.yml
```diff
@@ -50,7 +50,7 @@ jobs:
         id: gh-script-issue
         uses: actions/github-script@v6
         with:
-          # github-token: ${{ secrets.GH_TOKEN }}
+          github-token: ${{ secrets.ONTOBOT_TOKEN }}
```
hrshdhgd, line 53: This one may not be needed.

.github/workflows/ontobot.yml
```diff
@@ -60,7 +60,7 @@ jobs:
         id: gh-script-repo
         uses: actions/github-script@v6
         with:
-          # github-token: ${{ secrets.GH_TOKEN }}
+          github-token: ${{ secrets.ONTOBOT_TOKEN }}
```
hrshdhgd, line 63: This one may not be needed.

hrshdhgd (approved):
The 2 extra places where the secret is passed made no difference to the workflow (when I first tested it). I kept it there  at the time just in case. Looking at what the `steps` are doing, we may not need them at all but it doesn't really impact anything. So I leave it unto you.

Issue #2342, opened 2024-04-19
Add definitions to terms for neurons that lack them


dosumis:
We have 71 terms for types of neuron that lack a text definition (see table below).  We should define them

Strategies for finding more info:

1.  Many of these are  We need to be able to see xrefs coming back in from external ontologies (particularly FBbt). These now live externally. @gouttegd  - what is the easiest way to query for them? Can we use UberGraph?
2. Many are grouping terms with logical defs, we can potentially use these logical defs to auto-roll textual definitions. @gouttegd - I believe you have an updated version of the software that FBbt uses for this.  How easy would that be to deploy here?
3. Some may be used by GO.  These can be easily queried for in UberGraph or CL-KB (Cypher).

@AvolaAmg  - can you prioritise terms that are obviously applicable to mammalian brain?

c.curie | c.label | c.definition | c.database_cross_reference
-- | -- | -- | --
CL:0000544 | slowly adapting mechanoreceptor cell | null | null
CL:0000200 | touch receptor cell | null | null
CL:0000203 | gravity sensitive cell | null | null
CL:0000006 | neuronal receptor cell | null | null
CL:1000082 | stretch receptor cell | null | null
CL:0000200 | touch receptor cell | null | null
CL:0000204 | acceleration receptive cell | null | null
CL:0000707 | R7 photoreceptor cell | null | null
CL:0000287 | eye photoreceptor cell | null | null
CL:0000705 | R6 photoreceptor cell | null | null
CL:0000690 | R2 photoreceptor cell | null | null
CL:0000694 | R3 photoreceptor cell | null | null
CL:0000697 | R4 photoreceptor cell | null | null
CL:0000702 | R5 photoreceptor cell | null | null
CL:0000687 | R1 photoreceptor cell | null | null
CL:0000709 | R8 photoreceptor cell | null | null
CL:0010009 | camera-type eye photoreceptor cell | null | null
CL:0000707 | R7 photoreceptor cell | null | null
CL:0000705 | R6 photoreceptor cell | null | null
CL:0000697 | R4 photoreceptor cell | null | null
CL:0000690 | R2 photoreceptor cell | null | null
CL:0000490 | photopic photoreceptor cell | null | null
CL:0000702 | R5 photoreceptor cell | null | null
CL:0000709 | R8 photoreceptor cell | null | null
CL:0000694 | R3 photoreceptor cell | null | null
CL:0000687 | R1 photoreceptor cell | null | null
[...]

PR #2347, opened 2024-04-19
Re-classify 'R7 photoreceptor cell' as part of an ommatidium.

gouttegd:
The `R7 photoreceptor cell` is explicitly classified as an `eye photoreceptor cell`. This is imprecise as `eye photoreceptor cell` refers to photoreceptors in any type of eye, while R7 photoreceptors are found in compound eyes.

This PR classifies `R7 photoreceptor cell` as being part of some `ommatidium` (as the other R1, R2, ... R8 photoreceptor cells). This automatically classifies it as a `compound eye photoreceptor cell`, as it should be.

(Found while working on #2342)


aleixpuigb (approved):
Agree with the edits.

Issue #2358, opened 2024-04-25
Let CL manage its own mappings


gouttegd:
The proposition outlined here will affect both CL and Uberon (and will require changes in both), but CL is the most concerned.

Currently, mappings between CL and other ontologies (especially species-specific ontologies) are mostly managed completely outside of CL.

Most of them are *edited* in CL (under the form of cross-references, aka the usual way), but CL itself does nothing with those cross-references. In particular, it does _not_ provide them as a SSSOM set.

Furthermore, two sets are completely external: the FBbt set (managed on the FBbt side, which provides it as a SSSOM file), and the ZFA set (managed on the ZFA side, which provides it as cross-references in the ZFA ontology).

The FBbt mapping set is used by CL to create the `mappings.owl` component, which contains the CL-to-FBbt mappings under the form of classical cross-references. This component is imported into CL, so that the CL-to-FBbt mappings, even if they are managed externally, are still present in the ontology itself (they are not **editable** there, but at least they are visible).

But CL is completely ignorant of the ZFA set. The CL-to-ZFA mappings are only extracted from ZFA as part of the **Uberon** build pipeline only. As a result, looking at CL itself, there is **no way**, even for CL **editors** (to say nothing of CL users) to know which CL terms are mapped to ZFA terms. The only way to learn of those mappings is either to look at ZFA, or to use the “meta” mapping set provided by Uberon (which contains all the mappings involving CL – including the CL-to-ZFA mappings –, not only the mappings involving Uberon). I don’t think this is satisfying.

Therefore I propose to “uncouple” CL from Uberon as far as managing the mappings are concerned. That is, CL should provide its own mapping set (including all the mappings from remote sources such as ZFA) without having to depend on the Uberon pipeline.

More precisely, that would mean that CL would:

* fetch the FBbt mapping set (as it already does, no changes there);
* fetch a copy of ZFA and create a CL-to-ZFA mapping set from its xrefs (what is currently done by Uberon);
* use those two sets to create the `mappings.owl` component (instead of using only the FBbt set, as is done now), so that both the FBbt and ZFA mappings are visible within the ontology;
* publish a combined SSSOM set containing *all* the mappings related to CL (what is currently done by Uberon).

On the Uberon side, the changes will be that the Uberon pipeline will no longer have to deal with the ZFA-to-CL mappings; it will only have to fetch the CL mapping set, which will already and by construction contains everything.

[...]

PR #2361, opened 2024-04-26
Add proper mappings management

gouttegd:
This PR implements the CL side of the proposition outline in #2358.

That is, it defines 4 SSSOM mapping sets (declared in the ODK, but as “manually maintained”, so all the code to produce them is in the custom Makefile):

* `fbbt.sssom.tsv`: This is basically a local mirror of the FBbt mapping set, except that it only contains mappings relevant to CL (FBbt-to-CL mappings; all FBbt-to-something-else mappings are removed).
* `zfa.sssom.tsv`: A mapping set obtained by extracting the CL-pointing cross-references in ZFA (ZFA is the source of truth for the CL-ZFA mappings).
* `cl-local.sssom.tsv`: A mapping set obtained by extracting cross-references from the CL edit file, for all foreign ontologies for which CL is the source of truth of mappings (all of them except FBbt and ZFA).
* `cl.sssom.tsv`: A “meta” set made by combining all three sets above. This is the set that will ultimately be published.

The FBbt and ZFA sets are committed to the repository (in `src/mappings` rather than `src/ontology/mappings`, to align CL with the recommendations of the ODK) since they are dependent on remote resources. This is the same logic as the one we use in Uberon. 

The CL-local and CL sets are _not_ committed since they can always be re-generated as needed from files that are present in any copy of the repository.

The FBbt and ZFA sets are also used to populate the `mappings.owl` component, which contains the FBbt and ZFA mappings as old-style cross-references for backwards compatibility (previously only FBbt mappings were present in that component).


src/ontology/Makefile
```diff
@@ -733,6 +739,34 @@ $(PATTERNDIR)/definitions.owl: $(DOSDP_OWL_FILES_DEFAULT)   $(DOSDP_OWL_FILES_CL
...
+
+# This mappingset is manually curated, so we only check that the file actually exists.
+$(MAPPINGDIR)/fbbt.sssom.tsv:
+	test -f $@
+
+# This mappingset is manually curated, so we only check that the file actually exists.
+$(MAPPINGDIR)/zfa.sssom.tsv:
+	test -f $@
+
+validate_mappings:
```
matentzn, line 766: As an aside Seems we should add this to test: make goal

src/ontology/cl-odk.yaml
```diff
@@ -32,6 +33,16 @@ subset_group:
     - id: eye_upper_slim
     - id: general_cell_types_upper_slim
     - id: kidney_upper_slim
+sssom_mappingset_group:
+  products:
+    - id: cl-local
+      maintainance: manual
+    - id: cl
+      maintainance: manual
```
matentzn, line 41: Seems the two mappings sets are not checked in - intended?

gouttegd: > The FBbt and ZFA sets are committed to the repository (in src/mappings rather than src/ontology/mappings, to align CL with the recommendations of the ODK) since they are dependent on remote resources. This is the same logic as the one we use in Uberon.
>
> The CL-local and CL sets are not committed since they can always be re-generated as needed from files that are present in any copy of the repository.

src/ontology/Makefile
gouttegd: Agreed; we can do that when we’ll fix the other SSSOM-related issues in the ODK.

matentzn (approved):

matentzn:
Awwesome, THANKS!

PR #2369, opened 2024-05-09
Refresh imports

aleixpuigb:


aleixpuigb:
This is failing currently because imported terms from Uberon are using the obsoleted go term `reproduction`.

aleixpuigb:
Once it has been fixed in Uberon and we have the new release, I will refresh the imports again.

Issue #2305, opened 2024-03-04
Links to GO terms for types of cilia and villi


dosumis:
Background:  I would like a general push on improving links to cell component terms in GO. I'd like to start with a focus on cleaning up terms for ciliated cell types.

<img width="482" alt="image" src="https://github.com/obophenotype/cell-ontology/assets/112839/5e8d0661-f181-4b03-a8e4-bd802d5608a9">

Issues:

1. The current structure includes cells that are not ciliated, but instead have microvilli.  These are the brush border epithelial cells.  GO has a term for microvillus that we can link these to (unfortunately it does not have a term for brush border, but we could request.
2. Ideally we would distinguish motile from immotile cilia & link to the appropriate term in GO.  I think all of the multiciliated epithelial cells have motile cilia.  Also I believe ependymal cells are well known examples of multicilliated cells  
3. 'single cilliated cell' is a slightly odd term in that I believe most cells have one immotile cilium, which sometimes has a specialised function.  Sperm has one motile cilium so strictly should be under ciliated (classification will be automated by adding has_part 'sperm flagellum'
4. We lack textual definitions for a number of cells including brush border cell and have rather minimal defs for others.

```[tasklist]
### Tasks
- [ ] https://github.com/obophenotype/cell-ontology/pull/2440
- [ ] https://github.com/obophenotype/cell-ontology/pull/2384
- [ ] https://github.com/obophenotype/cell-ontology/pull/2401
- [ ] https://github.com/obophenotype/cell-ontology/issues/2456
- [ ] https://github.com/obophenotype/cell-ontology/issues/2458
- [ ] https://github.com/obophenotype/cell-ontology/issues/2460
```

dosumis:
Hi @Caroline-99 & @aleixpuigb  - I'm co-assigning this one as perhaps not an ideal early ticket.  However, it is a good example of how I would like to clean up links to GO - which will be one of the themes of Caroline's work.

dosumis:
IIRC there is no relationship to villi and the term is defined for
epithelia not epithelial cells.  We could request that GO fix, but will
need a work around if this takes time.

dosumis:
Sorry - ignore last comment.  GO term looks good https://www.ebi.ac.uk/ols4/ontologies/go/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FGO_0005903

PR #2384, opened 2024-06-04
Addresses #2384 Brush Border cell Classifications and Assertions

dosumis:
- Removed incorrect assertion that brush border cells have cilia.
- Added logical def of brush border cell.
- Used a hidden GCI pattern to assert that all brush border cells are columnar/cuboidal.
- brush border cell or proximal tubule is now inferred to be a brush border cell
- Removed assertion that all proximal tubule cells have brush borders because there are rare exceptions.  Manually moved major subclasses under  'brush border cell of proximal tubule. **THIS NEEDS FOLLOW UP**
- Added "brush border" from GO as a bare IRI @anitacaron can you help update imports?


dosumis:
See also #2305

dosumis:
#gogoeditdiff

aleixpuigb:
> * Added "brush border" from GO as a bare IRI @anitacaron can you help update imports?

To [update imports](https://github.com/obophenotype/cell-ontology/blob/master/docs/Adding_classes_from_another_ontology.md) you need to run `sh run.sh make imports/merged_import.owl` in src/ontology.

dosumis:
Still to do: @Caroline-99  to add textual def

anitacaron:
I've updated the imports. @dosumis, could you please update the title to something more specific? 😄

Caroline-99:
> Still to do: @Caroline-99 to add textual def

I have added a textual definition and references

dosumis (commented):
Approved (I can't actually do an official review because I made the PR).

src/ontology/cl-edit.owl
```diff
@@ -5721,9 +5722,13 @@ SubClassOf(obo:CL_0000238 obo:CL_0000240)
 
 # Class: obo:CL_0000239 (brush border epithelial cell)
 
+AnnotationAssertion(obo:IAO_0000115 obo:CL_0000239 "An epithelial cell characterized by the presence of a brush border on its apical surface, which increases the surface area for absorption. These cells are predominantly found lining the small intestine and the proximal tubules of the kidneys.")
+AnnotationAssertion(oboInOwl:hasDbXref obo:CL_0000239 <PMID:25422372>)
```
anitacaron, line 5726: The reference needs to be a string.

anitacaron (changes requested):

src/ontology/cl-edit.owl
```diff
@@ -5721,9 +5722,13 @@ SubClassOf(obo:CL_0000238 obo:CL_0000240)
 
 # Class: obo:CL_0000239 (brush border epithelial cell)
 
+AnnotationAssertion(obo:IAO_0000115 obo:CL_0000239 "An epithelial cell characterized by the presence of a brush border on its apical surface, which increases the surface area for absorption. These cells are predominantly found lining the small intestine and the proximal tubules of the kidneys.")
```
aleixpuigb, line 5725: > These cells are predominantly found lining the small intestine and the proximal tubules of the kidneys.

Since not all cells are found in this epitheliums, consider moving this sentence to the comments.

src/ontology/cl-edit.owl
```diff
@@ -5721,9 +5722,13 @@ SubClassOf(obo:CL_0000238 obo:CL_0000240)
...
+AnnotationAssertion(oboInOwl:hasDbXref obo:CL_0000239 <https://doi.org/10.1152/ajpgi.00005.2011>)
+AnnotationAssertion(oboInOwl:hasDbXref obo:CL_0000239 "PMID:25422372")
+AnnotationAssertion(rdfs:comment obo:CL_0000239 "All brush border cells are columnar-cuboidal.  The formal logical definition of this term asserts this.")
```
aleixpuigb, line 5728: > The formal logical definition of this term asserts this.

Consider removing this sentence.

aleixpuigb (changes requested):
Please, see inline comments
Branch need to be updated.

src/ontology/cl-edit.owl
```diff
@@ -16886,9 +16890,10 @@ SubClassOf(obo:CL_0002306 ObjectSomeValuesFrom(obo:BFO_0000050 obo:UBERON_000413
 AnnotationAssertion(Annotation(oboInOwl:hasDbXref "GOC:tfm") Annotation(oboInOwl:hasDbXref "ISBN:1-4160-2328-3") obo:IAO_0000115 obo:CL_0002307 "A brush border epithelial cell located in the proximal tubule of the kidney that detects fluid flow.")
 AnnotationAssertion(terms:contributor obo:CL_0002307 <https://orcid.org/0000-0003-1980-3228>)
 AnnotationAssertion(oboInOwl:creation_date obo:CL_0002307 "2010-09-14T11:29:20Z"^^xsd:dateTime)
+AnnotationAssertion(rdfs:comment obo:CL_0002307 "The vast majority of proximal tubule epithelial cells have a brush border, but there are rare exceptions.")
```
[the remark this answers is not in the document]
dosumis: We have one rare type of epithelial cell in the proximal tubules that does not have a brush border.  

This causes problems for the other subclass terms we have.  The rearrangment is my attempt at a quick fix.  Another option would be to use the general term for the brush border cells and to come up with a different classification for the exception.  If we do that we probably can't use the logical def pattern 'epithelial cell' and part of some 'proximal tubule'

src/ontology/cl-edit.owl
dosumis: I understand why you'd want to remove.  It's my attempt at leaving a node to explain the hidden GCI pattern in the definition:

EquivalentTo: 'epithelial cell' and has_part some 'brush border'
SubClassOf: columnar/cuboidal epithelial cell

This pattern is justified by that all brush border epiethlial cells are columnar cuboidal.  Use of the pattern means that ontology editors don't need to remember to classify as columnar cuboidal in order to get automated classification.  However because hidden GCIs* are powerful and their effects can only be traced using reasoner explanations, I think they need to be used sparingly and with text that supports/explains them.

I think in this case we should just move the assertion ("all brush border epithelial cells are columnar cuboidal) into the definition.   It is also possible to leave a comment on the subClassOf axiom directly, although that would be non-standard.


\* This is a hidden GCI because it could be rewritten as the GCI:
  'epithelial cell' and has_part some 'brush border' subClassOf columnar/cuboidal epithelial cell

aleixpuigb (approved):

anitacaron (approved):

PR #2419, opened 2024-07-02
Addresses #2354 to create a design pattern for cycling/proliferating cells

### LINKED ISSUE #2354: NTR for cell states from Azimuth and DOSDP to add them
An increasing number of single cell transcriptomics are identifying cell states that are transcriptionally unique. CL editors have decided on the call of November 2023 that these cell states can be added in CL:
image

The cell types can be added using a dessign pattern. I have compiled a [list of terms from Azimuth](https://docs.google.com/spreadsheets/d/1kmw-9wKF7kjdNcfzZGeFRWjvo35bY7rKuEKNXc7i9A4/edit#gid=0) that can be added.
Need:
- Parent cell term
- GO term (object property 'is capable of')
- Create automatic definition: Any "parent term" that is capable of "GO term".
- References

@emquardokus, feel free to add more cell states in the spreadsheet.

aleixpuigb:
Fixes #2354 to create a design pattern for cycling/proliferating cells


docs/editing_guidelines.md
```diff
@@ -178,3 +178,21 @@ The textual definition of a logically defined class should be a plain English eq
 ### Asserting relationships
 
 Please refer to the [relations guide](relations_guide.md) for detailled guidelines about which relations to use for most cases.
+
+## Design Pattern Usage with DOSDP
+
+
+The cell ontology is composed of a large number of terms, classifications and relationships, which continue to increase. As maintaining manually all these classifications and relationships would be an arduous task, a substantial portion of the maintenance is automated. This automation depends, to a significant extent, on the systematic application of design patterns.
+
```
AvolaAmg, line 186: if you want, perhaps try:
The cell ontology contains a large number of terms, classifications, and relationships that are constantly expanding. Manually maintaining all these elements would be a daunting task, so a substantial portion of the maintenance is automated. This automation heavily relies on the systematic use of design patterns.

docs/editing_guidelines.md
```diff
@@ -178,3 +178,21 @@ The textual definition of a logically defined class should be a plain English eq
...
+CL uses **Dead simple OWL design patterns** (DOSDP, [Osumi-Sutherland *et al.*, 2017](https://doi.org/10.1186/s13326-017-0126-0)) to document simple patterns, as they require minimal programming expertise, and once implemented, it is easy to edit.
+
+All patterns are stored in [/src/patterns/dosdp-patterns](https://github.com/obophenotype/cell-ontology/tree/master/src/patterns/dosdp-patterns), while the editable tables are located in [/src/patterns/data/default](https://github.com/obophenotype/cell-ontology/tree/master/src/patterns/data/default).
+
+### Cycling cell states
+Pattern name: cyclingCellStates.yaml
+
+Cell cycling is a fundamental biological mechanism that leads to cell division and duplication into two daughter cells. This process is carried out by a large number of cells, which have a distinct transcriptional profile compared with non cycling cells. There has been an increasing number of single-cell transcriptional datasets that identify cycling cells, which need to be annotated to CL terms.
```
AvolaAmg, line 194: [...] a fundamental biological mechanism that brings a cell to divide and duplicate into two daughter cells. [...], cell undergoing cell cycling have a distinct transcriptional profile compared to non cycling cells. Cycling cells need to be annotated and described by CL terms as these cells have been identified in a multitude of single-cell transcriptional datasets.

AvolaAmg (dismissed):

src/patterns/dosdp-patterns/cyclingCellStates.yaml
```diff
@@ -0,0 +1,64 @@
+pattern_name: cyclingCellStates
+pattern_iri: http://purl.obolibrary.org/obo/cl/cyclingCellStates.yaml
+description: "This pattern is for cells that are cycling/proliferating."
+
+contributors:
+  - https://orcid.org/0000-0001-6677-8489  # Aleix Puig
+
+classes:
+  cell: "CL:0000000"
+  M phase: "GO:0000279"
```
rays22, line 10: I am afraid 'M phase' is too narrow to define cycling/proliferating cells.
Many markers that are widely used to detect cycling/proliferating cells in 'S phase' (*e.g.* [BrdU](https://en.wikipedia.org/wiki/Bromodeoxyuridine) in flow cytometry, histology; anti-[PCNA](https://en.wikipedia.org/wiki/Proliferating_cell_nuclear_antigen); anti-[MCM2](https://en.wikipedia.org/wiki/MCM2), *etc.*). I am not sure if any cell-cycle phase would work for this pattern, because they would be too narrow.
Please, consider
GO:0022402 cell cycle process: [The cellular process that ensures successive accurate and complete genome replication and chromosome segregation.](https://www.ebi.ac.uk/ols4/ontologies/go/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FGO_0022402?lang=en).

rays22 (changes requested):
'M phase' is too narrow to define cycling/proliferating cells.

src/patterns/dosdp-patterns/cyclingCellStates.yaml
aleixpuigb: Thank you for your input @rays22. It was [discussed in the issue](https://github.com/obophenotype/cell-ontology/issues/2354#issuecomment-2075513180) and in a call about which GO term could we use.
That is a good point and I agree that cycling cells can be in 'S phase', but the problem of 'cell cycle process' is that it also involve the G0 transition.
What about [mitotic cell cycle process](http://purl.obolibrary.org/obo/GO_1903047)? It also involves DNA replication: 
> A process that is part of the mitotic cell cycle.

@rays22 @dosumis

rays22: > cycling cells can be in 'S phase', but the problem of 'cell cycle process' is that it also involve the G0 transition.

Cycling/proliferating cells go through **all** phases of the process, so I do not see any contradiction. These subdivisons of the cell cycle are very useful didactically, but they are also somewhat arbitrary and there are so many cell types in various species where some phases are skipped or irregular.

The narrower 'GO:1903047 mitotic cell cycle process' would work for most cell types, except for cells cycling/proliferating meiotically. If you make it clear in the text templates and labels what the pattern applies to (e.g. `mitotically cycling %`), than I am fine with using `mitotic cell cycle process`.

emquardokus: @rays22 @aleixpuigb I think the intended definition of proliferating or cycling cells was the gene expression profile signature derived from single cell RNA sequencing analysis rather than measuring the DNA replication status directly, which is another way to assay it.  
Riba, A., Oravecz, A., Durik, M. et al. Cell cycle gene regulation dynamics revealed by RNA velocity and deep-learning. Nat Commun 13, 2865 (2022). https://doi.org/10.1038/s41467-022-30545-8

rays22: > @rays22 @aleixpuigb I think the intended definition of proliferating or cycling cells was the gene expression profile signature derived from single cell RNA sequencing analysis rather than measuring the DNA replication status directly, which is another way to assay it. Riba, A., Oravecz, A., Durik, M. et al. Cell cycle gene regulation dynamics revealed by RNA velocity and deep-learning. Nat Commun 13, 2865 (2022). https://doi.org/10.1038/s41467-022-30545-8

@emquardokus , Just to clarify, the discussion here is about which specific GO process term to use, not about markers of M phase, S phase. It is about Gene Ontology terms that would fit the best.
The definition of cycling/proliferating cells involves one or more processes that are not specific to single cell RNA sequencing, or any particular assay.
The Gene Ontology already includes several biological process terms related to cycling/proliferation and they all look applicable to single cell expression pattern based categories as well. I can confirm that they also apply to the cell types in the paper you reference.
Using GO component terms in the axioms also ensures interoperability of expression and other types of biological data, such as linking gene expression to genetic perturbations.

anitacaron:
#gogoeditdiff

aleixpuigb:
#gogoeditdiff

dosumis:
Hi @JABelfiore  - I think @aleixpuigb  intends to add these programatically through a pattern - so need to co-ordinate on commits (sorry if this not clear)

aleixpuigb:
Hi @JABelfiore, you can add the information of this cell term in [the TSV file](https://github.com/obophenotype/cell-ontology/blob/1049f1d326fbbb2074191ee390d2dae6413284d4/src/patterns/data/default/cyclingCellStates.tsv). Let me know if you need help. I will undo the changes of your commit.

JABelfiore:
thank you @aleixpuigb I will update the TSV file

dosumis:
#gogoeditdiff

src/patterns/data/default/cyclingCellStates.tsv
```diff
@@ -0,0 +1,23 @@
...
+CL:4033078	cycling mononuclear phagocyte	CL:0000113	PMID:37468583	PMID:37468583	https://orcid.org/0000-0001-6677-8489	2024-07-02T09:14:23Z
+CL:4033079	cycling endothelial cell of lymphatic vessel	CL:0002138	PMID:37291214	PMID:37291214	https://orcid.org/0000-0001-6677-8489	2024-07-02T09:14:23Z
+CL:4033080	cycling pulmonary alveolar type 2 cell	CL:0002063	PMID:37291214	PMID:37291214	https://orcid.org/0000-0001-6677-8489	2024-07-02T09:14:23Z
+CL:4033081	cycling follicular dendritic cell	CL:0000442	PMID:38301653	PMID:38301653	https://orcid.org/0000-0001-6677-8489	2024-07-02T09:14:23Z
+CL:4033082	cycling myeloid cell	CL:0000763	PMID:38301653	PMID:38301653	https://orcid.org/0000-0001-6677-8489	2024-07-02T09:14:23Z
+CL:4033083	cycling naive B cell	CL:0000788	PMID:38301653	PMID:38301653	https://orcid.org/0000-0001-6677-8489	2024-07-02T09:14:23Z
+CL:4033084	cycling basal cell	CL:0000646	PMID:33208946	PMID:33208946	https://orcid.org/0000-0001-6677-8489	2024-07-02T09:14:23Z
+CL:4033085	cycling stromal cell	CL:0000499	PMID: 34497389	PMID: 34497390	https://orcid.org/0009-0005-7919-4905	2024-07-19T09:14:23Z
```
aleixpuigb, line 19: @JABelfiore you need to use CL ID from your range.

aleixpuigb: PMID are wrong, it looks like you scrolled down and it has serialized them instead of copy.

scheuerm:
I don't think we should add "cycling" to cell type classes without a clear need.  In some cases this could result in nonsensical types.  For example, I'm not sure cycling naive B cells exist.

src/patterns/data/default/cyclingCellStates.tsv
```diff
@@ -0,0 +1,21 @@
...
+CL:4033078	cycling mononuclear phagocyte	CL:0000113	PMID:37468583	PMID:37468583	https://orcid.org/0000-0001-6677-8489	2024-07-02T09:14:23Z
+CL:4033079	cycling endothelial cell of lymphatic vessel	CL:0002138	PMID:37291214	PMID:37291214	https://orcid.org/0000-0001-6677-8489	2024-07-02T09:14:23Z
+CL:4033080	cycling pulmonary alveolar type 2 cell	CL:0002063	PMID:37291214	PMID:37291214	https://orcid.org/0000-0001-6677-8489	2024-07-02T09:14:23Z
+CL:4033081	cycling myeloid cell	CL:0000763	PMID:38301653	PMID:38301653	https://orcid.org/0000-0001-6677-8489	2024-07-02T09:14:23Z
+CL:4033082	cycling basal cell	CL:0000646	PMID:33208946	PMID:33208946	https://orcid.org/0000-0001-6677-8489	2024-07-02T09:14:23Z
+CL:4047001	cycling stromal cell	CL:0000499	PMID:34497389	PMID:34497389	https://orcid.org/0009-0005-7919-4905	2024-07-19T09:14:23Z
+CL:4047002	cycling glial cell	CL:0000125	PMID:34497389	PMID:34497389	https://orcid.org/0009-0005-7919-4905	2024-07-19T09:14:23Z
+CL:4047003	cycling plasma cell	CL:0000768	PMID:34497389	PMID:34497389	https://orcid.org/0009-0005-7919-4905	2024-07-19T09:14:23Z
```
rays22, line 19: You might want to refresh `definitions.owl`, because it has the label 'cycling immature basophil' for CL_4047003
which is different form the one here in the dosdp filler table.

>Class: <http://purl.obolibrary.org/obo/CL_4047003> (cycling immature basophil)

src/patterns/data/default/cyclingCellStates.tsv
```diff
@@ -0,0 +1,21 @@
...
+CL:4047004	cycling type EC enteroendocrine cell	CL:0000577	PMID:34497389	PMID:34497389	https://orcid.org/0009-0005-7919-4905	2024-07-19T09:14:23Z
+CL:4047005	cycling neuroblast	CL:0000031	PMID:34497389	PMID:34497389	https://orcid.org/0009-0005-7919-4905	2024-07-19T09:14:23Z
```
rays22, line 21: The label here differs from the one in `definitions.owl` where it is [cycling neuroblast (sensu Vertebrata)](http://purl.obolibrary.org/obo/CL_4047005).

rays22 (changes requested):
The yaml pattern file and the logical definition look fine to me.
I have spot checked some of the newly generated terms in Protege. They look fine, except two labels [cycling immature basophil](http://purl.obolibrary.org/obo/CL_4047003) and [cycling neuroblast (sensu Vertebrata)](http://purl.obolibrary.org/obo/CL_4047005). They disagree with the labels provided in the dosdp data table `cyclingCellStates.tsv`.

src/patterns/data/default/cyclingCellStates.tsv
aleixpuigb: Thanks @rays22! @JABelfiore I will change the label to cycling neuroblast (sensu Vertebrata)

rays22 (approved):

cmungall:
I know this has been merged, but I feel the need to register multiple concerns

The logical definition in the pattern is quite fragile and leans heavily on convention and tacit knowledge. I don't know why we didn't just mint a characteristic (in PATO or here) that says precisely what we mean rather than relying on convention and an unusual reading of a generic GO class, a generic RO property, and a very generic PATO class ("active"). However, it could be argued that so long as there is good discipline and documentation the exact form of the DP doesn't matter.

I also think it would be really useful for the user looking at some of the unusual class structures that will arise from this to have a path back to see the full context about how single cell classification works. This may be obvious to those who attend CL calls but it will not be obvious to most users.

But I have other concerns raised repeatedly in this tracker about the lack of strategy for ensuring subclasses are populated. I have a particularly strong concern here because there are already many types in CL that are *inherently* cycling, these do not have the weaker subclass form of the axiom, and hence we will have many, many errors of omission.

PR #2439, opened 2024-07-17
2397 ns forest markers

dosumis:
Turning this into a draft PR as a place for comments and discussion 

@aleixpuigb  - hope that's OK.


dosumis:
> ERROR:root:Failed to load pattern file: ../patterns/dosdp-patterns/MarkersToCells.yaml

aleixpuigb:
The Marker to cell pattern is not finished, I will leave it in draft mode while I solve the issues.

dosumis:
@hkir-dev - Could you co-ordinate with @aleixpuigb on this.  High priority.  We will also need these patterns in the CL-KG build.  We need something very similar to BDSO.

Main issue:  
- How to efficiently co-ordinate axioms needed for term partly defined by NS-Forest markers & the marker set term.  I guess we used some Python in BDSO.  Might be best to have this here so we're not relying on duplicating content between TSVs.

Specific issues:
1. We need a slot for xrefs.  These should take the DOIs for notebooks detailing the analysis.  Maybe we could add this as an xref on some text in the comment field on the marker set.
2. Type marker set as   "sequence_collection": "SO:0001260"
3. Marker set needs regular var for cell type (range "CL:0000000")
3. We have a new AP for   fbetaConfidenceScore: "STATO:0001000"
4. Patterns have the same pattern_name

hkir-dev:
Recently STATO added a new term for F Beta Score https://github.com/ISA-tools/stato/pull/86#issuecomment-2221606864.

Should we use [STATO:0000663](http://purl.obolibrary.org/obo/STATO_0000663)(F-beta score) instead of `fbetaConfidenceScore: "PCL:0010062"`

src/patterns/dosdp-patterns/NSForestMarkers.yaml
```diff
@@ -0,0 +1,57 @@
...
+  Minimal_markers: "'sequence_feature'"
+
+data_vars:
+  Marker_set_of: "xsd:string"
+  Organ: "xsd:string"
+  Species_abbv: "xsd:string"
+  FBeta_confidence_score: "xsd:double"
+
+annotationProperties:
+  fbetaConfidenceScore: "PCL:0010062"
```
hkir-dev, line 27: Recently STATO added a new term for F Beta Score https://github.com/ISA-tools/stato/pull/86#issuecomment-2221606864.

Should we use [STATO:0000663](http://purl.obolibrary.org/obo/STATO_0000663)(F-beta score) instead of `fbetaConfidenceScore: "PCL:0010062"`

src/patterns/dosdp-patterns/MarkersToCells.yaml
```diff
@@ -0,0 +1,44 @@
+pattern_name: NSForestMarkers
```
hkir-dev, line 1: I don't think this pattern is running at all. I couldn't see any terms generated by this template in the `definitions.owl`. I couldn't see any terms in my local build as well.
Errors are:
- missing column for pattern variable <Marker_set>
- missing column for pattern variable <Marker_set_xref>
- missing column for pattern variable <has_characterization_set>

hkir-dev (changes requested):

src/patterns/dosdp-patterns/NSForestMarkers.yaml
dosumis: yes please

src/patterns/dosdp-patterns/MarkersToCells.yaml
dosumis: Quite likely,  I haven't looked closely yet.  Will try to find time tomorrow.

src/patterns/dosdp-patterns/NSForestMarkers.yaml
```diff
@@ -0,0 +1,57 @@
...
+      - Organ
+
+logical_axioms:
+   - axiom_type: subClassOf
+     multi_clause:
+      sep: " and "
+      clauses:
+        - text: "'has_part' some %s"
+          vars:
+            - Minimal_markers
```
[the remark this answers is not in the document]
dosumis: We will also need an import file for these to get labels. (I think we had ensembl & NCBI for BDSO).  I'm wondering whether, longer term, a more sustainable approach would be to have a separate repo for generating marker objects and make imports from that to the KG and CL.

hkir-dev:
>How to efficiently co-ordinate axioms needed for term partly defined by NS-Forest markers & the marker set term. I guess we used some Python in BDSO. Might be best to have this here so we're not relying on duplicating content between TSVs.

Yes, in BDSO, we have a step before the ODK `prepare_release` that reads the source files and programmatically generates the template data files. I didn't notice any script modifications in the CL repo, which makes me hesitant to add these scripts as it could complicate the maintenance of the CL build pipeline. However, if the number of markers will significantly increase in the future and making programmatic management inevitable, I can propose two solutions:

1. Create a new source file at `src/patterns/data/raw/NSForestMarkersSource.tsv` programmatically populate the contents of `MarkersToCells.tsv` and `NSForestMarkers.tsv` before executing the dosdp templates.
2. Modify the makefile to include a goal similar to the existing [goal to download a dosdp template from a google sheet](https://github.com/obophenotype/cell-ontology/blob/2f18f6a974afb86f0cc01eda761c087669ee1cc9/src/ontology/cl.Makefile#L305) (though this goal seems inactive currently). Similarly, we can manage markers in a separate repo and pull them at runtime.

dillerm:
> Specific issues:
> 
>     1. We need a slot for xrefs.  These should take the DOIs for notebooks detailing the analysis.  Maybe we could add this as an xref on some text in the comment field on the marker set.

Is this referring to GO's hasDbXref annotation? I'm wondering if IAO's 'definition source' (IAO:0000119) might be more appropriate here.

dosumis:
This has been superseded by work on https://github.com/Cellular-Semantics/CellMark/tree/main - which will be a source of imports for for CL as well as a source a source of markers more broadly for https://github.com/Cellular-Semantics/CL_KG

dosumis:
> Is this referring to GO's hasDbXref annotation? I'm wondering if IAO's 'definition source' (IAO:0000119) might be more appropriate here.

Using GO's hasDbXref aligns better with current CL standards.

PR #2487, opened 2024-08-13
Adresses #2261 edit trophoblast cell text definition and hierarchy to describe origin and location

### LINKED ISSUE #2261: [Class hierarchy] Not all trophobast cells are part of blastocyst
**CL term**
[trophoblast cell](http://purl.obolibrary.org/obo/CL_0000351)

**Comments**
There are extravillous trophoblast cells that invade the decidualized endometrium and myometrium or migrate in a retrograde direction along the spiral arteries ([PMID:11337369](https://pubmed.ncbi.nlm.nih.gov/11337369)). Therefore, these trophoblast cells are not part of the blastocyst and the axiom should be changed. The main problem is that these cells are in both fetal and maternal origin structures.
A possible solution might be to add in Uberon `trophoblast' SubclassOf 'has part' some 'trophoblast cell'`.
This cell type differentiates from 'trophectodermal cell'. Therefore, `'develops to' some 'trophectodermal cell'` should be added.

The definition should also be modifies to reflect trophoblast location and functions.

**Current definition**
A cell lining the outside of the blastocyst. After binding to the endometrium, trophoblast cells develop into two distinct layers, an inner layer of mononuclear cytotrophoblast cells and an outer layer of continuous multinuclear cytoplasm, the syncytiotrophoblast cells, which form the early fetal-maternal interface.

**Proposed definition**
An extraembryonic cell that develops from a trophectodermal cell and is part of the outer layer of the blastocyst and can invade other structures in the uterus once the blastocyst implant into the uterine wall. A trophoblast cell is involved in the implantation of the embryo into the uterine wall, placental formation, remodellation of maternal vasculature in the uterus, nutrient and gas exchange, hormone production, and immune modulation to support fetal development.
doi:10.1007/978-3-030-84725-8
PMID:37630754

aleixpuigb:
Fixes #2261 trophoblast cell not only part of blastocyst


aleixpuigb:
#gogoeditdiff

JABelfiore (approved):
Minor edit suggestion for the text definition "An extraembryonic cell that develops from a trophectodermal cell to form part of the outer layer of the blastocyst and that can invade other structures in the uterus once the blastocyst implants into the uterine wall. A trophoblast cell is involved in the implantation of the embryo into the uterine wall, placental formation, remodelling of maternal vasculature in the uterus, nutrient and gas exchange, hormone production, and immune modulation to support fetal development."
The rest looks great!

aleixpuigb:
Thank you for your suggestions! I realised that I shouldn't include "is part of" because can be found in other structures. Same way, trophoblast is a general class and have many child terms. Some of them directly develops from trophoectoderm, and others indirectly, so I prefer to keep both statements separate (develops doesn't imply a direct differentiation process). I have applied the remodeling edit (with one 'l' in american English).

JABelfiore (approved):
Explanation makes sense, updated definition looks good.

aleixpuigb:
> Consider "insert into other structures" rather than "invade other structures"

Thank you for the suggestion. In this case, I prefer to use 'invade into other structures' as it is the terminology that I have seen associated with extravillous trophoblast. Here is an example of Lawless et al., 2023 PMID:[37630754](https://pubmed.ncbi.nlm.nih.gov/37630754)
image

scheuerm:
One of the definitions of invasion is "an unwelcome intrusion into another's domain".  The insertion of these trophoblasts into the uterine wall is essential for life, and so implying that it is unwelcome seems incorrect, regardless of how some in the community have used it.

emquardokus:
@scheuerm @aleixpuigb “trophoblast invasion” is not a fluke usage— if you search pubmed 6483 manuscripts come up using this term. Given the context of trophoblasts invading blood vessels I think this is correct usage here in definition.

dillerm:
This is probably a moot point since the pull request has been merged, but when defining ontology terms it's typically best practice to avoid using jargon. This is even more important when that jargon carries semantic baggage with it, like 'invade' does, that might lead folks who are unfamiliar with this domain to misinterpret what the class represents.

scheuerm:
I accept the argument that this is a term that the community has used for a while.  I just find it curious that a process that is essential for the survival of the species should be described with such a negative connotation.

PR #2491, opened 2024-08-15
 Fixes #2464

### LINKED ISSUE #2464: Make sure all gut epithelial cells are part of epithelium, rather than of a more general organ
Work down from each of the general cell epithelial cells mapped to the GCA hierarchy, making sure each has a part relationships to a term for the epithelium in the appropriate part of the gut.

JABelfiore:
part of relationships updated


dosumis:
@JABelfiore  - Seems the ontology is inconsistent. Let's work through & find the problem in our catchup this week.

dosumis:
@JABelfiore  - I afraid we need to start again.  Too many problems.  Most of them look like this:

image

This says that:
1. Every epithelial cell is a goblet cell
2. Everything that is part of an epithelium is a goblet cell.

So - good we have a reasoner to catch it.

Another example:

image

 Everything that is part of an epithelium is a squamous epithelial cell.  Not sure how that got into an edit on gut cells.

Looking at the diff I can see lots of cases of:

X Equivalent to Y (Everything that is a Y is an X and vice versa) 
and
X Equivalent to part_of some Y (Everything that is part_of some Y is an X)

Too many to fix in Protege.  

I think important to be able to mentally translate OWL expressions into English and ask yourself if they make sense.

PR #2504, opened 2024-08-19
new term immature pericyte added

JABelfiore:
description added, ref added

PR #2505, opened 2024-08-19
new term fetal artery endothelial cell added 

JABelfiore:
definition update, synonyms and refs


src/ontology/cl-edit.owl
```diff
@@ -3058,6 +3058,7 @@ Declaration(Class(obo:CL_4042013))
 Declaration(Class(obo:CL_4042014))
 Declaration(Class(obo:CL_4042015))
 Declaration(Class(obo:CL_4042016))
+Declaration(Class(obo:CL_4047006))
```
aleixpuigb, line 3061: Be careful, [immature pericyte](https://github.com/obophenotype/cell-ontology/pull/2504) has the same ID, this will create conflicts in the future. If you add two cells in two different branches at the same time, you have to manually (in protege) change the IR/ID

JABelfiore: Immature pericyte has been updated to  CL:4047007

aleixpuigb (approved):
I have been wondering if this term should be part_of the [embryonic cardiovascular system](http://purl.obolibrary.org/obo/UBERON_0011695). Since this is a related synonym of fetal circulatory system, I won't request it and approve the PR.

Issue #2297, opened 2024-02-23
[NTR] resident cardiac macrophage


lubianat:
**Preferred term label**
        
*resident cardiac macrophage*

**Synonyms** (add reference(s), please)
cardiac tissue resident macrophage, heart resident macrophage

**Definition** (free text, with reference(s), please. PubMed ID format is PMID:XXXXXX)

A tissue-resident macrophage located in the heart. 

References:
Resident cardiac macrophages: Heterogeneity and function in health and disease

[PMID:36103852](https://pubmed.ncbi.nlm.nih.gov/36103852/)

**Parent cell type term** (check the hierarchy here [Cell Ontology CL](https://www.ebi.ac.uk/ols4/ontologies/cl))

resident macrophage
CL IDs: [CL_0000864](http://purl.obolibrary.org/obo/CL_0000864)

**Anatomical structure where the cell type is found** (check Uberon for anatomical structures: [Uberon Ontology](https://www.ebi.ac.uk/ols4/ontologies/uberon))

heart
Uberon IDs: [UBERON_0000948](http://purl.obolibrary.org/obo/UBERON_0000948)

**Your ORCID**
https://orcid.org/0000-0003-2473-2313

**Additional notes or concerns**
The draft for this request was autogenerated from Wikidata. For more details, see the [Wikidata item](https://www.wikidata.org/wiki/Q104745975) or the [source script](https://www.wikidata.org/wiki/User:TiagoLubiana/cell-ontology.js).

lubianat:
@balhoff  and @cmungall suggested on Slack using KGCL to generate an structured version. Here is a manual version of it based on the [examples](https://github.com/INCATools/kgcl/blob/main/src/data/examples/examples.yaml). 

Naturally, this is missing some bits, namely the definition and its references and the contributor of the term.  

I am also not sure if this is properly formatted, as I never used KGCL before. 

Changes:

```
---
id: CHANGE:001
type: ClassCreation
about_node_representation: curie
node_id: CL:9999999
command_with_curie: create CL:9999999
command_with_uri: create http://purl.obolibrary.org/obo/CL_9999999

---
id: CHANGE:002
type: NodeCreation
about_node: CL:9999999
about_node_representation: curie
node_id: CL:9999999
name: 'resident cardiac macrophage'
command_with_curie: create node CL:9999999 'resident cardiac macrophage'
command_with_uri: TODO

---
id: CHANGE:005
type: EdgeCreation
object_type: curie
[...]

PR #2522, opened 2024-08-25
Add an ID range for KGCL-mediated changes.

gouttegd:
This commit adds a new ID range intended for classes created by Ontobot using the KGCL language.

This will allow editors to command the creation of new classes directly from within a GitHub ticket, without having to explicitly specify the ID of the class to be created:

```
create class AUTOID:1 'my new class'
create edge AUTOID:1 rdfs:subClassOf CL:9999
create exact synonym 'my synonym' for AUTOID:1
```

The ROBOT KGCL plugin would then automatically create the new class using an ID picked from within the "KGCL" range.

### DIFF
--- a/src/ontology/cl-idranges.owl
+++ b/src/ontology/cl-idranges.owl
@@ -486,3 +486,9 @@ Datatype: idrange:71
         allocatedto: "Caroline Eastwood"
     EquivalentTo:
         xsd:integer[>4052000, <= 4059999]
+
+Datatype: idrange:72
+    Annotations:
+        allocatedto: "KGCL"
+    EquivalentTo:
+        xsd:integer[>4060000, <= 4069999]


gouttegd:
This is a follow-up to #2297, in particular this [comment](https://github.com/obophenotype/cell-ontology/issues/2297#issuecomment-1964409107):

> But the main problem is that, as far as I know, we have no mechanisms in KGCL (yet?) to allow a user to create a new class without knowing in advance the ID of the future class.

The KGCL plugin for ROBOT now does have such a mechanism, and having a range named `KGCL` in the range file will enable it.

gouttegd:
> So once the range values are all used up, one would manually move the window, correct?

Yes. But with a range of 10,000 IDs, I’d say it should be quite a long time before we have to worry about that. :)

> Also how are these IDs chosen? First check all the ones in this range used and select the next immediate one available?

Almost. The ID generator will first find the smallest ID in the range that is not already used in the ontology, and then pick a _random_ ID after that.

The randomness bit is to reduce the likelihood of clashing IDs if two (or more) PRs are created by the Ontobot at the same time.

PR #2545, opened 2024-09-06
BEST2+ goblet cell definition 

JABelfiore:
Definition and references added


dosumis (commented):
Having looked more closely at the papers, I'm not sure that we need a separate cell type for BEST2 goblet cells in the colon as it marks all goblet cells in the colon  (unlike BEST4 which marks a distinct subset of enterocytes).  At least that is my reading of https://pubmed.ncbi.nlm.nih.gov/24223998/.   It could be a synonym of goblet cell of colon and could be added as a marker.

JABelfiore:
Closing as agreed we don't need separate cell type for BEST2 goblet cells in the colon as it marks all goblet cells in the colon

PR #2574, opened 2024-09-17
Fixes #2521 logical defs of syncytial cell and subclasses updated 

### LINKED ISSUE #2521: participates_in --> capable_of
image

See https://obophenotype.github.io/cell-ontology/relations_guide/#recording-function

(although in this case it feels like it would be more accurate to have a relation 'participated_in' as not clear that mature syncytia always maintain a potential/capability of participating in fusion).

Caroline-99:
capable of is replaced with output of


Caroline-99:
#gogoeditdiff

Caroline-99:
#gogoeditdiff

dosumis (approved):

PR #2610, opened 2024-09-24
NTR mesenchymal lymphoid tissue organizer cell

JABelfiore:
definition, synonyms and refs added

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -3070,6 +3070,7 @@ Declaration(Class(obo:CL_4047006))
 Declaration(Class(obo:CL_4047007))
 Declaration(Class(obo:CL_4047011))
 Declaration(Class(obo:CL_4047012))
+Declaration(Class(obo:CL_4047022))
 Declaration(Class(obo:CL_4047051))
 Declaration(Class(obo:CL_4047052))
 Declaration(Class(obo:CL_4052001))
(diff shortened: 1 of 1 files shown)


dosumis:
#gogoeditdiff

src/ontology/cl-edit.owl
```diff
@@ -32490,6 +32491,16 @@ AnnotationAssertion(terms:date obo:CL_4047012 "2024-09-06T08:41:39Z"^^xsd:dateTi
 AnnotationAssertion(rdfs:label obo:CL_4047012 "angiogenic pericyte"@en)
 SubClassOf(obo:CL_4047012 obo:CL_0000669)
 
+# Class: obo:CL_4047022 (mesenchymal lymphoid tissue organizer cell)
+
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref "PMID:31201308") Annotation(oboInOwl:hasDbXref "PMID:34497389") obo:IAO_0000115 obo:CL_4047022 "A stromal cell found in the adult gut that plays a crucial role in the formation and maintenance of gut-associated lymphoid tissues (GALTs). This cell is characterized by it's ability to interact with lymphoid tissue inducer (LTi) cells, express lymphotoxin beta receptor (LTβR) and vascular cell adhesion molecule (VCAM), produce chemokines and cytokines for immune cell recruitment and positioning, and contribute to the structural organization of lymphoid tissues within the intestinal environment. mLTo cells are distinguished from other stromal cells by their specific function in orchestrating the development and maintenance of organized lymphoid structures in the gut, which are essential for intestinal immune responses and homeostasis.")
```
dosumis, line 32496: "PMID:31201308" not a relevant paper.  Typo?

dosumis (changes requested):
I think more discussion needed on this one.    Looks like we should have a general cell type + maybe one for fetal gut (based on Elementaite).  Needs ref that supports adult.  I think assertions are complex enough that it would be much better if we could add minirefs in text defs so that users and future editors know exactly which refs to go to to check assertions.

src/ontology/cl-edit.owl
dosumis: Def reads nicely. Ref still needs fixing so can't check ref for how it supports def.  Would like to close this by end of sprint if poss.

dosumis (changes requested):

dosumis:
Still needs some discussion but please fix PMIDs ASAP.

JABelfiore:
Iceboxed as still need discussion

Issue #2637, opened 2024-10-10
Generation of DOSDP matches is broken


gouttegd:
Trying to find matches for the DOSDP patterns yields the following error when processing the `cellPartOfAnatomicalEntity` pattern:

```
dosdp-tools query --ontology=tmp/edit-merged.owl --catalog=catalog-v001.xml --reasoner=elk --obo-prefixes=true --batch-patterns="cellBearerOfQuality cellCapableOfBiologicalProcess cellHasPlasmaMembranePartX cellPartOfAnatomicalEntity cyclingCellStates taxonSpecific" --template="../patterns/dosdp-patterns" --outfile="../patterns/data/matches/"
log4j:WARN No appenders could be found for logger (org.semanticweb.elk.config.ConfigurationFactory).
log4j:WARN Please initialize the log4j system properly.
2024-10-10 19:21:24.232Z error [] [context: command=query;pattern=../patterns/dosdp-patterns/cellPartOfAnatomicalEntity.yaml;output=../patterns/data/matches//cellPartOfAnatomicalEntity.tsv] Failed to parse class expression: :
Encountered |EOF| at line 1 column 1. Expected one of:
	Class name
	Object property name
	Data property name
	inverse
	not
	(
	{

2024-10-10 19:21:24.331Z error [] [context: ] Failed to parse class expression: 
make: *** [cl.Makefile:194: matches] Error 1
```

Unclear what the cause of the error is since the error message is not really helpful (it does not give the actual expression it is failing to parse). AFAICT I don’t see anything obviously wrong in the `cellPartOfAnatomicalEntity` pattern, and neither does `dosdp validate`.

balhoff:
Try removing this line: https://github.com/obophenotype/cell-ontology/blob/7abc8231c7a3181cd89b3f380710b781913534fe/src/patterns/dosdp-patterns/cellPartOfAnatomicalEntity.yaml#L31

gouttegd:
Thanks, that was it indeed.

PR #2645, opened 2024-10-11
Cleanup custom makefile

gouttegd:
This PR cleans up the custom `cl.Makefile`. Basically, it does for CL what was done for Uberon [here](https://github.com/obophenotype/uberon/pull/3028).

The most important cleanup is the complete removal of all obsolete rules and their associated comments. Keeping deprecated rules, even if they are not used, is _not_ harmless as they make the Makefile needlessly more difficult to read and understand, and therefore more difficult to maintain.

In particular, anyone reading the current Makefile could reasonably conclude, based on those rules and their comments, that ROBOT is utterly broken and that CL has no choice but to keep doing its releases with OWLTools’s Ontology-Release-Runner (OORT) -- that may have been true in the past, but it has not been true for a long time now.

Along the way, two DOSDP patterns are updated to fix an error causing the search for matches to fail (#2637).

Other changes are purely cosmetic: rules are reorganised for clarity into clearly labelled sections, re-formatted for readability, and standard ODK variables are used whenever possible. Comments are added and, when appropriate, links to relevant tickets are included.

Overall, the PR does not change _anything_ in the generated ontological contents.


matentzn (approved):
I have gone through all of your commits, and everything looks great. A few `FIXMEs` but since they are all coupled with issues, I think this is good to go!

Awesome work, thanks.

matentzn (approved):

PR #2664, opened 2024-10-16
Fixes #2651 NTR diffuse bipolar 5 cell

### LINKED ISSUE #2651: NTR diffuse bipolar 5 cell
**Preferred term label**
NTR diffuse bipolar 5 cell

DB5 cell
PMID:32032773

An ON diffuse bipolar cell that stratifies in the stratum 3 of the inner plexiform layer to ON parasol ganglion cells. This cell type can be transcriptomically distinguished from other bipolar cells by the expression of MYO16.
PMID:38014002
PMID:27833534
PMID:32032773

[ON-bipolar cell](http://purl.obolibrary.org/obo/CL_0000749)

retina

**Your ORCID**
https://orcid.org/0000-0001-6677-8489

A diffuse bipolar 5 cell typically contact 7-8 cone photoreceptors. The number of dendritic processes and thickness of the dendrites is intermediate between a diffuse bipolar 6 cell (less dense and thin) and diffuse bipolar 4 cell (more dense and thick). [PMID:27833534](https://pubmed.ncbi.nlm.nih.gov/27833534)

aleixpuigb:
Fixes #2651 NTR diffuse bipolar 5 cell


aleixpuigb:
MYO16 marker can be found in the article supplementary figure:
image

scheuerm:
I think we need to have a discussion of what we mean by a marker gene.  In the figure above, MYO16 is also expressed at lower levels in a few other bipolar cells.  But in the image below, it is also highly expressed in several amacrine cell types in the Li et al. dataset.  So what would we assert in this case?
Screenshot 2024-10-16 at 6 03 59 PM

aleixpuigb:
Thank you @scheuerm.
When I was curating these cells, I realised that MYO16 was highly expressed in non bipolar cells, and that is the reason I wanted to make the distinction only with other bipolar cells:
> "This cell type can be transcriptomically distinguished from other bipolar cells by the expression of MYO16."

I am not sure if there is a published gene or gene set that would be good to distinguish for all other cell types in the retina (if we had markers identified by NS-Forest would be great!). It is true that it is expressed at lower levels in other bipolar cells, and therefore it is far from perfect.
The reason that I used this gene was because the Single Cell Atlas of the Human Retina are using MYO16 as a canonical marker for this cell type. However, I can remove this sentence if it is conflictive and once we have a more specific gene set we can add them through the pipeline that is being developed. They also provide a gene set for DB5 that is more enriched, but it is also enriched at lower levels in other cell types:
image

dosumis:
@aleixpuigb Let's discuss.  I think we shouldn't just take "cannonical marker" lists to include in term definitions.  They are more to do with evidence for identity - linking back to earlier studies, often pre-transcriptomics.

AvolaAmg:
#gogoeditdiff

scheuerm:
@aleixpuigb , COL19A1 is one of the combo markers selected by NS-Forest.  Here are the data for the combo:

image

The F-beta score of 0.91 is very high, but as with most markers in these single cell datasets, false negative calls result in lower recall.  But COL19A1 would be much better than MYO16.

src/ontology/cl-edit.owl
```diff
@@ -32136,6 +32137,16 @@ SubClassOf(obo:CL_4033084 obo:CL_0000501)
...
+AnnotationAssertion(terms:contributor obo:CL_4033085 <https://orcid.org/0000-0001-6677-8489>)
+AnnotationAssertion(terms:date obo:CL_4033085 "2024-10-16T12:46:53Z"^^xsd:dateTime)
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref "PMID:32032773") Annotation(oboInOwl:hasSynonymType obo:OMO_0003000) oboInOwl:hasRelatedSynonym obo:CL_4033085 "DB5 cell")
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref "PMID:27833534") rdfs:comment obo:CL_4033085 "A diffuse bipolar 5 cell typically contacts 7–8 cone photoreceptors. The number and thickness of its dendritic processes are intermediate between those of a diffuse bipolar 6 cell (less dense and thinner) and a diffuse bipolar 4 cell (denser and thicker).")
+AnnotationAssertion(rdfs:label obo:CL_4033085 "diffuse bipolar 5 cell")
+SubClassOf(obo:CL_4033085 obo:CL_0000749)
+
```
AvolaAmg, line 32149: PMID: 29114208
This reference might have some extra synonyms such as Type 5, T5 but unsure if you want to associate to diffuse bipolar cells, these are referred to as bipolar cells. 
also it subdivide them further into type 5 inner (5i), outer (5o) and thick (5t) types.
just a suggestion.

aleixpuigb:
After discussing with the other CL ontology editors, we agreed that having the NS-Forest markers would be ideal, but we need a published reference to use them. Once the data for all retinal cells is available, we can include them all. Temporally, to have the term available for annotation, we will have in the definition MYO16 as a marker to distinguish ONLY with other bipolar cells. Thank you for all your feedback @scheuerm

src/ontology/cl-edit.owl
aleixpuigb: Thank you for the suggestion. These synonyms are for the mouse type 5 bipolar cell, and the subclasses are orthologs of DB5, DB4a and DB4b. However, I will look into adding them in a separate ticket.

AvolaAmg: Thank you!

PR #2667, opened 2024-10-17
Update  monthly call times in README

aleixpuigb:
Should we maintain the Uberon calls in the CL README?

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,8 +42,8 @@ Training materials from the 2020 CL Training Workshop are available at https://g
 
 ## Twice monthly calls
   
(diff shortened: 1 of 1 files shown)


gouttegd (approved):
No strong opinion as to whether the schedule should appear directly on the README.

I would be _slightly_ inclined to remove it since we already have a link to the agenda, and the agenda includes the scheduled times. That would mean that, if we change the schedule, there would be only one place to update.

On the other hand, not having to open the GoogleDoc document just to see the schedule is nice.

In any case, as long as we do mention the schedule on the README it obviously has to be correct, so I approve here.

PR #2703, opened 2024-10-29
Break down SubClassOf axioms with an intersection as the superclass.

### LINKED ISSUE #2692: CL:4042013 missing in graph
CL:4042013

Lamp5 Lhx6 neuron (CL:4042013) | subClassOf (rdfs:subClassOf) | lamp5 GABAergic cortical interneuron (CL:4023011)

Hello Cell Ontology Team!

We have a general ontology parser that loads ontologies in memory. The program is attempting to build parents/children and is failing on term CL:4042013. Upon another inspecting of the cl-base.json file there is no connection with this term in the graph. Can you please ensure that the above sub class prop exists?

gouttegd:
We have a handful of classes that are at least partially defined with SubClassOf axioms where the superclass is an object intersection, as in the following example:

>  CL:0000104 **SubClassOf:** CL:0000540 _and_ (RO:0000053 _some_ PATO:0070026)

This is logically equivalent to individual SubClassOf axioms for each member of the intersection:

>  CL:0000104 **SubClassOf:** CL:0000540
>  CL:0000104 **SubClassOf:** RO:0000053 _some_ PATO:0070026

but the first form may not always be handled correctly and is therefore best avoided. In particular, the OboGraphs JSON converter does not know how to handle those axioms, and simply drops them, resulting in those classifications being absent from the JSON artifacts.

So here, we replace known cases of SubClassOf axioms where the superclass is an object intersection (7 found) by equivalent individual SubClassOf axioms.

closes #2692

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -4577,7 +4577,8 @@ AnnotationAssertion(Annotation(oboInOwl:hasDbXref "FMA:67287") Annotation(oboInO
 AnnotationAssertion(oboInOwl:hasDbXref obo:CL_0000104 "FMA:67287")
 AnnotationAssertion(oboInOwl:inSubset obo:CL_0000104 cl:BDS_subset)
 AnnotationAssertion(rdfs:label obo:CL_0000104 "multipolar neuron")
-SubClassOf(obo:CL_0000104 ObjectIntersectionOf(obo:CL_0000540 ObjectSomeValuesFrom(obo:RO_0000053 obo:PATO_0070026)))
+SubClassOf(obo:CL_0000104 obo:CL_0000540)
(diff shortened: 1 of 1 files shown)


dosumis (approved):
Thanks for this.  These cause problems with other pipelines.

Longer term solutions:
 - Add a validation check for these (Reasonable, but a bit annoying as the difference is it means semantically correct files fail)
 - Extend ROBOT relax to split SubClassOf intersections into separate SubClassOf axioms

gouttegd:
> Extend ROBOT relax to split SubClassOf intersections into separate SubClassOf axioms


ROBOT’s `relax` command already has an [option to do that](https://robot.obolibrary.org/relax#include-subclass-of) (`--include-subclass-of`). :) It’s just not enabled by default.

Arguably the standard ODK pipelines could enable that option.

PR #2711, opened 2024-10-30
Fixes #2707 - Logical and textual defs acinar cell of sebaceous gland

### LINKED ISSUE #2707: [Logical def] acinar cell of sebaceous gland is classified under epithelial cell of sweat gland
**CL term**
acinar cell of sebaceous gland 

**Current classification**
image

**Suggested revision of logical definition**

image

Caroline-99:
- Fixed the logical defintion
- Improved textual defs to align with the logical definition and provided references
- Relabelled to sebocyte and skin sebocyte.
- Updated logical and textual definitions of sebocyte


dosumis:
#gogoeditdiff

dosumis:
Looks like link to sebum secretion is lost.  Ideally we'd have a link to a GO sebum secretion term.  In the absence of that we may need to manually classify under sebum secreting cell.

Caroline-99:
> Looks like link to sebum secretion is lost. Ideally we'd have a link to a GO sebum secretion term. In the absence of that we may need to manually classify under sebum secreting cell.

I agree. It is currently manually classified under sebum secreting cell.

dosumis:
Went down a bit of a rabbit hole on this one.  As far as I can tell, there is one type of cell - sebocyte - that corresponds to 'sebum secreting cell' and 'acinar cell of sebaceous gland'.  Both have sebocyte as as synomym and the former has the logical def: 
image

"Secretion" involves the cells dying and releasing their contents. 

image

So - I think the two terms should be merged.   

Longer term we should make GO terms to define these:
We should request holocrine secretion as a GO term (https://en.wikipedia.org/wiki/Holocrine) as a subclass of apoptosis, and maybe specific terms for holocrine secretion of sebum and of mebum.

aleixpuigb:
> So - I think the two terms should be merged.

According to the definition and the logical definition, it looks like 'acinar cell of sebaceous gland' is a narrower term specific for skin. 'sebocyte' is a broad synonym of it. While the majority of sebaceus glands are found in the skin, they can be found in other tissues. Perhaps a relabel would be appropriate in this case?
image


> Longer term we should make GO terms to define these:
We should request holocrine secretion as a GO term (https://en.wikipedia.org/wiki/Holocrine) as a subclass of apoptosis, and maybe specific terms for holocrine secretion of sebum and of mebum.

I agree

dosumis:
> Perhaps a relabel would be appropriate in this case?

If we want to keep two terms with different anatomical scope, we should definitely relabel.  Maybe "sebocyte" and  "skin sebocyte" ?

Caroline-99:
#gogoeditdiff

Caroline-99:
> > Perhaps a relabel would be appropriate in this case?
> 
> If we want to keep two terms with different anatomical scope, we should definitely relabel. Maybe "sebocyte" and "skin sebocyte" ?

I have relabelled to sebocyte and skin sebocytes, and updated the textual and logical definition of sebocyte.

aleixpuigb (approved):

PR #2724, opened 2024-11-01
Compute diff on -base rather than -simple.

### LINKED ISSUE #1737: Update diff to use base plus with new ODK
Reminder ticket to update automated diff to use base plus once that is ready with next ODK release

gouttegd:
When computing the automated diff on pull requests, for the "classification diff" (diff on the classified ontology), we use the cl-base.owl product rather than the cl-simple.owl product.

Until now, we were using -simple as it was the only product that was (1) reasoned over and (2) stripped of all "foreign" classes.

Since ODK 1.5, cl-base also fits those criteria, and it is in fact more sensible to run the diff on that product since it is intended to be the main product that other ontologies are supposed to be importing.

We also bump the ODK version that is being used to compute the diffs.

closes #1737

### DIFF
--- a/.github/workflows/diff.yml
+++ b/.github/workflows/diff.yml
@@ -53,7 +53,7 @@ jobs:
     needs: [branch_status]
     if: ${{ github.event.issue.pull_request }}
     runs-on: ubuntu-latest
-    container: obolibrary/odkfull:v1.5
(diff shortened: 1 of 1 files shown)


matentzn (approved):
No doubt! Thanks!

PR #2746, opened 2024-11-11
Remove hasAlternativeID for FMA:83808.

### LINKED ISSUE #1857: Tech request: audit for has_alternative_ID
Related to #1851 (especially [this comment](https://github.com/obophenotype/cell-ontology/issues/1851#issuecomment-1443386690)):

This ticket is a request for the Tech Group to audit CL to ensure each has_alternative_ID is converted to be compatible with the latest system.

@dosumis, kindly add more context if needed.

gouttegd:
CL term 'cardiac muscle cell' (CL:0000746) carries a `oboInOwl:hasAlternativeId` annotation to FMA:83808.

In effect, because of the way the `oboInOwl:hasAlternativeId` annotation is interpreted, this means that CL is declaring that FMA:83808 is deprecated in favour of CL:0000746, and is therefore akin to an axiom injection from CL into FMA.

Issue #1857 is asking that all `oboInOwl:hasAlternativeId` annotations be converted into the modern way of representing deprecation replacement (by a combination of `owl:deprecated` and `IAO:0100001` annotations). But in this instance (which is the last instance of oboInOwl:hasAlternativeId in CL), because the (supposedly) deprecated term is in FMA, it is up to FMA to act. On the CL side, the correct thing to do is simply to remove the annotation, which is what we do here.

closes #1857

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -3335,7 +3335,6 @@ Declaration(AnnotationProperty(oboInOwl:SubsetProperty))
 Declaration(AnnotationProperty(oboInOwl:SynonymTypeProperty))
 Declaration(AnnotationProperty(oboInOwl:consider))
 Declaration(AnnotationProperty(oboInOwl:creation_date))
-Declaration(AnnotationProperty(oboInOwl:hasAlternativeId))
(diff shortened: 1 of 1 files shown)


matentzn (approved):
Agreed.

PR #2748, opened 2024-11-11
Update foreign SSSOM set from ontology

gouttegd:
This PR amends the rules that refresh the externally maintained SSSOM mapping sets (the sets from FBbt and ZFA) to include a step that check those sets against the current version of the ontology to

(1) remove any mapping to a Uberon entity that does not exist or is deprecated;
(2) add any missing label, using the labels found in the ontology.

This necessitates, as a preliminary:

(1) making sure we are using ODK 1.5.3, as previous versions do not include a recent enough version of SSSOM-Java;
(2) removing a bogus empty xml:base attribute in the XML catalog, that prevents sssom-cli from reading it.

### DIFF
--- a/.github/workflows/qc.yml
+++ b/.github/workflows/qc.yml
@@ -19,7 +19,7 @@ jobs:
   ontology_qc:
     # The type of runner that the job will run on
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


matentzn (approved):
nice

PR #2750, opened 2024-11-12
2464 gut epithelial cells

JABelfiore:
make sure all gut epithelial cells are part of epithelium rather than of a more general organ


dosumis:
#gogoeditdiff

dosumis:
HI @JABelfiore  - looks like we have two terms that are now equivalent to each other:

2024-11-12 12:12:32,870 ERROR org.obolibrary.robot.ReasonOperation - Equivalence: <http://purl.obolibrary.org/obo/CL_0002655> == <http://purl.obolibrary.org/obo/CL_0002654>
make: *** [Makefile:105: reason_test] Error 1

Could you post an explanation from Protege here?

JABelfiore:
> HI @JABelfiore - looks like we have two terms that are now equivalent to each other:
> 
> 2024-11-12 12:12:32,870 ERROR org.obolibrary.robot.ReasonOperation - Equivalence: http://purl.obolibrary.org/obo/CL_0002655 == http://purl.obolibrary.org/obo/CL_0002654 make: *** [Makefile:105: reason_test] Error 1
> 
> Could you post an explanation from Protege here?

Both of these terms have the logical definition: 

'epithelial cell of esophagus'
 and ('part of' some 'epithelium of esophagus')

To differentiate further, I need to add two new uberon terms for 'stratum spinosum of esophageal epithelium' and 'stratum corneum of esophageal epithelium' which I will create tickets for now.

dosumis:
Solutions sounds good to me. Maybe to move things forward you could just switch these axioms to subClassOf for now.  This is not wrong, just a bit vague.  With that in place we should be able to run gogoeditdiff which will make reviewing the rest much easier.

BTW - I went down a bit of a rabbit hole looking at this.  Adding details here, bit we need to turn into a separate (epic) ticket rather than fix on this PR:

Having looked into this issue a bit further, it looks like the stratum spinosum cells in the esophagus are a type of prickle cell. https://www.perplexity.ai/search/does-the-epithelium-lining-the-1KoagdP.SXeELKdsODcikw
https://pubmed.ncbi.nlm.nih.gov/416593/
https://pubmed.ncbi.nlm.nih.gov/745022/
https://pubmed.ncbi.nlm.nih.gov/7959182/

And that the epithelial cells of the esophagus are keratinocytes (they produce different keratins from the skin and so the epithelium is not considered keratinsed).  The same applies to the epithelium of the oral mucosa.

However, we also have 
image

This term probably corresponds to the stratum corneum cells in the esophagus.

TBD: should we broaden keratinocyte to cover epithelial cells of oral mucosa and esophagus?

'epithelial cell of stratum spinosum of esophagus' SubClassOf 'prickle cell' (But need to make sure prickle cell not limited to skin.

To fix:  esophogeal epithelium subClassOf stratified squamous epithelium (http://purl.obolibrary.org/obo/UBERON_0006915)

AvolaAmg (approved):

Issue #2644, opened 2024-10-11
Better management of remote resources


gouttegd:
Beyond the ontologies it imports, CL is making use of two remote resources:

* the “HRA subset”, maintained in [hubmapconsortium/ccf-validation-tools](https://github.com/hubmapconsortium/ccf-validation-tools) and downloaded from [here](https://raw.githubusercontent.com/hubmapconsortium/ccf-validation-tools/master/owl/CL_ASCTB_subset.owl) into `src/ontology/components/hra_subset.owl`;
* the “Cell×Gene reference subset”, maintained in [Cellular-Semantics/cellxgene-cell-reporter](https://github.com/Cellular-Semantics/cellxgene-cell-reporter) and downloaded from [here](https://raw.githubusercontent.com/hkir-dev/cellxgene-cell-reporter/main/templates/cellxgene_subset.tsv) into `src/templates/cellxgene_subset.tsv`.

The HRA subset is automatically and forcefully downloaded anew as part of the release pipeline. This causes two problems: (1) it introduces a risk that a release (or QC) pipeline could fail solely because of a network issue, and (2) any changes made to `src/ontology/components/hra_subset.owl` is always ignored since the file is completely re-downloaded from its original location. If editors are not aware of (2), they could make changes to this file thinking that it is the correct thing to do, not knowing that their changes are ignored (this is exactly what happened in #2472).

As for the Cell×Gene subset, it is _never_ automatically refreshed, and in fact can only be refreshed if an editor forcefully deletes the local file -- only then will the file be downloaded again from its original location. This raises the question of whether that subset is intended to be maintained by Cellular-Semantics/cellxgene-cell-reporter or by CL: if the former, then we need a simple mechanism to update it (requiring editors to remove the file is not a satisfying solution); if the latter, then the rule in `cl.Makefile` to download the file from its remote location is useless.

Overall, CL should adopt for its external resources the same approach as Uberon: clearly uncouple the refreshing of external resources from the release/QC pipelines. That is, we make the downloading of external resources strictly conditioned to `IMP=true` (as for imports). When `IMP=false` (as is the case when running QC and release), only the locally committed files are used, and we make no attempt to refresh them. This both prevents any build failure due to a network problem, and allows local fixes (aka hacks) to these external resources if needed (best avoided, but sometimes necessary until we can get a fix upstream).

PR #2763, opened 2024-11-15
Fixes #2725 clm-cl.owl component edits

### LINKED ISSUE #2725: Add cellmark CL release artefact to CL components
This contains marker content approved for CL:

https://github.com/Cellular-Semantics/CellMark/blob/main/clm-cl.owl

I think it makes most sense to incorporate as a component as we need the whole thing rather than just a module.

- [ ] Add step to makefile to pull content into components folder
- [ ] Add import statement to cl-edit.owl

I think that's it/

Caroline-99:
clm-cl.owl component edits - This work was done under the supervision of @hkir-dev 

- [ ] Add step to makefile to pull content into components folder
- [ ]  Add import statement to cl-edit.owl

blocked by https://github.com/Cellular-Semantics/CellMark/issues/9


dosumis:
Looks good, but maybe some unrelated edits here too - e.g. on taxon constraints?

gouttegd (changes requested):
On the addition of the `clm-cl.owl` component itself, no problem except that at some point we will need to make the downloading of the component conditional, so that it is only refreshed when a refresh is explicitly required and not when running the QC pipelines (#2644). But this is out of scope of this PR.

The real issue here is the overwriting of our custom QC workflow file (`.github/workflows/qc.yml`, which must be reverted.

gouttegd:
> The real issue here is the overwriting of our custom QC workflow file (.github/workflows/qc.yml), which must be reverted.

Unfortunately it won’t be practical to do this starting from this PR, since the changes to `qc.yml` have been committed as part of a single commit with all the other changes.

Better to actually start anew in a completely distinct PR. I can take care of that.

PR #2764, opened 2024-11-15
Gut images test - epithelial cell of stomach

JABelfiore:
layers of GI tract image added


dosumis:
Don't see image files.  cl-edit.owl looks like it might be out of date?

PR #2772, opened 2024-11-18
2725 Add CellMark CLM-CL component

gouttegd:
This PR supersedes #2763. It adds the CLM-CL component, but does so while avoiding to overwrite CL’s customised QC workflow.


gouttegd:
Draft PR as in its current state, the CLM-CL component would cause a build failure because of duplicated labels in CLM-CL (https://github.com/Cellular-Semantics/CellMark/issues/9).

(This is exactly why it is important to uncouple the refreshing of external resources from the QC/release workflows: when external resources are unconditionally refreshed as part of all workflows, as is currently the case, we can end up with build failures _at any time_ because of changes that have happened elsewhere and are totally out of CL’s control. #2644)

gouttegd:
PR updated to only refresh the CLM-CL components (as well as all other components) when `MIR` is true (e.g. when imports are updated with `make refresh-imports`).

hkir-dev (approved):

PR #2777, opened 2024-11-19
Fixes #2774 - Added GO links to deuterosomal cell

### LINKED ISSUE #2774: Link deuterosomal cell to centriole replication
Based on text of def this is valid:

An epithelial cell part of respiratory tract epithelium that is a precursor of a multi-ciliated cell. This cell actively amplifies centrioles, a required step for multiciliogenesis.

\+
SubClassOf capable_of some centriole replication (GO:0007099)

Caroline-99:
Imported and added a GO term to deuterosomal cell


dosumis (dismissed):
Looks good.  Any idea why the mapping files also changed?

dosumis:
GO update seems to have pulled in an obsolete term 

> label: hepatocyte
> relationship: capable_of GO:0044237 ! obsolete cellular metabolic process

This is way too general to be useful so I'd just delete.  (I'm sure there are many more specific metabolic processes we could link to, but that may be a bit of a rabbit hole)

Caroline-99:
> GO update seems to have pulled in an obsolete term
> 
> > label: hepatocyte
> > relationship: capable_of GO:0044237 ! obsolete cellular metabolic process
> 
> This is way too general to be useful so I'd just delete. (I'm sure there are many more specific metabolic processes we could link to, but that may be a bit of a rabbit hole)

will delete it now

dosumis (approved):
Edits look good.

May not be an issue, but still wondering why there is an update to mapping file format @gouttegd  - is this from some update of yours?  Could it be coupled to running the import update?

Issue #2785, opened 2024-11-20
Automatically pull in HuBMAP 2D images


dosumis:
Example image here:

https://humanatlas.io/2d-ftu-illustrations?version=2.1&organ=Liver

Includes pub DOI +  mapping table to Uberon and CL

Query here: 

https://apps.humanatlas.io/api/grlc/#get-/hra/ftu-parts

Response
ftu_digital_object,image_url,organ_iri,ftu_iri,ftu_part_iri
https://purl.humanatlas.io/2d-ftu/liver-liver-lobule,https://cdn.humanatlas.io/digital-objects/2d-ftu/liver-liver-lobule/v1.2/2d-ftu-liver-liver-lobule.png,http://purl.obolibrary.org/obo/UBERON_0002107,http://purl.obolibrary.org/obo/UBERON_0004647,http://purl.obolibrary.org/obo/CL_1000398

We also need a way to get DOIs @bherr2 - can you let us know how to do this?

columns:
 - CL ID (CURIE)
 - FOAF:Depiction --> image PURL
 - >> license
 - >> reference
 - >> comment (will be blank when autogenerated)

- [x] Write Python script to generate a ROBOT template from the output of this query - folding in DOIs.  STATUS: DRAFT
     -  Python script should update a single existing template which is also used for manual addition
     - TBD: populate sheet with existing image links?  Could retrieve via SPARQL?
- [ ] Incorporate template into automatic build
- [ ] Allow for manual addition of comments mapping annotations to cell types where annotation labels are different.

PR #2805, opened 2024-11-26
Addresses #2785 Automatically pull in HuBMAP 2D images

aleixpuigb:
Addresses #2785 Automatically pull in HuBMAP 2D images

- [x]  Write Python script to generate a ROBOT template from the output of this query - folding in DOIs.
- [x]  Incorporate template into automatic build
- [ ]  Allow for manual addition of comments mapping annotations to cell types where annotation labels are different.
- [x] Create a list of reviewed object to not be added in the list.


dosumis:
Looks good.  Could also potentially use GRLC.

aleixpuigb:
> Looks good. Could also potentially use GRLC.

This comment comes from my ignorance, but what is the advantage of using GRLC? Thanks!

dosumis:
> This comment comes from my ignorance, but what is the advantage of using GRLC? Thanks!

GRLC endpoint allows a simple REST API query rather than having the whole SPARQL query in your code.  It's just simpler and we rely on them to write the SPARQL

dosumis:
Missing ROBOT template strings row: https://robot.obolibrary.org/template#class-template-strings.  Add annotation axiom on Class to link to image.  Then annotate this with further annotation axioms for license, ref etc.  (>A)

dosumis:
@aleixpuigb  - what's holding this back?  IRRC you have some local code that at leasr partially fixes/ Could we have a quick hacking session to get it over the line?

aleixpuigb:
> Hard to judge the extent of this from the diffs or viewing table on PR as no labels. Would you be able to generate a quick report of labels of CL terms used in annotation?

In this [SPARQL query](https://api.triplydb.com/s/OoC8sNz2y) you can see the 123 terms.

> I think that would make sense. There's not a massive number so curating each one for improved CL terms could be viable as long as only needs to be done once. Would be easier if template table had an unused column with CL term name in it (I think that's possible).

I can use oaklib to pull the labels. I think if the second row is blank for a certain column, that column is unused.

aleixpuigb:
@Caroline-99 please review that all terms in the table are adequate to incorporate the images.
@hkir-dev Can you review the script and the generation of the owl file
@gouttegd To merge the master branch, I had to solve conflicts in the Makefile (not cl.Makefile) which should not be edit. Let me know if I should Revert commit [8ab6e4e](https://github.com/obophenotype/cell-ontology/pull/2805/commits/8ab6e4ee4fc0d4d2806b3565e73aca179de99416) if I have done something wrong. The majority of edits were easy, only I was not sure was the fingerprint that I left the incoming from Master. Again, I know the file should not be edited, so I don't have a problem reverting and solve conflicts in a different way. Thank you!

src/ontology/Makefile
```diff
@@ -57,7 +57,7 @@ OBODATE ?=                  $(shell date +'%d:%m:%Y %H:%M')
 VERSION=                    $(TODAY)
 ANNOTATE_ONTOLOGY_VERSION = annotate -V $(ONTBASE)/releases/$(VERSION)/$@ --annotation owl:versionInfo $(VERSION)
 ANNOTATE_CONVERT_FILE =     annotate --ontology-iri $(ONTBASE)/$@ $(ANNOTATE_ONTOLOGY_VERSION) convert -f ofn --output $@.tmp.owl && mv $@.tmp.owl $@
-OTHER_SRC =                 $(PATTERNDIR)/definitions.owl $(COMPONENTSDIR)/hra_subset.owl $(COMPONENTSDIR)/mappings.owl $(COMPONENTSDIR)/blood_and_immune_upper_slim.owl $(COMPONENTSDIR)/eye_upper_slim.owl $(COMPONENTSDIR)/general_cell_types_upper_slim.owl $(COMPONENTSDIR)/kidney_upper_slim.owl $(COMPONENTSDIR)/cellxgene_subset.owl $(COMPONENTSDIR)/clm-cl.owl $(COMPONENTSDIR)/wmbo-cl-comp.owl 
+OTHER_SRC =                 $(PATTERNDIR)/definitions.owl $(COMPONENTSDIR)/hra_subset.owl $(COMPONENTSDIR)/mappings.owl $(COMPONENTSDIR)/blood_and_immune_upper_slim.owl $(COMPONENTSDIR)/eye_upper_slim.owl $(COMPONENTSDIR)/general_cell_types_upper_slim.owl $(COMPONENTSDIR)/kidney_upper_slim.owl $(COMPONENTSDIR)/cellxgene_subset.owl $(COMPONENTSDIR)/clm-cl.owl $(COMPONENTSDIR)/wmbo-cl-comp.owl $(COMPONENTSDIR)/2DFTU_HRA_illustrations.owl
```
gouttegd: For reference, the proper way to fix conflicts affecting the standard ODK Makefile is to

1. fix any conflict affecting the `cl-odk.yaml` file, and
2. run `sh run.sh update_repo` to get the ODK to re-generate the Makefile entirely.

No need to revert already committed changes, but please do run the `sh run.sh update_repo` command to ensure the Makefile is generated correctly, and commit the resulting changes.

hkir-dev (dismissed):

src/ontology/Makefile
aleixpuigb: Thank you Damien!

gouttegd (dismissed):

PR #2811, opened 2024-11-27
Add taxon subsets

gouttegd:
This PR adds taxon-specific subsets to CL.

A taxon-specific subset for a given taxon is automatically generated by using the taxon constraints declared in the ontology to exclude any class that, due to the constraints, is known not to exist in the taxon. This is the same approach used for similar subsets in Uberon (obophenotype/uberon#3363).

Two new release artefacts are added: `human-view.owl` (the human-specific subset) and `mouse-view.owl` (the mouse-specific subset). In addition, in the main release product (`cl.owl`), classes that were found to belong in the human (respectively mouse) subset are tagged with a `http://purl.obolibrary.org/obo/cl#human_subset` (respectively `http://purl.obolibrary.org/obo/cl#mouse_subset`) subset annotation (again, similar to what was done in Uberon).


src/ontology/cl.Makefile
```diff
@@ -34,6 +34,16 @@ cl-plus.owl: $(ONT)-full.owl
 		 annotate --ontology-iri $(ONTBASE)/$@ \
 		 $(ANNOTATE_ONTOLOGY_VERSION) --output $@
 
+# Main release artefact (cl.owl)
+# We override the standard ODK-generate rule so that we can merge in the
+# tags for the taxon subsets.
+POSTPROCESS_ADDITIONS = subsets/human-tags.ofn \
+                        subsets/mouse-tags.ofn
+$(ONT).owl: $(ONT)-full.owl $(POSTPROCESS_ADDITIONS)
```
matentzn, line 42: There is some mild itch in my nose to overwrite x.owl rather than x-full.owl, just because it means that the primary release does not correspond direct to any of the known ontology types. Minor itch though, if you don't share that sentiment I am fine with this.

matentzn: I guess this way you avoid circularity..

matentzn (approved):
Looks great. One minor nitpick but I will leave it to you to ignore or act on

src/ontology/cl.Makefile
gouttegd: > the primary release does not correspond direct to any of the known ontology types

Never thought this could be a concern. o_O `uberon.owl` also does correspond to any of the ODK-defined “ontology types“.

And in this case, the difference between `cl-full.owl` and `cl.owl` is quite minor, since it is only the additions of the `oboInOwl:inSubset` annotations, whereas in Uberon, `uberon.owl` uses a completely different pipeline than `oberon-full.owl`.

gouttegd: A longer-term alternative could be to have some kind of `POSTPROCESS` step in the ODK (similar to the `PREPROCESS` that is already in there), which would be a no-op by default but that could be overridden by projects if they need to perform some last-minute changes at the very end of the build process.

PR #2847, opened 2024-12-06
Addresses #2786 NTR CD57-positive enterocyte

### LINKED ISSUE #2786: [NTR] CD57-positive enterocyte described by Stellar pipeline snRNAseq + CODEX/Phenocycler John Hickey

**Preferred term label**
- [ ] **CD57-positive enterocyte**

CD57+ enterocyte

Description:
Definition: An enterocyte of the human duodenum (small intestine) expressing CD57 or HNK-1.
Localization: Primarily enriched in the duodenum in gland structures, with rare presence in other regions of the intestine.
Characteristics: These cells exhibit a distinct profile compared to other enterocytes. They are less abundant and often found in glandular regions of the duodenum.
Significance: While CD57 is traditionally a marker for NK cells and certain T cells, in this context, it identifies a unique subset of enterocytes that may have specialized functions in the small intestine (CD57 is present on many cell types but its combination with CD45 may most likely mark immune cells. This is similar to many molecules like CD56 happy to explain more).
**Definition for Cell Ontology:**
CD57-positive enterocyte: An enterocyte of the human duodenum expressing beta-1,3-glucuronyltransferase 1 (B3GAT1), also known as CD57 or HNK-1. These cells have a distinct expression profile compared to other enterocytes and are enriched within the duodenal epithelium.

Hickey, John W., Winston R. Becker, Stephanie A. Nevins, Aaron Horning, Almudena Espin Perez, Chenchen Zhu, Bokai Zhu, et al. 2023. “Organization of the Human Intestine at Single-Cell Resolution.” Nature 619 (7970): 572–84. https://doi.org/10.1038/s41586-023-05915-x.

enterocyte of epithelium proper of duodenum CL:1000340

human duodenum UBERON:0002114

**Your ORCID**
Ellen M. Quardokus 0000-0001-7655-4833
John W. Hickey 0000-0001-9961-7673

**Additional Clarifications**
**Regional Localization:**
CD57-positive enterocytes: Enriched in the duodenum (small intestine).
**Markers Associated with Immune Cells:**
While CD57 is traditionally associated with immune cells (e.g., NK cells), its expression on enterocytes marks unique epithelial subsets with specific functions or maturation states.
**Disease Association:**
Hickey et al 2023. Nature 619 (7970): 572–84. study focused on healthy intestinal tissues. The expression of this marker in enterocytes within healthy samples suggests that they are part of normal physiological processes rather than disease states.
**Need for Specific Designations:**
CD57-positive enterocytes: The expression of this marker highlights a subset with distinct maturation stages or functional roles, contributing to the diversity of the intestinal epithelium.
**Summary**
We identified this enterocyte type to capture the heterogeneity within the intestinal epithelium:
CD57-positive enterocytes represent a rare, specialized population in the duodenum.

aleixpuigb:
Fixes #2786 NTR CD57-positive enterocyte


aleixpuigb:
#gogoeditdiff

aleixpuigb:
CellMark comments have been imported when refreshing the imports, causing that some terms have two comments and failing the QC (I might be wrong on the assessment, please correct me if I am wrong). We might need a different AP to avoid having this problem.
CC @hkir-dev @gouttegd @Caroline-99

gouttegd:
> causing that some terms have two comments and failing the QC (I might be wrong on the assessment, please correct me if I am wrong)

Your assessment looks correct to me.

Looking at the CellMarks comments:

> * The marker set C1orf194, MS4A8 can identify the Human cell type ciliated columnar cell of tracheobronchial tree in the Lung with a confidence of 0.86 (NS-Forest FBeta value).
> * The marker set SST, PLCH1 can identify the Human cell type sst GABAergic cortical interneuron in the Brain with a confidence of 0.842367005 (NS-Forest FBeta value).
> * The marker set MSR1, FABP4 can identify the Human cell type alveolar macrophage in the Lung with a confidence of 0.80 (NS-Forest FBeta value).

It seems to me that such an information should rather belong to a dedicated annotation. Using `rdfs:comment` as a “joker” annotation to store any kind of structured data (such as a confidence score) is never a good idea.

If we can’t get the CellMarks folks to change that we might have to strip `rdfs:comment` annotations from the CLM component when we update it.

dosumis:
> Using rdfs:comment as a “joker” annotation to store any kind of structured data (such as a confidence score) is never a good idea.

The idea was to add markers and confidence scores in human readable form to go along with a formal specification  - including marker IRIs and confidence scores as floats.  The one comment restriction is an annoying OBO legacy.  My ideal solution would be to add SPARQL update to the pipeline to concatenate comments.  @gouttegd  - what's your opinion on this option? (I have a feeling you might object.)

gouttegd:
> what's your opinion on this option? (I have a feeling you might object.)

I am certainly not very fond of it. That being said, it’s not like I had anything much better to propose…

(A truly ideal solution for me would be, either: (1) use dedicated annotation properties for confidence scores and marker IRIs, and leave it to the ontology browser(s) to display those informations in a human-friendly form, or (2) get rid of the OBO format and its limitations entirely. Neither of which are really under our control.)

But adding a step in the pipeline to merge comments could be quite tricky as it involves comments spread over several files…

How about merging comments, but _only_ for OBO artefacts?

dosumis:
> How about merging comments, but only for OBO artefacts?

Works for me.

gouttegd:
> (Although breaking OBO<->OWL round-tripping makes me a bit queasy)

Pretty sure that horse has already left the barn… CL contains axioms that are not representable in OBO and that are rendered in the `owl-axioms` header tag, but we forcefully strip that tag when generating the OBO products, which therefore already do not contain the same contents as the products in other formats.

Caroline-99:
#gogoeditdiff

src/ontology/cl-edit.owl
```diff
@@ -32366,6 +32367,18 @@ AnnotationAssertion(Annotation(oboInOwl:hasDbXref "PMID:37124720") oboInOwl:hasE
...
 
+# Class: obo:CL_4033092 (CD57-positive enterocyte)
+
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref "PMID:37468586") obo:IAO_0000115 obo:CL_4033092 "An enterocyte that is part of a duodenal epithelium and expresses beta-1,3-glucuronyltransferase 1 (B3GAT1), also known as CD57 or HNK-1.")
+AnnotationAssertion(terms:contributor obo:CL_4033092 <https://orcid.org/0000-0001-6677-8489>)
+AnnotationAssertion(terms:contributor obo:CL_4033092 <https://orcid.org/0000-0001-7655-4833>)
+AnnotationAssertion(terms:contributor obo:CL_4033092 <https://orcid.org/0000-0001-9961-7673>)
+AnnotationAssertion(terms:date obo:CL_4033092 "2024-12-06T14:32:49Z"^^xsd:dateTime)
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref "PMID:37468586") oboInOwl:hasExactSynonym obo:CL_4033092 "CD57+ enterocyte")
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref "PMID:37468586") rdfs:comment obo:CL_4033092 "While CD57 is traditionally associated with immune cells (e.g., NK cells), its expression on enterocytes marks unique epithelial subsets with specific functions or maturation states. These cells are speciallyenriched in areas of the duodenum within submucosal glands.")
```
Caroline-99, line 32378: Everything looks good. There is only a typo in the comments section - should have a space here `speciallyenriched`

Caroline-99 (changes requested):

Caroline-99 (approved):

PR #2854, opened 2024-12-10
Fix some uncaught violations

gouttegd:
This PR fixes some violations of the rules regarding the use of annotation properties and their values:

* invalid use of `oboInOwl:created_by` instead of `dcterms:contributor`;
* invalid use of `oboInOwl:note` instead of `rdfs:comment`;
* invalid use of `dc:date` instead of `dcterms:date`;
* invalid typing of `dcterms:date` value as `xsd:string` instead of `xsd:dateTime`.

Some of those violations were directly in the `-edit` file and therefore definitely should have been caught automatically by the [illegal-annotation-property-violation](https://github.com/obophenotype/cell-ontology/blob/master/src/sparql/illegal-annotation-property-violation.sparql) and [illegal-date-violation](https://github.com/obophenotype/cell-ontology/blob/master/src/sparql/illegal-date-violation.sparql) SPARQL checks -- I have no idea why they were not.

Other violations came from the pattern-derived `definitions.owl` file, and were not caught because this file, as all other components, is _not_ covered by the SPARQL check -- I believe this is a bug of the ODK, I see no reason for not covering those files.


gouttegd:
@matentzn If you have any idea why the violations contained in the `-edit` file were not caught by the mentioned SPARQL checks -- I am at a loss on this one.

I kinda expect you to disagree about removing the `dcterms:date` annotation from the `cyclingCellState` pattern, on the ground that an incorrectly typed date annotation is better than no date annotation at all -- why not, but then I will insist that the `illegal-date-violation` check be removed: we cannot ask our editors to be careful about how they enter date annotations, flag their PRs as invalid when they mistakenly create a wrongly typed date, and then decide that it is actually perfectly OK to have a pattern that creates wrongly typed dates.

matentzn (approved):
Nice one.

> I kinda expect you to disagree about removing the dcterms:date annotation from the cyclingCellState pattern, on the ground that an incorrectly typed date annotation is better than no date annotation at all 

I am not so concerned about term level dates personally, I am totally fine with this!

PR #2861, opened 2024-12-13
Stomach smooth muscle inner oblique layer cell

JABelfiore:
defs, refs, uberon import - stomach smooth muscle inner oblique layer imported from uberon


Caroline-99 (changes requested):

Caroline-99:
#gogoeditdiff

Caroline-99 (changes requested):
Thank you for the changes. I am not sure how relevant this reference is **PMID:33630273** as I don't have access to the full article, and the abstract does not seem to be relevant to this cell type **stomach smooth muscle inner oblique layer cell** or the definition. I would suggest removing or replacing it with a more relevant reference.

Caroline-99 (approved):

PR #2930, opened 2025-02-04
Fixes #2923

### LINKED ISSUE #2923: Many cell types have a useless subclass to BFO:0000002 (continuant)
E.g leukocyte: http://purl.obolibrary.org/obo/CL_0000738

Image

in OAK:

Image

Protege:

Image

Explanation:

Image

it's a complicated one but at the end of the day there are domain constraints that reference an above the shoreline BFO class. The fix is to remove axioms that reference above the shoreline or dangling classes (either) in the release product. Not sure if there is a standard ODK setup for this.

(of course base files remove all these issues, but browsers cannot use base files yet)

Note: before someone says that COB should fix this, COB won't fix it until this is decided on: https://github.com/oborel/obo-relations/issues/780

dosumis:


dosumis:
I guess injecting a label is considered bad form, but have done it anyway.  Aim at least is a temp hack.

matentzn:
This is not how you do it with ODK; I don't understand what your proposed solution is here: simply adding BFO as an import and punt the "above the shoreline" discussion for later?

dosumis:
> This is not how you do it with ODK; I don't understand what your proposed solution is here: simply adding BFO as an import and punt the "above the shoreline" discussion for later?

Mapping to BFO has to live somewhere.  Adding direct BFO fixes the graph - one link to CL:Cell vs a mess of inferences on individual CL terms from RO relation domains. Adding  in a label is a nasty hack but means we don't get a bare root term in our release.  A more complete fix would be to add BFO to our imports.  This would fulfill the use case of having domain axioms on relations - as we can then use it to test for incoinstencies. 

This is one specific example of a much wider problem -  e.g. we have similar inferences on GO imports to CL

Note: the ODK base pipeline is basically broken for these mappings. Relevant mappings to BFO in Uberon don't make it through to its base.  Presumably fixed by adding BFO as import to Uberon.  But we would also need to be true for all other importing ontologies (which we don't control)  + there may be conflicts with plan to move to COB?

I would like a fix ASAP.

Alternative suggestion:

We develop a more generic BFO mapping file (with its own repo?) covering Uberon, GO, Pro etc.  This is available for import (as a component or regular import?) to any ontology that wants it.  This would ensure BFO mappings for all imports (Uberon, Pro, GO etc) are available for all cases.  Ontologies that use it would need to import BFO.  We might want to consider some minimal import.  I'll be happy to switch this to COB at a future date if that doesn't break anything.

Thoughts?

PR #2952, opened 2025-02-11
2887 type 1 submucosal plexus glia synonym 

Caroline-99:
Changed exact synonym to related synonym with abbreviation tag.

related #2946


Caroline-99:
#gogoeditdiff

JABelfiore (approved):

PR #2969, opened 2025-02-18
type 4 muscularis propria 

JABelfiore:
definition and references


dosumis (changes requested):
image
 Failing because used def instead of dbxref. (You can see the line failed on in report.  I just searched with the PMID  from that line to find this in Protege)

dosumis (approved):

PR #3008, opened 2025-03-18
small keratin cholangiocyte

JABelfiore:
def, refs, logical


dosumis:
#gogoeditdiff

dosumis:
> This cell is defined by its expression of keratin 7 (K7), a key marker of cholangiocyte identity

If this is a general cholangiocyte marker, why name a subtype for it?  How to small keratin cholangiocytes differ from other small chaolangiocytes?  Is there anything in this def that would not be true for the parent class?  (Worth looking at its extended description).

Also rather a lot of detail on pathology.

Caroline-99:
> > This cell is defined by its expression of keratin 7 (K7), a key marker of cholangiocyte identity
> 
> If this is a general cholangiocyte marker, why name a subtype for it? How to small keratin cholangiocytes differ from other small chaolangiocytes? Is there anything in this def that would not be true for the parent class? (Worth looking at its extended description).
> 
> Also rather a lot of detail on pathology.

@JABelfiore I agree with David, I am not sure how small keratin cholangiocytes differ from other small cholangiocytes. 

Also, in CL, we currently have 'intrahepatic cholangiocyte' with 'small bile duct cholangiocyte' as an exact synonym.` I think this needs to be revised as intrahepatic cholangiocytes refer to all cholangiocytes lining the intrahepatic bile ducts, which include both small and large cholangiocytes.

Caroline-99 (changes requested):

JABelfiore:
I have now spoken to Rachel about this issue. She informed me that for the small keratin cholangiocytes they included the term keratin because they saw this distinguished these small cholangiocytes from the small ApoLipo cholangiocytes. So these would be the more typically reported small cholangiocytes, but since they saw the previously unreported small ApoLipo cholangiocytes they wanted to be specific. We have agreed to use intrahepatic cholangiocyte at the CL term for these and the liver atlas team will just clarify in their manuscript how small ApoLipo differ from these.  

ACTION: Closing this issue.

PR #3040, opened 2025-04-04
Fixes #2808 gut endothelial cell text/logical def

### LINKED ISSUE #2808: gut endothelial cell has no subclasses and no textual definition
[gut endothelial cell](http://purl.obolibrary.org/obo/CL_0000131)  has no subclasses and no textual definition

Caroline-99:
Fixes #2808
- Added a textual def with references
- logical def - updated the logical defs


Caroline-99:
#gogoeditdiff

Issue #3015, opened 2025-03-19
[Obsolete] defensive cell
Presumably most of the immune system should be here if we were to keep it, rather than the current arbitrary handful.


aleixpuigb:

addiehl:
I approve of this obsoletion.

PR #3056, opened 2025-04-09
obsolete defensive cell 

JABelfiore:


Caroline-99:
The QC is failing because the subclasses of the obsoleted term haven’t been handled. You’ll need to check those subclasses and make sure they are reclassified/updated appropriately. (I also usually remove the obsoleted term from their subClassOf relationships -i.e., from the child terms - when obsoleting a term.)

image

src/ontology/cl-edit.owl
```diff
@@ -7655,11 +7655,12 @@ AnnotationAssertion(obo:IAO_0100001 obo:CL_0000472 obo:PO_0000076)
 AnnotationAssertion(rdfs:label obo:CL_0000472 "obsolete chlorenchyma cell")
 AnnotationAssertion(owl:deprecated obo:CL_0000472 "true"^^xsd:boolean)
 
-# Class: obo:CL_0000473 (defensive cell)
+# Class: obo:CL_0000473 (OBSOLETE defensive cell)
```
Caroline-99, line 7658: please add a term tracker item annotation with the link to the ticket as an iri (ie. https://github.com/obophenotype/cell-ontology/issues/3015)

Caroline-99 (changes requested):

Caroline-99:
#gogoeditdiff

Caroline-99:
#gogoeditdiff

PR #3068, opened 2025-04-17
Fixed #3062 Oligodendrocyte Precursor Cell

### LINKED ISSUE #3062: Review definition and classification OPC

RiveraAndrea83:
Changed Definition AND Included a synonim AND included references AND Improved Axiomatisation.

I would like to add capable of some expression to include the expression of key genes but I am not sure how.


dosumis (commented):
Is type II astrocyte differentiation just in vitro ?

dosumis (approved):
Looks good. One small change - can you add OPC on its own as a synonym?

RiveraAndrea83:
I am happy to close it as is, however looking at it again, I could remove the O2A mention and change it to NG2? Otherwise we can do that at a later date?

PR #3114, opened 2025-05-06
2922 class hierarchy terminally differentiated cells

JABelfiore:
obsolete non-terminally differentiated cell, all terms grouped under this now groped under progenitor cell


Caroline-99:
#gogoeditdiff

src/ontology/cl-edit.owl
```diff
@@ -4297,8 +4299,8 @@ AnnotationAssertion(oboInOwl:hasDbXref obo:CL_0000062 "CALOHA:TS-0720")
 AnnotationAssertion(oboInOwl:hasDbXref obo:CL_0000062 "FMA:66780")
 AnnotationAssertion(rdfs:comment obo:CL_0000062 "non-encoded relationships from VSAO - capable_of_producing VSAO:0000020")
 AnnotationAssertion(rdfs:label obo:CL_0000062 "osteoblast")
```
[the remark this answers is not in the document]
Caroline-99: This is beyond the scope of this PR, but shouldn't **preosteoblast** be classified under **non-terminally differentiated osteoblast**? And the same thing for **preodontoblast**? 

If accurate, it might be worth doing these edits in a different PR?

src/ontology/cl-edit.owl
```diff
@@ -3921,7 +3921,7 @@ SubClassOf(obo:CL_0000029 ObjectSomeValuesFrom(obo:RO_0002202 obo:CL_0002676))
```
Caroline-99, line 3921: Looking at the reasoned diff, a few cells have lost the 'eukaryotic cell' inference after replacing a non-terminally differentiated cell with a progenitor cell. It might be worth adding this classification manually.

Caroline-99 (changes requested):

src/ontology/cl-edit.owl
JABelfiore: agreed after discussion: eukaryotic cell will NOT be added

Caroline-99 (approved):

PR #3125, opened 2025-05-21
Fix bug in contact detail line

aleixpuigb:
The @ symbol was causing the apostrophes to keep moving and CI failing.


gouttegd (approved):

gouttegd:
Just for the record: the bug is not in the contact line, which is perfectly valid according to the OWL Functional Syntax format. The bug is _somewhere_ in our software stack, which at times is interpreting the `@` as the beginning of a language tag, which should not happen.

PR #3128, opened 2025-05-30
Update repository to ODK 1.6

gouttegd:
This PR updates the repository to the new ODK 1.6.

It uses newly available (or fixed) features from the ODK to

* automatically produce “clean” OBO artefacts;
* treat the `cl.sssom.tsv` mapping set as a release artefact;
* automatically download remote-sourced components.

It also fixes some duplicated IDs in the ID policy file.


matentzn (approved):
Great step for mankind

PR #3133, opened 2025-06-05
FIXED defition for Cajal-Retzius cell

RiveraAndrea83:
Fixed definition AND added references AND improved axiomatisation


dosumis:
#gogoeditdiff

dosumis (approved):
Looks great.  Would like a couple of changes:

1.  Would be good to add some text on where in the brain they survive in adults and the species specificity of this.
2. Axiom locating to HPF is too strong as clearly broader than that earlier in dev.  We can keep this in text only for now. In future we might consider how we could add it on the T-type defined with reference to WMBO (better mechanism needed).

dosumis (changes requested):
See previous comment

dosumis (approved):

PR #3141, opened 2025-06-17
Update cl-edit.owl

dosumis:
Changed to defined class.  Moved Lugaro cell to granular layer (molecular layer is definitely wrong - see text def)


dosumis:
> There are seven terms that are adding images, but only one show in the diff file and in protégé. Did you copy the code line to the others? I also did that in https://github.com/obophenotype/cell-ontology/pull/3144 and the annotation property was not shown in protégé when I hadn't added manually. I am just wondering if this will affect on displaying the image in OLS as well.

Oops - screwed up manual edits.  Fixing now.

dosumis:
#gogoeditdiff

PR #3144, opened 2025-06-17
Addresses #2920 NTR Schlemm's canal endothelial cell and subclasses

### LINKED ISSUE #2920: [NTR] Schlem's canal endothelial cell

**Preferred term label**
Schlem's canal endothelial cell

**Synonym(s)** (add reference(s), please)

A specialised endothelial cell of the canal of Schlem. This cell type share characteristics of both blood and lymphatic endothelial cell [2](https://en.wikipedia.org/wiki/Schlemm's_canal)[3](https://journals.plos.org/plosbiology/article?id=10.1371%2Fjournal.pbio.1001912)[4](https://elifesciences.org/reviewed-preprints/96459).  The inner wall endothelial cells of SC form "giant vacuoles" and transcellular pores that facilitate the passage of aqueous humor into collector channels which are critical for maintaining proper intraocular pressure [7](https://pmc.ncbi.nlm.nih.gov/articles/PMC4268318/).  They regulate of aqueous humor outflow resistance by modulating pore formation and responding to mechanical strain.

endothelial cell

http://purl.obolibrary.org/obo/UBERON_0004029 canal of Schlem

**Your ORCID**

**Additional notes or concerns** If there are markers that uniquely identify the cell type in a particular species, they can be added here.

Def adapted from https://www.perplexity.ai/search/are-the-endothelial-cells-of-s-niggOTOzSiCWHHDGST62IQ
Refs need checking!  
Also curious about whether we need separate subtype for inner wall cells as these feature giant vacuoles.  

Open access figures:  https://www.mdpi.com/2079-4983/6/3/963.

aleixpuigb:
Fixes #2920 NTR Schlemm's canal endothelial cell and subclasses Specific terms for the inner and outer wall have also been included


aleixpuigb:
#gogoeditdiff

aleixpuigb:
Thanks @dosumis. That was what I was doing, as I was not sure why not all the terms that have added images are showing in protege (e.g., granule cell, while Lugaro cell shows it). I will follow the pattern for now and ask you to review

src/ontology/cl-edit.owl
```diff
@@ -32655,6 +32662,38 @@ AnnotationAssertion(rdfs:label obo:CL_4033096 "diffuse bipolar 4b cell")
...
 
+# Class: obo:CL_4033097 (Schlemm's canal endothelial cell)
+
+AnnotationAssertion(obo:IAO_0000115 obo:CL_4033097 "A specialised endothelial cell that is part of the canal of Schlemm. This cell type shares characteristics and expresses markers of both blood and lymphatic endothelial cells (Kizhatil et al., 2014) and regulates aqueous humor outflow resistance by modulating pore formation and responding to mechanical strain (Stamer et al., 2016).")
+AnnotationAssertion(terms:contributor obo:CL_4033097 <https://orcid.org/0000-0001-6677-8489>)
+AnnotationAssertion(terms:date obo:CL_4033097 "2025-06-16T13:55:29Z"^^xsd:dateTime)
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref "PMID:25051267") Annotation(oboInOwl:hasSynonymType obo:OMO_0003000) oboInOwl:hasRelatedSynonym obo:CL_4033097 "SEC")
+AnnotationAssertion(rdfs:label obo:CL_4033097 "Schlemm's canal endothelial cell")
+AnnotationAssertion(Annotation(dc:license <http://creativecommons.org/licenses/by/4.0/>) Annotation(oboInOwl:hasDbXref <PMID:26402712>) foaf:depiction obo:CL_4033097 <http://purl.obolibrary.org/cl/images/Schlemms_canal_endo_cell_PMID26402712.png>)
```
aleixpuigb, line 32672: @gouttegd is this the correct format? I normally add dbxrefs as strings, but other images have them as IRI.
 Thank you!
 Edited for the second part, since I found out the problem.

aleixpuigb:
#gogoeditdiff

src/ontology/cl-edit.owl
dosumis: Should be strings. 
Also think this image would benefit from an attached comment with details of what's depicted - unpacking acronyms and mentioning cell types.

dosumis: Have merged CB neurons, but will make a quick fix to DBxrefs there in a separate PR.

dosumis (changes requested):
Annotate image with an rdfs:comment  -  details of what's depicted - unpacking acronyms and mentioning cell types.

src/ontology/cl-edit.owl
```diff
@@ -32663,6 +32670,38 @@ AnnotationAssertion(rdfs:label obo:CL_4033096 "diffuse bipolar 4b cell")
...
```
gouttegd, line 32680: As noted by @dosumis, the cross-reference here should be a string, not a IRI – especially since that IRI, `<PMID:26402712>`, is in fact bogus. When using the IRI syntax, the _full_ IRI must be used, as in `<https://pubmed.ncbi.nlm.nih.gov/26402712>`.

gouttegd, line 32680: Regarding the link to the image:

(1) CL is currently inconsistent in how links to images are represented, with three different forms being used:

* IRI (as here): `<http://purl.obolibrary.org/cl/images/Schlemms_canal_endo_cell_PMID26402712.png>`
* String literal: `"http://purl.obolibrary.org/cl/images/Schlemms_canal_endo_cell_PMID26402712.png"`
* URI literal: `"http://purl.obolibrary.org/cl/images/Schlemms_canal_endo_cell_PMID26402712.png"^^xsd:anyURI`

I don’t think I am aware of a decision having been made somewhere as to which form should be used. For what it’s worth: (i) I personally think the most “correct” form is the last one (URI literal); (ii) Protégé doesn’t care about which form is used, and can show the image thumbnail in all 3 cases; (iii) I don’t know if maybe OLS has a preference for one form over another.

(2) But regardless of which form is used, what matters most is that the link should be correct (i.e. it should resolve to the image, taking into account the PURL redirection). Here, `http://purl.obolibrary.org/cl/images/Schlemms_canal_endo_cell_PMID26402712.png` does _not_ resolve to the image, but merely to the CL repository. The correct link would be `http://purl.obolibrary.org/obo/cl/images/Schlemms_canal_endo_cell_PMID26402712.png` (note the additional `obo/` component at the beginning of the path).

aleixpuigb: Thank you so much for all the information! I will do all the changes.

dosumis: I'm pretty sure OLS needs an IRI.  I don't know whether it matter whether it is URI literal or IRI.  Could check against OLS if we have both.

@gouttegd  - Thanks for spotting the path errors.

aleixpuigb: I have checked with OLS team and OLS accepts IRI and URI. In CL we have both, so it would be good to make a decision on one to have consistency. @dosumis, are you ok adopting URI as standard from now on? I can go through all images and edit them.

dosumis: > @dosumis, are you ok adopting URI as standard from now on? I can go through all images and edit them.

Sure

dosumis (approved):

PR #3154, opened 2025-06-20
CL release v2025-06-20

aleixpuigb:


subsets/eye_upper_slim.tsv
```diff
@@ -1,1244 +1,979 @@
-?x	?label
-_:B8a35cd25X2De76eX2D417dX2DaacaX2D87e295d99a71genid1008
```
gouttegd, line 2: Just for information: this should be the last release where the `_slim.tsv` files contain so many random changes. With the ODK 1.6, the files will no longer contain blank nodes ID that change upon every release.

gouttegd (approved):

PR #3205, opened 2025-07-23
Addresses #3162 Create 120 terms for PNS neurons

aleixpuigb:
Addresses #3162
120 terms were created using a ROBOT template with information from the PNS ASCT+B table.
Waiting for https://github.com/obophenotype/pro_obo_slim/pull/66 to be merged to refresh the imports.

PR #3215, opened 2025-07-29
CL release 29-07-2025

Caroline-99:


gouttegd (changes requested):
The `2D_FTU_HRA_illustrations` and `PNS_neurons` subsets should not exist.

In the case of the `2D_FTU_HRA_illustrations` subset, it was mistakenly created in #2805 – unfortunately, when reviewing I missed the fact that `2D_FTU_HRA_illustrations` was added to the `cl-odk.yaml` file both as a component (as it should have been) _and_ as a subset (which it should _not_ have been).

In the case of the `PNS_neurons`, it was mistakenly created in #3205 (same reason: it was added both as a component and as a subset). The mistake was caught, but the Makefile had _not_ been re-regenerated after the subset had been removed from `cl-odk.yaml`, so it still contains the code to create that (empty) subset.

So please:

* Remove all `src/ontology/subsets/2DFTU_HRA_illustrations.*` and `src/ontology/subsets/PNS_neurons.*` files.
* Remove [line 48](https://github.com/obophenotype/cell-ontology/blob/8a812043da616c101b0b64f2f0b9d156a60ceb69/src/ontology/cl-odk.yaml#L48) in the `src/ontology/cl-odk.yaml` file (the line declaring `2D_FTU_HRA_illustrations` as a subset.
* Re-generate the Makefile by running `sh run.sh update_repo` and commit the re-generated Makefile.
* Run the release process again.

(Your call whether you want to do that in this PR, or discard this PR and start anew.)

Issue #3182, opened 2025-07-09
Move LAMP5 Lhx6 to under MGE & update description + add Mouse and Human subclasses


dosumis:
WMB taxonomy has Lamp5 Lhx6 under MGE, but Jorstad and earlier papers have it under CGE.  Recent studies of lineage strongly support MGE origin: https://platform.futurehouse.org/trajectories/b5b8c08e-8a39-4785-b2a1-638e789557a3  & https://www.biorxiv.org/content/10.1101/2024.06.18.599583v1 

### Edits to current term def (species neutral term)

> A transcriptomically distinct lamp5 GABAergic cortical interneuron located in the cerebral cortex that expresses Lamp5 and Lhx6. 

A transcriptomically distinct GABAergic interneuron marked by Lamp5 and Lhx6 expression, derived from the Medial Ganglionic Eminence (MGE), located in { add location summary }.

comment: Add details supporting MGE origin.  (Note - you can remove the NS-Forest marker summary, it will be automatically added back if we keep the mapping in CellMark - but see below).

Species specific subtypes:

### Manually add human subtype adding:

"A Lamp5 Lhx6 GABAergic interneuron of Homo sapiens. The standard transcriptomic reference data for this cell type can be found on the CellxGene census under the collection: 'Transcriptomic cytoarchitecture reveals principles of human neocortex organization', dataset: 'Supercluster: CGE-derived interneurons', Author Categories: 'CrossArea_subclass', Lamp5 Lhx6"

Make sure this also has link to CxG.  This should use the same pattern as WMBO: 
seeAlso: <URL link> @rdfs_label: "Reference data on CELLxGENE"

Update CellMark to move NS-Forest markers to human specific term:
 - https://github.com/Cellular-Semantics/CellMark/blob/main/src/markers/nsforest_human_neocortex_global_subclass_resultsSource.tsv

### Semi-automated addition of Mouse type: 

 - Edit WMBO to set add_to_CL to True here: https://github.com/Cellular-Semantics/whole_mouse_brain_ontology/blob/cl_product/src/dendrograms/supplementary/version2/CL_ontology_subset.tsv

Everything else should be automatic, although you can choose to update name and text here: https://github.com/Cellular-Semantics/whole_mouse_brain_ontology/blob/cl_product/src/patterns/data/default/CCN20230722_class_curation.tsv (defined_class_name & defined_class_definition columns respectively)

### Notes:
[...]

PR #3220, opened 2025-08-01
#3182 Lamp5 Lhx6 interneuron for species (general term, human, mouse)

RiveraAndrea83:
Please refer to issue ticket!

Edited the original term to general

Created new term species specific for Human AND added CxG reference (need to edit).

Create term for same cell specific for mouse which should be autocompleted after editing.


dosumis (changes requested):
Human - add comment on MGE classification despite Jorstad Supercluster.

src/ontology/cl-edit.owl
```diff
@@ -34664,6 +34665,23 @@ AnnotationAssertion(rdfs:label obo:CL_4072021 "corticotropin-releasing neuron")
...
+AnnotationAssertion(obo:RO_0002175 obo:CL_4072022 obo:NCBITaxon_9606)
+AnnotationAssertion(terms:date obo:CL_4072022 "2025-07-31T14:01:52Z"^^xsd:dateTime)
+AnnotationAssertion(oboInOwl:hasDbXref obo:CL_4072022 "https://cellxgene.cziscience.com/e/e4ddac12-f48f-4455-8e8d-c2a48a683437.cxg/")
+AnnotationAssertion(Annotation(oboInOwl:hasDbXref "PMID:32999462") rdfs:comment obo:CL_4072022 "Despite being clustered as CGE in this dataset, other studies suggest that the cell is of MGE origin because it expresses transcription factors such as LHX6 and NKX2-1, which are hallmark indicators of MGE-derived interneurons. Fate mapping experiments in mice have demonstrated that Lamp5+/Lhx6+ interneurons in both the hippocampus and neocortex arise from the MGE, and single-cell transcriptomic comparisons further show that primate neocortical Lamp5+/Lhx6+ cells closely resemble mouse hippocampal MGE-derived interneurons rather than classic CGE-derived types.")
```
Caroline-99, line 34736: Everythig looks good. I think this comment should be moved to the more general term as it is not specific to humans.

Caroline-99 (changes requested):

src/ontology/cl-edit.owl
RiveraAndrea83: the reason this comment is in the human specific term is because in mouse and other studies this cluster is mapped to MGE derived. it is only in this human study that the Lamp5 lhx6 is under CGE. Thus, the comment is specific to the human dataset/mapping.

Issue #3323, opened 2025-09-15
[Text def] Create human term for L5/6 NP neuron


RiveraAndrea83:
**CL term**
generalise original term and create human specific term for L5/6 NP

**Suggested revision of textual definition**

Links #3172

PR #3328, opened 2025-09-16
#3323 Created human term for L5/6 near projecting neuron

RiveraAndrea83:
Created human term, added human tags and generalised term


Caroline-99 (approved):
I have approved it but this is more of a note - the homo sapiens specific term is missing a CxG link. I think it woould be good to add it as aa reference to the definition if available to remain consistent with the other terms you have added.

RiveraAndrea83:
> @RiveraAndrea83 I have approved it but this is more of a note - the homo sapiens specific term is missing a CxG link. I think it woould be good to add it as aa reference to the definition if available to remain consistent with the other terms you have added

Added the reference

PR #3333, opened 2025-09-16
Dont relabel imported annotation properties

### LINKED ISSUE #3332: Re-labelling of imported annotation properties in the -edit file
Currently, CL’s `-edit` file contains some `rdfs:label` annotations for several imported annotation properties (APs) that _already_ have a label annotation (or sometimes even _several_ label annotations) from the import modules.

For example:

```
# Annotation Property: obo:IAO_0000115 (definition)

AnnotationAssertion(rdfs:label obo:IAO_0000115 "definition")

# Annotation Property: obo:IAO_0000424 (expand expression to)

AnnotationAssertion(rdfs:label obo:IAO_0000424 "expand expression to")

# Annotation Property: obo:IAO_0000700 (has ontology root term)

AnnotationAssertion(rdfs:label obo:IAO_0000700 "preferred_root")

# Annotation Property: obo:IAO_0100001 (term replaced by)

AnnotationAssertion(rdfs:label obo:IAO_0100001 "term replaced by")

# Annotation Property: oboInOwl:hasBroadSynonym (has_broad_synonym)

AnnotationAssertion(rdfs:label oboInOwl:hasBroadSynonym "has_broad_synonym")

# Annotation Property: oboInOwl:hasDbXref (has cross-reference)

AnnotationAssertion(rdfs:label oboInOwl:hasDbXref "database_cross_reference")

# Annotation Property: oboInOwl:hasExactSynonym (has exact synonym)

AnnotationAssertion(rdfs:label oboInOwl:hasExactSynonym "has_exact_synonym")
```

Not only are those annotations useless (since they are redundant with the _imported_ annotations), they are a frequent cause of spurious diffs, because the name of the annotation (as it appears in the comment immediately before the annotation assertion axiom) will frequently change (for not completely clear reasons, at least to me).

Is there are profound reason why those redundant label annotations still exist in the -edit file? In my opinion, they have nothing to do here (CL has no business stating what the label of, say, `IAO:0000115` should be), and I’d like to propose to remove them all.

gouttegd:
The -edit file contains `rdfs:label` annotation for several annotation properties (such as `oboInOwl:hasDbXref`, `IAO:0000115`, `oboInOwl:hasExactSynonym`, etc.) that already have a label defined in the `merged_import` module.
    
Not only are those labels useless (since redundant), but they are a frequent cause of spurious diffs, so they are better removed, which we do in this PR.

(As a side-effect, this PR also re-serializes the edit file, which is needed because a previous AI-generated change inserted a class at the wrong place, because of course LLMs don’t know that the axioms in a OFN file are conventionally _sorted_ – e.g., the axioms about a CL:7770002 class should appear _before_ the axioms about a CL:7770003 class.)

closes #3332


matentzn (approved):
Big time YES (spurious diff a bit weird but all ok)

PR #3334, opened 2025-09-16
Support temporary IDs.

gouttegd:
This PR ports to CL the temporary ID management system already in use in Uberon, for allowing AI agents (or in fact any editor, AI or not) to use “temporary” IDs that are to be later replaced by definitive IDs.

It allocates two new ID ranges:

* the `Temporary IDs` range (CL:99NNNNN): any ID picked within that range is treated as a temporary ID, slated for later replacement by a definitive ID;
* the `Automation` range (CL:0020000–CL:0120000): this is a “normal” ID range (IDs picked within that range are not treated differently from any other ID), but is reserved for automated processes – notably, for the process that replaces temporary IDs by definitive ones.

It also adds a new GitHub Action workflow, `allocate-definitive-ids`, which when triggered finds all temporary IDs (all IDs in the CL:99NNNNN range) in the cl-edit.owl file and replaces them by newly allocated IDs from the `Automation` range. That workflow can be used in two different ways:

(A) Manual trigger: When a PR contains temporary IDs, a CL maintainer can manually trigger the `allocate-definitive-ids` workflow on the PR’s underlying branch, _just before_ merging the PR into the master branch. This will automatically add a commit to the PR, in which temporary IDs are replaced by definitive ones. The PR should then be merged _as soon as possible_ once that commit has been added.

(B) Automatic trigger: If a PR containing temporary IDs is merged into the master branch (with the temporary IDs still present), the `allocate-definitive-ids` workflow will automatically be triggered and push a new commmit to replace the temporary IDs by definitive ones, directly on the master branch.

### DIFF
--- a/.github/workflows/allocate-definitive-ids.yml
+++ b/.github/workflows/allocate-definitive-ids.yml
@@ -1,0 +1,38 @@
+name: Allocate definitive IDs
+
+on:
+  push:
+    branches: [ master ]
+    paths:
+      - 'src/ontology/cl-edit.owl'
+  workflow_dispatch:
+
+jobs:
+  assign_ids:
+    runs-on: ubuntu-latest
+    container: obolibrary/odkfull:v1.6
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+          fetch-depth: 0
+
+      # Reallocate all temporary IDs as definitive IDs and rewrite
+      # axioms accordingly; this will have the side-effect of forcing a
+      # reserialisation of the -edit file if needed.
+      - name: Allocate definitive IDs
+        run: |
+          cd src/ontology
+          make allocate-definitive-ids
+
+      # Commit the changes; the 'push' action takes care of checking
+      # whether there are changes to commit at all, so if no IDs were
+      # reallocated (and no reserialisation was needed), nothing will
+      # happen here.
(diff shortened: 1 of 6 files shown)


matentzn (commented):
This is a very smart and well designed system for temporary ids. Despite the seemingly bullet proof GHA, I would propose a second line of protection which is a QC check that, at release times, ensures no temporary ids are released in the wild. 

How could that happen? -- Someone creates a component "my awesome celltypes" which is driven directly from a robot or dosdp template

I know they would be in violation of the id range specification, but given that this system here will likely be rolled out to all ontologies in OBO over time, I believe its prudent to at least protect against this case.

I did not approve yet as I want two more review on it which I will chase now: @balhoff and @cmungall.

gouttegd:
>  I would propose a second line of protection which is a QC check that, at release times

Which raises the question of how to run a test specifically _at release time_. We do not want such a test to run as part of the “normal” test suite, because it would then systematically flag any PRs that contain temporary IDs (even though it would be perfectly normal and expected for PRs to contain temporary IDs).

> I did not approve yet as I want two more review on it which I will chase now

FYI this PR implements _exactly_ the same system as the one in place in Uberon, which was discussed [here](https://github.com/obophenotype/uberon/issues/3564) and explicitly approved by @cmungall [here](https://github.com/obophenotype/uberon/pull/3567).

PR #3354, opened 2025-09-29
Fixes #3353 Created specific human term for chandelier pvalb gabaergic interneuron

### LINKED ISSUE #3353: [Text def] Create human specific term for chandelierl Pvalb GABAergic neuron
**CL term**

**Suggested revision of textual definition**

RiveraAndrea83:
Generalised definition and removed cortical term from label and definition. Created human term and added in taxon human.


Caroline-99:
#gogoeditdiff

PR #3386, opened 2025-10-15
Standardised names and in_taxon axioms. Minor update to doc.

dosumis:
Edited using Claude. 

Aim is to standardise all mention of species in names + ensure complete in_taxon relationships

Reports generated by Claude:
[cl_intaxon_analysis_report.txt](https://github.com/user-attachments/files/22934387/cl_intaxon_analysis_report.txt)
[cl_intaxon_final_report.txt](https://github.com/user-attachments/files/22934388/cl_intaxon_final_report.txt)


dosumis:
#gogoeditdiff

src/ontology/cl-edit.owl
```diff
@@ -12345,7 +12345,8 @@ AnnotationAssertion(oboInOwl:hasExactSynonym obo:CL_0000938 "CD16-negative, CD56
 AnnotationAssertion(oboInOwl:hasExactSynonym obo:CL_0000938 "CD56-bright cytokine secreting NK cell")
 AnnotationAssertion(oboInOwl:hasExactSynonym obo:CL_0000938 "CD56-bright cytokine secreting natural killer cell")
 AnnotationAssertion(rdfs:comment obo:CL_0000938 "This cell type is compatible with the HIPC Lyoplate markers for 'CD16-CD56bright NK cell'. Markers are associated with human cell types.")
-AnnotationAssertion(rdfs:label obo:CL_0000938 "CD16-negative, CD56-bright natural killer cell, human")
+AnnotationAssertion(rdfs:label obo:CL_0000938 "CD16-negative, CD56-bright natural killer cell (Hsap)")
+AnnotationAssertion(oboInOwl:hasExactSynonym obo:CL_0000938 "CD16-negative, CD56-bright natural killer cell, human")
```
gouttegd, line 12349: Another illustration of the fact that AI agents don’t know how to edit a OFN file correctly, and cause spurious diffs. Those axioms have been inserted in an out-of-order fashion.

I am going to [once again](https://github.com/obophenotype/uberon/pull/3511#issuecomment-2826999537) strongly suggest that any AI-based workflow should _systematically_ include a ROBOT-based re-serialisation step so as to avoid polluting the history with countless of spurious diffs.

src/ontology/cl-edit.owl
```diff
@@ -13501,6 +13512,7 @@ SubClassOf(obo:CL_0001054 obo:CL_0000576)
 # Class: obo:CL_0001055 (CD14-positive, CD16-low monocyte)
 
 AnnotationAssertion(Annotation(oboInOwl:hasDbXref "GOC:add") Annotation(oboInOwl:hasDbXref "PMID:20628149") obo:IAO_0000115 obo:CL_0001055 "An intermediate monocyte that is CD14-positive and with low amounts of CD16.")
+AnnotationAssertion(oboInOwl:hasExactSynonym obo:CL_0001055 "intermediate monocyte (Hsap)")
```
gouttegd, line 13515: It’s unclear whether that term is human-specific.

It does _not_ have a `in_taxon some 'Homo sapiens'` relationship (and, unless I am mistaken, does not inherit one). It has a `present_in_taxon some 'Homo sapiens'` annotation, which merely indicates that the cell type is known to exist in humans but does not imply it _only_ exists in humans.

There’s a comment saying that “the definition is valid for human intermediate monocytes”, but again, that doesn’t necessarily imply that the definition is not valid for other taxa.

Either the cell type is actually human-specific, in which case it should have the proper `in_taxon` restriction and “(Hsap)” should be added to the label (not just on a synonym); or we don’t know if it is human-specific, in which case _exact_ synonyms should not include a species qualifier (regardless of whether it is “human” or “Hsap”).

src/ontology/cl-edit.owl
```diff
@@ -18187,6 +18228,7 @@ SubClassOf(Annotation(oboInOwl:is_inferred "true") obo:CL_0002395 obo:CL_0000860
 AnnotationAssertion(Annotation(oboInOwl:hasDbXref "GOC:tfm") Annotation(oboInOwl:hasDbXref "PMID:20870168") obo:IAO_0000115 obo:CL_0002396 "A patrolling monocyte that is CD14-low and CD16-positive.")
 AnnotationAssertion(terms:contributor obo:CL_0002396 <https://orcid.org/0000-0003-1980-3228>)
 AnnotationAssertion(oboInOwl:creation_date obo:CL_0002396 "2010-10-04T12:25:42Z"^^xsd:dateTime)
+AnnotationAssertion(oboInOwl:hasExactSynonym obo:CL_0002396 "non-classical monocyte (Hsap)")
```
gouttegd, line 18231: Similar case to “CD14-positive, CD16-low monocyte” (CL:0001055) above. It is unclear whether that term is really human-specific.

src/ontology/cl-edit.owl
```diff
@@ -24008,33 +24063,36 @@ TA: Transit Amplifying Cells") foaf:depiction obo:CL_0009080 <http://purl.obolib
 EquivalentClasses(obo:CL_0009080 ObjectIntersectionOf(obo:CL_0019032 ObjectSomeValuesFrom(obo:BFO_0000050 obo:UBERON_0002108)))
 SubClassOf(obo:CL_0009080 ObjectSomeValuesFrom(obo:BFO_0000050 obo:UBERON_0001902))
 
-# Class: obo:CL_0009081 (specified double negative thymocyte (Homo sapiens))
+# Class: obo:CL_0009081 (specified double negative thymocyte (Hsap))
```
gouttegd, line 24066: Sigh… You gotta love the consistency of AI tools. For all the terms above whose label has been changed, the agent didn’t bother to update the comment to reflect the new label, and here all of a sudden it decides that it won’t do, and that the comment must be updated.

I guess deterministic tools are just boring.

src/ontology/cl-edit.owl
```diff
@@ -29928,6 +29992,7 @@ AnnotationAssertion(oboInOwl:hasExactSynonym obo:CL_4023019 "L5/6 CCK Vip (Mus m
 AnnotationAssertion(oboInOwl:inSubset obo:CL_4023019 cl:BDS_subset)
 AnnotationAssertion(rdfs:label obo:CL_4023019 "L5/6 cck, VIP GABAergic interneuron (Mmus)")
 EquivalentClasses(obo:CL_4023019 ObjectIntersectionOf(obo:CL_4023071 ObjectSomeValuesFrom(obo:RO_0002292 obo:PR_P32648)))
+SubClassOf(obo:CL_4023019 ObjectSomeValuesFrom(obo:RO_0002162 obo:NCBITaxon_10090))
```
gouttegd, line 29995: This is not wrong per se, but not necessary as CL:4023019 “inherits” the taxon restriction from CL:4023071.

src/ontology/cl-edit.owl
dosumis: Agree we should do this.

src/ontology/cl-edit.owl
dosumis: Human editors can be pretty inconsistent too...

dosumis:
> I shall also note that if we go ahead with the policy outlined here, we must remember to also apply it to outsourced components that are imported “as is” into CL.

> In particular, the wmbo-cl-component, imported from [here](https://github.com/Cellular-Semantics/whole_mouse_brain_ontology), contains hundreds of mouse-specific terms that all have the correct “(Mmus)” label suffix, but several of them are lacking an explicit taxon restriction (direct or inherited).

Agreed.  As external component is built with ODK, we could potentially enforce with common QC checks.

src/ontology/cl-edit.owl
dosumis: This session used only extremely basal tools.  Aim in future will be to have direct OWL query tools (and editing tools?) and to work on refining doc for agents (and humans).  With a decent DL query tool it would be very straightforward to detect redundancies like this.

src/ontology/cl-edit.owl
addiehl: Based on [PMID:19965649](https://pmc.ncbi.nlm.nih.gov/articles/PMC2810986/#:~:text=1–4,shared%20in%20humans%20and%20mice.), I am convinced that this is a human specific term.

src/ontology/cl-edit.owl
addiehl: Based on [PMID:20870168](https://pubmed.ncbi.nlm.nih.gov/20870168/) (the original reference) and [PMID:19965649](https://pubmed.ncbi.nlm.nih.gov/19965649/), I am convinced these are human specific cells.

PR #3389, opened 2025-10-16
Include full release report in CL

Caroline-99:
In the release notes (ie., https://github.com/obophenotype/cell-ontology/releases/tag/v2025-07-30), we have limited space to display the full report. The full release report (src/ontology/reports/summary_release.md) is now committed to the CL repository.

With the addition of CL links to other ontologies, the list of updates has gotten a lot longer, and it does not fit into the summary release notes.


.gitignore
```diff
@@ -32,6 +32,7 @@ src/ontology/mirror
 src/ontology/mirror/*
 src/ontology/reports/*
 !src/ontology/reports/release-diff.md
+!src/ontology/reports/summary_release.md
```
gouttegd, line 35: Just FYI: Adding that exception to the `.gitignore` file was not, in fact, necessary.

You could have added the file with `git add -f src/ontology/reports/summary_release.md` to force Git to accept it despite the `src/ontology/reports/*` exclusion rule), after which Git would have tracked the file regardless of the contents of `.gitignore`.

gouttegd (approved):

PR #3397, opened 2025-10-22
Produce and publish release artefacts as SemSQL DB files

### LINKED ISSUE #3396: Include sqlite with release process, to allow OAK to easily pick up the most recent version
Carrying on conversation from slack.

Damien Goutte-Gattat
  [44 minutes ago](https://obo-communitygroup.slack.com/archives/C03D93DEALA/p1761074799331719?thread_ts=1761067697.004829&cid=C03D93DEALA)
All that is needed to enable that is to add the value db to the export_format option in the ODK configuration.

I am volunteering CL to be the first group to try this

gouttegd:
This PR enables the production of SemSQL versions of the release artefacts, suitable for use by the Ontology Access Kit’s [SQL Database adapter](https://incatools.github.io/ontology-access-kit/packages/implementations/sqldb.html#sql-implementation).

Few things to note:

* This required changing the format of the various `-view.owl` subsets from OFN to RDF/XML, because [RDFTab.rs](https://github.com/ontodev/rdftab.rs) (a critical tool in the pipeline that produces the SemSQL files) only supports the RDF/XML format. This should not be a big deal, but it will cause massive diffs upon the next release (when the newly generated subsets will be committed), because all the `-view.owl` subsets will be completely rewritten from OFN to RDF/XML. (This could easily be avoided if we just stopped committing the subsets to the repository – they are _release artefacts_, not files that need to be tracked.)

* This adds ~2.6 GB worth of release artefacts to be uploaded upon every release. Depending on the network connection available, this could make publishing a release quite slow.

* The largest release artefact (`cl-plus`) is about ~1.1 GB in SemSQL format. The [limit for a release artefact](https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases#storage-and-bandwidth-quotas) on GitHub is 2 GB – so we have some margin, but maybe not that much.

* The PR also cleans up the `.gitignore` file, since it needed to be amended to be sure none of the SemSQL files will ever be inadvertently committed.

closes #3396

### DIFF
--- a/src/ontology/cl.Makefile
+++ b/src/ontology/cl.Makefile
@@ -221,7 +221,7 @@ subsets/%-view.owl subsets/%-tags.ofn: $(ONT)-full.owl | all_robot_plugins
 		 annotate --ontology-iri $(ONTBASE)/subsets/$*-view.owl \
 		          --version-iri $(ONTBASE)/releases/$(VERSION)/subsets/$*-view.owl \
 		          --annotation owl:versionInfo $(VERSION) \
-		          --output subsets/$*-view.owl
+		 convert --format ofn --output subsets/$*-view.owl
 
 
 # ----------------------------------------
@@ -307,6 +307,25 @@ endif
 
 update-HRA-illustrations:
 	python3 ./$(SCRIPTSDIR)/2D_FTU_images.py
+
+
+# ----------------------------------------
+# SemSQL release artefacts
+# ----------------------------------------
+# We do _not_ use the ODK builtin feature for now because it only
+# creates _uncompressed_ DB files, which are way too large to be
+# practical.
+
+DB_FILES = cl.db.gz $(foreach f, $(RELEASE_ARTEFACTS), $f.db.gz)
+RELEASE_ASSETS += $(DB_FILES)
+all_assets: $(DB_FILES)
+
+%.db.gz: %.owl
+	@rm -f $*.db $*.db.gz $*-relation-graph.tsv.gz .template.db .template.db.tmp
+	semsql make $*.db
+	@test -f $*.db || (echo "SQLite/SemSQL generation failed" && exit 1)
+	gzip $*.db
+	@rm -f $*-relation-graph.tsv.gz .template.db .template.db.tmp
 
 # ----------------------------------------
 # RELEASE DEPLOYMENT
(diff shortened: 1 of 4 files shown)


gouttegd:
Well, I do have concerns. The ODK is creating uncompressed DB files, and I don’t want people to start building pipelines that depend on such files.

I am amending the PR right now to make it publish compressed DB files instead (which unfortunately means that we must bypass the built-in ODK feature). Sorry about that.

src/ontology/cl.Makefile
```diff
@@ -308,6 +308,25 @@ endif
...
+DB_FILES = cl.db.gz $(foreach f, $(RELEASE_ARTEFACTS), $f.db.gz)
+RELEASE_ASSETS += $(DB_FILES)
+all_assets: $(DB_FILES)
+
+%.db.gz: %.owl
+	@rm -f $*.db $*.db.gz $*-relation-graph.tsv.gz .template.db .template.db.tmp
+	semsql make $*.db
+	@test -f $*.db || (echo "SQLite/SemSQL generation failed" && exit 1)
+	gzip $*.db
+	@rm -f $*-relation-graph.tsv.gz .template.db .template.db.tmp
```
matentzn, line 328: nitpick:

1. If you need db files locally, you need to ungzip them (you cant depend on say cl.db)
2. The code to generate semsql artefacts is duplicated compared to the odk default (which admittedly is not there anymore)
3. But since the use case is _specifically only_ to support tools like OAK etc remotely, this design choice is fine.

gouttegd: > If you need db files locally, you need to ungzip them (you cant depend on say cl.db)

When you say “depend”, you mean in a Makefile rule?

I don’t foresee that any CL workflow will ever depend on a DB file – there should never be any reason for that.

But even if do later need a DB file in a custom CL workflow, then we can revise that rule at that point.

As for someone needing the DB file outside of any workflow (say if you want to build the DB file so that you can use it locally with OAK), well, `gzip -d` is fast enough that it should not be a concern.

> The code to generate semsql artefacts is duplicated compared to the odk default

Yes. That’s because we _cannot_ use the ODK-generated rule. The way the SemSQL generation feature is implemented in the ODK right now, if you enable SemSQL production (so that you get the `%.db` production rule in the standard Makefile), you are automatically _also_ asking for the DB files to be published. But here we do _not_ want to publish the `.db` files, we want to publish only the `.db.gz` files – you can’t have that with the ODK for now.

PR #3402, opened 2025-10-27
Updared logical def CD25-positive, CD27-positive  immature gamma-delta T cell

Caroline-99:
Added CD27 positive annotation to CD25-positive, CD27-positive immature gamma-delta T cell


RiveraAndrea83:
#gogoeditdiff

RiveraAndrea83 (approved):

PR #3463, opened 2025-11-21
20251121-CL-release

Caroline-99:


cl-base.owl
```diff
@@ -210,6 +210,12 @@
     
 
 
+    <!-- http://purl.org/dc/elements/1.1/creator -->
+
+    <owl:AnnotationProperty rdf:about="http://purl.org/dc/elements/1.1/creator"/>
```
gouttegd, line 215: Why `dc/elements`? We use `dc/terms/contributor`, why not using `dc/terms/creator`?

cl.obo
```diff
@@ -48523,9 +48624,9 @@ property_value: seeAlso "https://purl.brain-bican.org/taxonomy/CCN20230722/CS202
 id: CL:4300351
 name: CBX MLI Megf11 Gaba molecular layer interneuron (Mmus)
 def: "A molecular layer interneuron of the Mus musculus brain. It is distinguished from other cells in the brain by selective expression of Homer3 (Mmus), Tfap2b (Mmus), Sorcs3 (Mmus), Cacna1e (Mmus). Reference transcriptomic data for this type can be found in the dataset/taxonomy - Yao et al. (2023), Whole Mouse Brain in cell set Subclass:311 CBX MLI Megf11 Gaba." [https://doi.org/10.1038/s41586-023-06812-z]
+comment: Location is inferred from registration of MerFish Data. Registration may be inaccurate, giving misleading location to adjacent anatomical regions.
 subset: mouse_subset
 synonym: "311 CBX MLI Megf11 Gaba" EXACT []
-synonym: "Homer3 (Mmus), Tfap2b (Mmus), Sorcs3 (Mmus), Cacna1e (Mmus) expressing molecular layer interneuron of brain (Mus musculus)" EXACT []
```
RiveraAndrea83, line 48528: I see the synonym tag a lot with gene sets being removed. I assume that is correct as they aren't really synonims.

src/ontology/reports/cl-base-diff.md
```diff
@@ -16,784 +16,1530 @@
...
+### Astro-NT NN_1 Six3os1 astrocyte (Mmus) `http://purl.obolibrary.org/obo/CL_4307031`
+#### Removed
+- [Astro-NT NN_1 Six3os1 astrocyte (Mmus)](http://purl.obolibrary.org/obo/CL_4307031) [has_exact_synonym](http://www.geneontology.org/formats/oboInOwl#hasExactSynonym) "Myoc (Mmus), Six3os1 (Mmus), Slc36a2 (Mmus) expressing astrocyte of brain (Mus musculus)"^^[string](http://www.w3.org/2001/XMLSchema#string) 
 
-- [D1-NUDAP medium spiny neuron](http://purl.obolibrary.org/obo/CL_4030054) [label](http://www.w3.org/2000/01/rdf-schema#label) "D1-NUDAP medium spiny neuron"^^[string](http://www.w3.org/2001/XMLSchema#string) 
+#### Added
+- [Astro-NT NN_1 Six3os1 astrocyte (Mmus)](http://purl.obolibrary.org/obo/CL_4307031) [comment](http://www.w3.org/2000/01/rdf-schema#comment) "Location is inferred from registration of MerFish Data. Registration may be inaccurate, giving misleading location to adjacent anatomical regions."^^[string](http://www.w3.org/2001/XMLSchema#string) 
 
 
-### Lamp5 Lhx6 GABAergic interneuron (Homo sapiens) `http://purl.obolibrary.org/obo/CL_4072022`
```
RiveraAndrea83, line 105: there are lots of terms removed in this way. is this correct?

cl.obo
hkir-dev: Yes, that was a style David and Richard used to like, but we’ve now decided it’s too long and doesn’t really add any value.

src/ontology/reports/cl-base-diff.md
hkir-dev: Checking diff of diffs can be misleading. In the generated [final diff report](https://github.com/obophenotype/cell-ontology/blob/e4e4d78386927f3669fa7ead58fbdf20a48cb3d7/src/ontology/reports/cl-base-diff.md), this term is not in the deleted terms list as far as I can see.

RiveraAndrea83: thanks :)

cl.obo
dosumis: I can confirm that this is correct. They made sense for some work we did a few years back, but not for these cell types.  The legacy code should have been removed earlier.  No-one will miss these synonyms as they are not being used anywhere.

cl-base.owl
dosumis: Don't know why this has changed. AFAIK we've standardised on dc/terms for all Dublin Core.  Not doing so should break QC checks.

gouttegd: > Not doing so should break QC checks.

It does not because `dc/elements/1.1/creator` has been [specifically added](https://github.com/obophenotype/cell-ontology/pull/3441) to the list of allowed annotation properties.

Caroline-99: I am working with @hkir-dev  to fix this. We will be using `dc/terms/creator` for agentic edits.

PR #3478, opened 2025-11-25
20251125-CL-release

Caroline-99:


src/ontology/reports/cl-edit.owl-obo-report.tsv
```diff
@@ -1,4 +1,24 @@
 Level	Rule Name	Subject	Property	Value
+WARN	duplicate_exact_synonym	CL:4310127	oboInOwl:hasExactSynonym	VIP
```
aleixpuigb, line 2: I will not stop the release for this, but I think these warnings should be fixed by next release

aleixpuigb (approved):

src/ontology/subsets/blood_and_immune_upper_slim.json
```diff
@@ -22078,6 +22078,9 @@
         }, {
           "pred" : "http://purl.obolibrary.org/obo/RO_0002175",
           "val" : "http://purl.obolibrary.org/obo/NCBITaxon_4751"
+        }, {
+          "pred" : "http://www.w3.org/2004/02/skos/core#prefLabel",
```
gouttegd, line 22082: We have a bunch of added `skos:prefLabel` annotations in that release…

There’s nothing inherently wrong with that, but I do wonder where they come from. They seem to concern mostly Uberon terms, but AFAIK Uberon does _not_ use `skos:prefLabel`?

Not a blocker for the release, just making a note.

gouttegd (approved):

src/ontology/subsets/blood_and_immune_upper_slim.json
Caroline-99: @hkir-dev does it come from the WMB?

hkir-dev: It is coming from the [Developing Human Brain Atlas Ontology](https://raw.githubusercontent.com/brain-bican/developing_human_brain_atlas_ontology/main/dhbao-base.owl)

Issue #3527, opened 2025-12-15
[Class hierarchy] Make STRd cholinergic-GABAergic neuron child of GABAergic interneuron


RiveraAndrea83:
**CL term**
CL_4310128

**Suggested revision of class hierarchy**
Subclass of some CL_0011005

PR #3528, opened 2025-12-15
#3527 Fix relationship of STRd cholinergic-GABAergic

RiveraAndrea83:
Changed subclass to GABAergic interneuron.

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -2284,6 +2284,7 @@ Declaration(Class(obo:CL_0020024))
 Declaration(Class(obo:CL_0020025))
 Declaration(Class(obo:CL_0020026))
 Declaration(Class(obo:CL_0020027))
+Declaration(Class(obo:CL_0020028))
 Declaration(Class(obo:CL_0700009))
 Declaration(Class(obo:CL_1000001))
 Declaration(Class(obo:CL_1000022))
@@ -3448,6 +3449,7 @@ Declaration(Class(obo:UBERON_0002162))
 Declaration(Class(obo:UBERON_0002197))
 Declaration(Class(obo:UBERON_0002219))
 Declaration(Class(obo:UBERON_0002222))
+Declaration(Class(obo:UBERON_0002282))
 Declaration(Class(obo:UBERON_0002283))
 Declaration(Class(obo:UBERON_0002339))
 Declaration(Class(obo:UBERON_0002439))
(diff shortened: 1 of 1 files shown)


matentzn:
@RiveraAndrea83 nice to meet you too :) Can you point me to the SOP for that component? "offline conversations" are usually hard antipatterns for community curated ontologies - but I do not who _exactly_ is resposible for src/ontology/components/bgo-cl-comp.owl. If there is no SOP, its worth creating an issue for that - by default I would apply the same guidelines to your change as to changes to cl-edit.owl. Which would be, in this case:

1. Create a corresponding issue motivating the change
2. Add the related issue like this: `AnnotationAssertion(obo:IAO_0000233 obo:CL_0000181 <https://github.com/obophenotype/cell-ontology/issues/720>)`

You can also have @dosumis overrule me but this is my religion :P

matentzn:
@dosumis are these terms tagged with a subset in such a way that we can say: 

If 1) a term is tagged and 2) a change is proposed and 3) the proposer has the greenlight by one of the people involved (or _is_ one of the people involved) the change is just approved without community review?

dosumis:
Having investigated further I have realised I was wrong about this:

> I think there's a broader technical issue here to do with the BG component*, which should be base but, from this and other phenotypes, looks like it is not. CC @ubyndr. 

On the substance and desirability of the edit - @RiveraAndrea83, this appears to be an attempt to edit the subclass of the the term mentioned in the PR and ticket - where that subclass is defined in the BG extension.  Having looked at this, I don't think that either needs editing.

> @dosumis are these terms tagged with a subset in such a way that we can say:
> If 1) a term is tagged and 2) a change is proposed and 3) the proposer has the greenlight by one of the people involved (or is one of the people involved) the change is just approved without community review?

I agree that we should tag these terms with a subset tag so that they are easily identifiable.  

The nature of the build requires a separate repo.  These classes are partly defined with reference to data. Community input comes from our discussions with the BICAN/HMBA community - much of which is pre-publication. Some of this is on the repo for this build, but much is in closed channels due to sharing of embargoed data.  However, we are committed to providing supporting evidence for mapping of these data-linked classes to previously published  classical and transcriptomic types, which informs their classification in CL.  e.g. see Rationale in comment here: http://purl.obolibrary.org/obo/CL_4310096. With help from of agentic AI and deepsearch, we aim to achieve good coverage and detail in these rationales.  I think this provides more transparency than the average discussion thread on a ticket.  The CL user community can, of course, request changes to these terms and their classification on here.  The PRs will just be on a separate repo.

PR #3643, opened 2026-05-19
Remove irrelevant comment from CL_4023106

matentzn:
See https://github.com/EnvironmentOntology/envo/issues/1656 

cc @jonquet


ubyndr:
Hi @matentzn . I did some edits. The check was scoped to CL_ terms only because the single violation flagged (RO_0002258 IAO_0000116 "false"^^xsd:boolean) is a bug in the upstream RO ontology, not in CL itself. The boolean false editor note on the RO property "developmentally preceded by" is imported via merged_import.owl and causes a false positive in our QC. Restricting the check to the CL namespace ensures we catch real issues in our own terms without being blocked by upstream data quality problems we don't control.

matentzn:
@ubyndr nice thanks! Can you approve and merge?

RiveraAndrea83 (approved):

PR #3717, opened 2026-09-04
Buccal mucosa cell edits

ct31-sanger:
Corrected endothelial to epithelial for buccal mucosa cell definition. Accepted suggested parent class of keratinocyte for buccal mucosa cell, with supporting reference. Added additional comment for buccal mucosa cell. Added missing cross reference for keratinocyte definition.

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -17755,7 +17755,7 @@ SubClassOf(obo:CL_0002335 obo:CL_0002334)
 AnnotationAssertion(Annotation(oboInOwl:hasDbXref "GOC:tfm") Annotation(oboInOwl:hasDbXref "MESH:D009061") obo:IAO_0000115 obo:CL_0002336 "An epithelial cell that lines the oral cavity including the mucosa of the gums, the palate, the lip, and the cheek.")
 AnnotationAssertion(terms:contributor obo:CL_0002336 <https://orcid.org/0000-0003-1980-3228>)
 AnnotationAssertion(oboInOwl:creation_date obo:CL_0002336 "2010-09-20T02:52:54Z"^^xsd:dateTime)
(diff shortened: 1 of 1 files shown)


ct31-sanger:
#gogoeditdiff /clara

dosumis:
@ubyndr - Interesting clara output.  I think this points to some tweaks we need to do. Specifically I would have expected it to test this change against the added ref.

. [keratinized cell of the oral mucosa](http://purl.obolibrary.org/obo/CL_0002170) [IAO_0000115](http://purl.obolibrary.org/obo/IAO_0000115) "A keratinized buccal mucosa cell. These are mostly located in the hard palate and gingiva."
. [hasDbXref](http://www.geneontology.org/formats/oboInOwl#hasDbXref) "PMID:12014572"

But it seems to have deliberately ignored. The reason seems to be here, but not sure I understand.

> The routed change for this target has predicate_iri/value/value_label all null, so it doesn't expose enough structure to faithfully verbalise the logical definition per the CLARA agent instructions.

I think this might be a relevant artefact from the [job](https://github.com/obophenotype/cell-ontology/actions/runs/33852530599): [routing.json](https://github.com/obophenotype/cell-ontology/actions/runs/33852530599/artifacts/9928944559)

dosumis (changes requested):
One minor change. 

I'm not super-keen on using URLs as links to papers as we can't be sure of their stability.
An alternative reference for the comment is

https://www.merriam-webster.com/dictionary/buccal

Qe can reference this as a compact URL using the standard set here (Bioregistry is our standard reference for such things and OLS uses it to roll hyper-links)

https://bioregistry.io/registry/merriamwebster

Please replace the URL with on comment with
database_cross_reference: merriamwebster:buccal

ct31-sanger:
#gogoeditdiff

dosumis (approved):

PR #3725, opened 2026-09-07
Fixing logical definition #3723

### LINKED ISSUE #3723: [Logical def] fibroblast of dense regular elastic tissue
### CL_1000307 fibroblast of dense regular elastic tissue → drags `tendon cell` into elastic tissue
```
def:  "A fibroblast that is part of the dense regular ELASTIC tissue."
EQ:   fibroblast and part_of  dense regular CONNECTIVE tissue   [UBERON_0007846]
+SC:  part_of  elastic tissue                                   [UBERON_0002521]
```
The EQ is broader than the label (connective, not elastic), so **`tendon cell` (CL_0000388)** classifies here —
and inherits `part_of elastic tissue`. Tendon is dense regular *collagenous* tissue. Both halves are wrong:
fix the EQ to use elastic tissue, and the extra axiom then becomes redundant rather than dangerous & can be removed.

--- General background ---

**The pattern.** A class `C` has both an `EquivalentClasses` axiom (so anything matching the genus+differentia
is *classified as* `C`) **and** extra `SubClassOf(C, R some F)` axioms. Those extra restrictions are not part
of what makes something a `C`, but every class the reasoner puts under `C` inherits them. Functionally this is
a general class axiom — "everything matching C's definition also R-some-F" — asserted invisibly, never
reviewed as such, and firing on classes the author never saw.  

This pattern is sometimes justified, but should be used with care.  When not, it causes problems like the one in this ticket.

ct31-sanger:
Removed "SubClass Of elastic tissue" for 'fibroblast of dense regular elastic tissue' to prevent tendon cell from inheriting 'part of elastic tissue'. #3723

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -26913,9 +26913,8 @@ SubClassOf(Annotation(oboInOwl:is_inferred "true") obo:CL_1000306 obo:CL_0000057
 AnnotationAssertion(Annotation(oboInOwl:hasDbXref "GOC:tfm") obo:IAO_0000115 obo:CL_1000307 "A fibroblast that is part of the dense regular elastic tissue.")
 AnnotationAssertion(oboInOwl:hasDbXref obo:CL_1000307 "FMA:261287")
 AnnotationAssertion(rdfs:label obo:CL_1000307 "fibroblast of dense regular elastic tissue")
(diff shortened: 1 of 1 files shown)


dosumis:
#gogoeditdiff

dosumis (approved):
the reasoned diff shows that the fix removed the incorrect inference.

Issue #3724, opened 2026-09-07
[Logical def] - hidden GCIs


dosumis:
```
EQ:   smooth muscle cell of large intestine and part_of colon    [UBERON_0001155]
+SC:  part_of  epithelium of large intestine                     [UBERON_0001278]
```
The hidden axiom is wrong on its face (a smooth muscle cell is not part of an epithelium) and is inherited by
the inferred child `smooth muscle cell of cecum` (CL_1000281).

----

```
EQ:   serous secreting cell and part_of saliva-secreting gland   -- says nothing about demilunes
+SC:  capable_of saliva secretion
```
`serous acinar cell of salivary gland` (CL_4052065) is inferred to be a *demilune* cell. Demilune is a specific
(and partly fixation-artefactual) crescent arrangement; the EQ omits it entirely.

----

```
EQ:   neuroendocrine cell and part_of epithelium of stomach
+SC:  capable_of  acid secretion                                 [GO_0046717]
```
Inferred children `type D cell of stomach` (somatostatin — *inhibits* acid) and `type G cell of stomach`
(gastrin — *stimulates* parietal cells). Neither secretes acid; parietal cells do.

----

```
+SC:  capable_of (glutamate secretion, neurotransmission and part_of excitatory chemical synaptic transmission)
+SC:  capable_of_part_of  visual perception
```
`interplexiform cell` (CL_0011104) is inferred here and inherits excitatory glutamatergic transmission.
Interplexiform cells are characteristically dopaminergic/GABAergic.

----

- [ ]  CL_1000280 smooth muscle cell of colon → smooth muscle asserted part of an epithelium

```
EQ:   smooth muscle cell of large intestine and part_of colon    [UBERON_0001155]
+SC:  part_of  epithelium of large intestine                     [UBERON_0001278]
```
The hidden axiom is wrong on its face (a smooth muscle cell is not part of an epithelium) and is inherited by
the inferred child `smooth muscle cell of cecum` (CL_1000281).

----

- [ ] CL_0020059 serous demilune cell of salivary gland → every salivary serous cell becomes a demilune cell

```
EQ:   serous secreting cell and part_of saliva-secreting gland   -- says nothing about demilunes
+SC:  capable_of saliva secretion
```
`serous acinar cell of salivary gland` (CL_4052065) is inferred to be a *demilune* cell. Demilune is a specific
(and partly fixation-artefactual) crescent arrangement; the EQ omits it entirely.

[...]

PR #3733, opened 2026-09-07
Fix interplexiform cell hidden GCI

### LINKED ISSUE #3732: CL_0000748 retinal bipolar neuron → `interplexiform cell` made glutamatergic/excitatory

ct31-sanger:
Removal of the equivalent to definition for retinal bipolar neuron to fix GCI of interplexiform cell. #3724 #3732

### DIFF
--- a/src/ontology/cl-edit.owl
+++ b/src/ontology/cl-edit.owl
@@ -10245,6 +10245,7 @@ AnnotationAssertion(Annotation(oboInOwl:hasDbXref "doi:10.1038/s41598-020-66092-
 AnnotationAssertion(Annotation(oboInOwl:hasDbXref "doi:10.1038/s41598-020-66092-9") Annotation(oboInOwl:hasSynonymType obo:OMO_0003000) Annotation(oboInOwl:hasSynonymType obo:OMO_0003004) oboInOwl:hasRelatedSynonym obo:CL_0000740 "RGCs")
 AnnotationAssertion(rdfs:label obo:CL_0000740 "retinal ganglion cell")
 SubClassOf(obo:CL_0000740 obo:CL_0000540)
+SubClassOf(obo:CL_0000740 obo:CL_0000748)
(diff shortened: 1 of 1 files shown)


ct31-sanger:
#gogoeditdiff

dosumis (changes requested):
Why remove the EC def?  Lots of subclassification lost (see reasoned diff)

ct31-sanger:
> Why remove the EC def? Lots of subclassification lost (see reasoned diff)

Ah, yes. Should the EC for rental bipolar neuron be more specific/include synaptic transmission to prevent subclass of inference for interplexiform cell?

dosumis:
YOu need to relax to subclassof rather than delete.  The standard way to do this is with two SubClassOf axioms.  This is semantically equivalent to having one statement with two clauses connected by an AND, but some downstream tooling prefers separated statements.

You then need to add back the retinal ganglion cells underneath the gropuing class as they are no longer automatically classified (alternative would be to add an extra clause as you suggest, but I think no so straightforward and manually classifying 3 classes underneath is easier).

ct31-sanger:
#gogoeditdiff

ct31-sanger:
#gogoeditdiff

ct31-sanger:
@ai4c-agent please show the branch conflicts

ct31-sanger:
@ai4c-agent please accept the master's version of the removed two SubClassOf axioms on CL:0002167 from PR #3734.

dosumis:
Edits have made ontology unsatisfiable

See QC reports here:

image

IIRC this is meant to automatically trigger posting of explanation. I'm looking into why this didn't happen?

You can do this manually using 
 - run ROBOT explain inconsistency (see https://robot.obolibrary.org/explain)
and then copy and paste the markdown into this PR

I can then help you interpret and suggest fixes.

dosumis (changes requested):
Your implementation of the fix I requested looks OK to me.  I have no idea why the ontology is now inconsistent.  Please investigate (see comment)

ct31-sanger:
#gogoeditdiff

dosumis (approved):

